PR #4, opened 2023-09-11
Use ServiceMonitor config to deploy Prometheus in K8s

Deflaimun:
Fixes https://github.com/redpanda-data/documentation-private/issues/1651

Relies on https://github.com/redpanda-data/redpanda/issues/13321 being resolved.

### DIFF
--- a/modules/manage/partials/monitoring.adoc
+++ b/modules/manage/partials/monitoring.adoc
@@ -27,11 +27,21 @@ https://prometheus.io/[Prometheus^] is a system monitoring and alerting tool. It
 
 NOTE: Redpanda recommends creating monitoring dashboards with `/public_metrics`.
 
+:env-kubernetes:
+
 ifdef::env-kubernetes[]
 
-To configure Prometheus to monitor Redpanda metrics, you can use the https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus[Prometheus Helm chart^]:
+To configure Prometheus to monitor Redpanda metrics in Kubernetes, you can use the https://prometheus-operator.dev/[Prometheus Operator^]:
 
-. Add annotations to your Pods to allow the Prometheus Helm chart to find and scrape the `/public_metrics` endpoint:
+. Install Prometheus Operator CRDs:
++
+[,bash]
+----
+LATEST=$(curl -s https://api.github.com/repos/prometheus-operator/prometheus-operator/releases/latest | jq -cr .tag_name)
+curl -sL https://github.com/prometheus-operator/prometheus-operator/releases/download/${LATEST}/bundle.yaml | kubectl create -f -
+----
+
+. Deploy Redpanda with monitoring enabled:
 +
 [tabs]
 ======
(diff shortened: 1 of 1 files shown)


modules/manage/partials/monitoring.adoc
```diff
@@ -27,11 +27,21 @@ https://prometheus.io/[Prometheus^] is a system monitoring and alerting tool. It
 
 NOTE: Redpanda recommends creating monitoring dashboards with `/public_metrics`.
 
+:env-kubernetes:
+
 ifdef::env-kubernetes[]
 
-To configure Prometheus to monitor Redpanda metrics, you can use the https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus[Prometheus Helm chart^]:
+To configure Prometheus to monitor Redpanda metrics in Kubernetes, you can use the https://prometheus-operator.dev/[Prometheus Operator^]:
```
Deflaimun, line 32:
```suggestion
To configure Prometheus to monitor Redpanda metrics in Kubernetes, you can use the https://prometheus-operator.dev/[Prometheus Operator^]: 
```

Feediver1:
@Deflaimun I noticed that the Netlify previews were using the URL for the updated files in the previews, but now is not. Does this need to be explicitly configured for public?

Deflaimun:
@Feediver1 preview was missing a key config. This has been addressed.

modules/manage/partials/monitoring.adoc
```diff
@@ -48,11 +58,9 @@ metadata:
...
   chartRef: {}
   clusterSpec:
-    statefulset:
-      annotations:
-        "prometheus.io/scrape": "true"
-        "prometheus.io/path": "public_metrics"
-        "prometheus.io/port": "<admin-api-port>"
+    monitoring:
+      enabled: true
+      scrapeInterval: 30s
```
alejandroEsc, line 81: I have a PR making this optional now, no longer required (scrapeInterVal). We should make a note that if you do not install prometheus, the servicemonitor will not be created even if you set enabled: true. This behavior may be changed soon however.

modules/manage/partials/monitoring.adoc
```diff
@@ -99,24 +106,129 @@ helm upgrade --install redpanda redpanda/redpanda \
...
+
+. Wait until Prometheus is ready:
++
+[,bash]
+----
+kubectl get --namespace default prometheus prometheus --watch
 ----
 
-. Install Prometheus:
+. Deploy a NodePort Service for external access to Prometheus:
```
alejandroEsc: Managing prometheus and getting access to it externally seems like it belongs to pointers for prometheus specific documentation. Not sure we need to here, we probably can point instead to just using port forwarding with kubectl as most documentation for prometheus does.

alejandroEsc (approved):
lgtm, just a few comments

modules/manage/partials/monitoring.adoc
alejandroEsc: OK my team agreed that we should never guard users like this. So we will explicitly fail if you enable monitoring but do not have the CRDs installed.

alejandroEsc (approved):
lgtm

Feediver1 (approved):
lgtm

PR #31, opened 2023-09-22
docs: Fix CRD template

JakeSCahill:
Fixed some issues in the CRD template:

- Missing `{p}` attribute.
- Headings out of order.


JakeSCahill:
Will fix in the Topic CRD PR

PR #34, opened 2023-09-25
engine: add tech-preview to dropdown list

Deflaimun:
Resolves https://github.com/redpanda-data/documentation-private/issues/1985

Please also check the integrity of https://github.com/redpanda-data/docs/tree/technical-preview

PR #35, opened 2023-09-25
engine: update gulp

Deflaimun:


Feediver1:
Is there an issue for this? Wondering what the nature of this update is.

JakeSCahill:
@Feediver1 this fixes an issue in local builds when using `npm run start`. The build output wasn't being cleaned which led to inconsistent behavior. This PR configures Antora to clean the build output before building the docs.

JakeSCahill (approved):

Deflaimun:
@Feediver1 It's part of https://github.com/redpanda-data/docs/pull/34
`npm run start` wasn't cleaning the cache before running, thus leading into errors that were not true.

PR #57, opened 2023-10-03
AKS nodes need external IP added for NodePort

vuldin:
The Azure UI doesn't give the option to add an external IP to nodes, so it's likely that many people trying to use NodePorts for external access will fail. This PR mentions this is the case and links to instructions for how to resolve this.


JakeSCahill (approved):
Awesome thanks @vuldin

PR #58, opened 2023-10-04
Minor edits to Migrate with MirrorMaker 2 doc

kbatuigas:
- Fix typo with connection details placeholder
- Also edited the opening paragraph to make it more succinct and a bit less "marketing" sounding. I did end up removing the link to the [Kafka compatibility](https://docs.redpanda.com/current/develop/kafka-clients/) page in the process though, so I'm open to modifying this PR to add it back in if that's better

Fixes https://github.com/redpanda-data/documentation-private/issues/1999

### DIFF
--- a/modules/upgrade/pages/migrate/data-migration.adoc
+++ b/modules/upgrade/pages/migrate/data-migration.adoc
@@ -2,9 +2,7 @@
 :description: Use MirrorMaker 2 to replicate data between Redpanda clusters.
 :page-aliases: data-management:data-migration.adoc, manage:data-migration.adoc
 
-When you want to migrate data from Kafka or replicate data between Redpanda clusters, you can use the MirrorMaker 2 (MM2) tool bundled in the Kafka download package.
-
-This section describes the process of configuring MirrorMaker 2 to replicate data from one Redpanda cluster to another.
+When you want to migrate data from Kafka or replicate data between Redpanda clusters, you can use the MirrorMaker 2 (MM2) tool bundled in the Kafka download package. This section describes the process of configuring MirrorMaker 2 to replicate data from one Redpanda cluster to another.
 
 MirrorMaker 2 reads a configuration file that specifies the connection details for the Redpanda clusters, as well as other options.
 After you start MirrorMaker 2, the data migration continues according to the configuration at launch time until you shutdown the MirrorMaker 2 process.
--- a/modules/upgrade/pages/migrate/data-migration.adoc
+++ b/modules/upgrade/pages/migrate/data-migration.adoc
@@ -61,7 +61,6 @@ cat << EOF > mm2.properties
 clusters = <cluster_name_1>, <cluster_name_2>
 
 // Assign IP addresses to the cluster names
-// <cluster>.bootstrap.servers may also be a comma-separated list of bootstrap hosts, if the cluster has several
 <cluster_name_1>.bootstrap.servers = <cluster_1_ip>:9092
 <cluster_name_2>.bootstrap.servers = <cluster_2_ip>:9092
 
--- a/modules/upgrade/pages/migrate/data-migration.adoc
+++ b/modules/upgrade/pages/migrate/data-migration.adoc
@@ -2,14 +2,14 @@
 :description: Migrate external data to Redpanda with MirrorMaker 2.
 :page-aliases: data-management:data-migration.adoc, manage:data-migration.adoc
 
-When you want to migrate data from Kafka or replicate data between Redpanda clusters, you can use the MirrorMaker 2 (MM2) tool bundled in the https://kafka.apache.org/downloads[Kafka download package^].
+When you want to migrate data from Kafka or replicate data between Redpanda clusters, you can use the MirrorMaker 2 (MM2) tool bundled in the Kafka download package.
 
 This section describes the process of configuring MirrorMaker 2 to replicate data from one Redpanda cluster to another.
 
 MirrorMaker 2 reads a configuration file that specifies the connection details for the Redpanda clusters, as well as other options.
 After you start MirrorMaker 2, the data migration continues according to the configuration at launch time until you shutdown the MirrorMaker 2 process.
 
-For details on MirrorMaker 2 and its options, see the https://kafka.apache.org/documentation/#georeplication[Kafka documentation^].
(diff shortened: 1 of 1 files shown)


modules/upgrade/pages/migrate/data-migration.adoc
```diff
@@ -2,9 +2,7 @@
 :description: Migrate external data to Redpanda with MirrorMaker 2.
```
micheleRP:
```suggestion
:description: Use MirrorMaker 2 to replicate data between Redpanda clusters.
```

micheleRP: @kbatuigas this description only appears on the index page--it doesn't get duplicated on data-migration.adoc. So this edit only updates https://deploy-preview-58--redpanda-docs-preview.netlify.app/current/upgrade/migrate/

modules/upgrade/pages/migrate/data-migration.adoc
```diff
@@ -2,9 +2,7 @@
...
 :page-aliases: data-management:data-migration.adoc, manage:data-migration.adoc
 
-One of the more elegant aspects of Redpanda is that it is compatible with the xref:develop:kafka-clients.adoc[Kafka API and ecosystem].
-So when you want to migrate data from Kafka or replicate data between Redpanda clusters,
-the MirrorMaker 2 (MM2) tool bundled in the https://kafka.apache.org/downloads[Kafka download package^] is a natural solution.
+When you want to migrate data from Kafka or replicate data between Redpanda clusters, you can use the MirrorMaker 2 (MM2) tool bundled in the https://kafka.apache.org/downloads[Kafka download package^].
 
 This section describes the process of configuring MirrorMaker 2 to replicate data from one Redpanda cluster to another.
```
micheleRP:
```suggestion
When you want to migrate data from Kafka or replicate data between Redpanda clusters, you can use the MirrorMaker 2 (MM2) tool bundled in the https://kafka.apache.org/downloads[Kafka download package^]. This section describes the process of configuring MirrorMaker 2 to replicate data from one Redpanda cluster to another.
```

micheleRP:
Since the 2 links in the intro section are both to Kafka doc, you could consider putting them together at the end (instead of splitting them); e.g., 
See the Kafka documentation for [download instructions](url) and [details about cross-cluster data mirroring](url).

micheleRP:
Hi @hcoyote! Kat [implemented this change](https://deploy-preview-58--redpanda-docs-preview.netlify.app/current/upgrade/migrate/data-migration/), but would you please confirm if these editable variables would be clearer as: 
<cluster_name1>, <cluster_name1_ip>
<cluster_name2>, <cluster_name2_ip> 

OR just update this: 
```
// Assign IP addresses to the cluster names
<cluster_name_1>.bootstrap.servers = <cluster_name_1>_cluster_ip>:9092
<cluster_name_2>.bootstrap.servers = <cluster_name_2>_cluster_ip>:9092
```

modules/upgrade/pages/migrate/data-migration.adoc
```diff
@@ -2,9 +2,7 @@
 :description: Migrate external data to Redpanda with MirrorMaker 2.
 :page-aliases: data-management:data-migration.adoc, manage:data-migration.adoc
 
-One of the more elegant aspects of Redpanda is that it is compatible with the xref:develop:kafka-clients.adoc[Kafka API and ecosystem].
-So when you want to migrate data from Kafka or replicate data between Redpanda clusters,
-the MirrorMaker 2 (MM2) tool bundled in the https://kafka.apache.org/downloads[Kafka download package^] is a natural solution.
+When you want to migrate data from Kafka or replicate data between Redpanda clusters, you can use the MirrorMaker 2 (MM2) tool bundled in the https://kafka.apache.org/downloads[Kafka download package^].
```
kbatuigas: @micheleRP Regarding this suggestion: https://github.com/redpanda-data/docs/pull/58#issuecomment-1747190424 Did you mean, take out the link to Kafka downloads here, and then edit [line 12](https://github.com/redpanda-data/docs/pull/58/files#diff-312d00c897a59bbc918242d8ec5ccaaf2ea580eb5cd30cd81976228719749cbdR12) so it has the downloads link there instead?

micheleRP: Yes!

Feediver1 (approved):
Looks great!

hcoyote:
@micheleRP for the bootstrap.servers detail:

not sure I understand the first part of the question. The fix that's in the preview does the right thing now.

aside: bootstrap.servers can also be a hostname or (I think) a list of multiple bootstrap hosts in the event that your cluster has several. Not sure if we need to mention a comment in the example or not.

modules/upgrade/pages/migrate/data-migration.adoc
```diff
@@ -65,8 +61,9 @@ cat << EOF > mm2.properties
 clusters = <cluster_name_1>, <cluster_name_2>
 
 // Assign IP addresses to the cluster names
-<cluster_name_1>.bootstrap.servers = <cluster_1>_cluster_ip>:9092
-<cluster_name_2>.bootstrap.servers = <cluster_2>_cluster_ip>:9092
+// <cluster>.bootstrap.servers may also be a comma-separated list of bootstrap hosts, if the cluster has several
```
kbatuigas: @micheleRP What do you think of this comment? I also found some info in the [kafka docs](https://kafka.apache.org/documentation/#georeplication-flow-create)

micheleRP: I don't think the comment is necessary, since bootstrap.servers is plural and defined as a list. But @hcoyote please confirm! Also, my confusing comment about `<cluster_1_ip>` is because I thought bootstrap.servers required both cluster name and cluster IP (and I wanted `<cluster_name_1>` and `<cluster_1>` to match, so if a user edited that code in the doc, the name would get edited in both places). But apparently only IP is required!

hcoyote: as far as I know, it's a comma separated list of bootstrap addresses (name or ip), so it doesn't need to have both the name and ip address of each host.

in this case, cluster_name_1 is a logical name for the mm2 config to refer to in various parts of the config, it would not be a host name

hcoyote (approved):

PR #65, opened 2023-10-10
New metrics records produced consumed

kbatuigas:
Add new metrics to https://docs.redpanda.com/current/reference/internal-metrics-reference/ and https://docs.redpanda.com/current/reference/public-metrics-reference/.

Also change section headings (e.g. https://docs.redpanda.com/current/reference/public-metrics-reference/#infrastructure-metrics) to H2 as most of them are currently H3, same level as the metrics. Changing to H2 will probably nest the metrics under the correct section heading and also make it easier for readers to navigate .

Fixes https://github.com/redpanda-data/documentation-private/issues/2020.

### DIFF
--- a/modules/reference/pages/internal-metrics-reference.adoc
+++ b/modules/reference/pages/internal-metrics-reference.adoc
@@ -37,11 +37,9 @@ If this is the last record received by the cluster, then the cluster is up-to-da
 
 ---
 
-=== vectorized_cluster_partition_last_stable_offset
+=== vectorized_cluster_partition_schema_id_validation_records_failed
 
-Last stable offset.
-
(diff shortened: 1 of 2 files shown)


JFlath (commented):
LGTM - I note that `vectorized_cluster_partition_last_stable_offset` was added and then removed in different commits, but this wasn't a part of the initial issue, so perhaps is targeted elsewhere?

JFlath (approved):

kbatuigas:
> LGTM - I note that `vectorized_cluster_partition_last_stable_offset` was added and then removed in different commits, but this wasn't a part of the initial issue, so perhaps is targeted elsewhere?

Ah, yes, that was a copy-paste mistake on my part @JFlath

micheleRP (approved):
lgtm

PR #66, opened 2023-10-11
Tiered storage behavior when disabling remote read/write

kbatuigas:
- Add a note in https://docs.redpanda.com/current/manage/tiered-storage/#enable-tiered-storage to clarify that data in object storage is not deleted if `redpanda.remote.read` and `redpanda.remote.write` are disabled

Fixes https://github.com/redpanda-data/documentation-private/issues/2015


modules/manage/partials/tiered-storage.adoc
```diff
@@ -637,6 +637,8 @@ rpk topic alter-config <topic_name> --set redpanda.remote.read=true --set redpan
 
 Topic-level properties override cluster-level properties. For example, for new topics, if `cloud_storage_enable_remote_write` is set to `true`, you can set `redpanda.remote.write` to `false` to turn it off for a particular topic.
 
+NOTE: If you disable remote writes and reads on a topic, that is, set `redpanda.remote.read` and `redpanda.remote.write` to `false`, the data that has been offloaded to object storage is not deleted. When you re-enable remote writes and reads, Redpanda resumes uploading data to object storage starting from the last write point, and resumes fetching data from object storage from the last read point.
```
kbatuigas: @JFlath I'd like to specifically get the "resumes uploading / fetching data from the last write / read point" part checked. Is that correct? Also if you know of more accurate and specific terminology or phrasing we can use here, I'd appreciate it!

VladLazar: `resumes fetching data from object storage from the last read point` should be worded differently. Readers can query any point in the log, so there's no resuming there. When remote read is re-enabled, the part of the log uploaded to object storage becomes available for reads again.

kbatuigas: @VladLazar I think how you phrased it for reads sounds pretty good, so I'll just use that if you don't mind!

JFlath: I agree with Vlad's clarification - +1 from me with that :)

JFlath (approved):

micheleRP (approved):
lgtm

PR #82, opened 2023-10-21
fix version of twmb/tlscfg

Q00:
Hi this is Q. I tried to make chat demo. But I got error when I tried to install tlscfg.

In twmb/tlscfg, there isn't v1.3.1
You can check its version in https://github.com/twmb/tlscfg/tags.
So I changes version v1.3.1 to v1.2.1 that latest version of tlscfg

Also I add newline to modules/develop/partials/chat-app.adoc.


Q00:
I am going to close this PR. Check the #83. Thankjs.

PR #83, opened 2023-10-23
fix version and dependency of golang chat demo

Q00:
Hi this is JQ. As I send [PR#82](https://github.com/redpanda-data/docs/pull/82) before about version, I realized that I needed to change a dependency in this demo.
Please check my go mod file in demo.
https://github.com/Q00/redpanda-chat/blob/main/go.mod

This demo needs github.com/twmb/franz-go/pkg/kadm instead of github.com/twmb/tlscfg.
When I run `go mod tidy`, it removes github.com/twmb/tlscfg.


JakeSCahill (approved):
Thanks @Q00

PR #94, opened 2023-10-31
update docker compose examples

Deflaimun:
Resolves https://github.com/redpanda-data/documentation-private/issues/2074

Preview: 
https://deploy-preview-94--redpanda-docs-preview.netlify.app/current/reference/docker-compose/


twmb (approved):

Feediver1:
What about:
https://docs.redpanda.com/current/reference/docker-compose/

Feediver1 (approved):
I left a comment on the Conversation tab--there was an additional page where I thought this might also apply. So these are the sum total of all our Docker Compose config examples? I saw some from previous releases that are EOL, but that would not apply. Just want to make sure all are captured. Thanks.

PR #96, opened 2023-11-01
Change listener name that public certificate is refering to

RafalKorepta:
Our documentation correctly describe the process to configure external cert manager cluster issuer to create public certificate. Unfortunately if user follows exact naming scheme, then Redpanda cluster deployment will not work as public certificate does not follow internal to Kubernetes cluster DNS naming scheme.

By default helm chart defines 2 listeners `default` and `external`. User can change those names. Those names are used to reference what TLS definition is assigned to them.

https://github.com/redpanda-data/helm-charts/issues/829#issuecomment-1787144184

### DIFF
--- a/modules/manage/pages/kubernetes/security/kubernetes-tls.adoc
+++ b/modules/manage/pages/kubernetes/security/kubernetes-tls.adoc
@@ -213,7 +213,7 @@ spec:
     tls:
       enabled: true
       certs:
-        default:
+        external:
           issuerRef:
             name: <issuer-name>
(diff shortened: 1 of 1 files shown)


JakeSCahill:
Thanks @RafalKorepta - I’m working on a change that updates this file. It’s going to be quite a big PR so would you mind if I incorporate these changes into that one?

JakeSCahill:
Fixed in https://github.com/redpanda-data/docs/pull/99

PR #104, opened 2023-11-07
Update Rpk 23.3

Deflaimun:
This PR adds the new rpk commands for 23.3.

Preview:
https://deploy-preview-104--redpanda-docs-preview.netlify.app/23.3/reference/rpk/rpk-registry/rpk-registry/
https://deploy-preview-104--redpanda-docs-preview.netlify.app/23.3/reference/rpk/rpk-transform/rpk-transform/

PR #113, opened 2023-11-14
Recovery mode

kbatuigas:
### Deploy preview
https://deploy-preview-113--redpanda-docs-preview.netlify.app/23.3/manage/recovery-mode/

Closes https://github.com/redpanda-data/documentation-private/issues/2040

Feature also requires doc updates to Admin API - this will be handled in a [separate PR](https://github.com/redpanda-data/docs/pull/148)


modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+== About recovery mode
+
+Recovery mode starts a cluster with limited functionality and deactivates normal traffic and data operations for a cluster. This provides a more stable environment for troubleshooting issues and restoring the cluster to a usable state. 
+
+Recovery mode helps to prevent triggering a storm of partition leader elections and replication (such as increased learner replica recovery load after a long broker downtime) that typically occur on startup, especially after some downtime. 
+
+Normal traffic and especially malicious traffic can disrupt availability for all users, including admin users. Recovery mode helps avoid placing the cluster under a heavy load of Kafka API and HTTP Proxy connections as you are working to resuscitate the cluster.
+
+== Requirements
```
kbatuigas: Any other requirements or prerequisites that need to be added? For example, user roles?

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+Recovery mode can only be entered during initial cluster formation as brokers start up, and not while they are already running.
+
+== Use recovery mode
+
+To use recovery mode, first configure the node property to enable recovery mode when a broker starts. When enabled, you can also disable partitions, if the cluster failure is localized to just a few partitions.
+
+=== Enable recovery mode
+
+. Run the `rpk redpanda mode` command to set the `recovery_mode_enabled` broker configuration property to `true`.
```
kbatuigas: Has this been merged yet? It isn't included in our PR to update rpk reference: https://github.com/redpanda-data/docs/pull/104

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+You may find that problems preventing normal cluster startup are often isolated to certain partitions or topics. In these situations, you can use the Admin API to manage thees partitions. You can disable partitions at the topic level, or individual partition level. 
+
+After you disable these partitions and exit recovery mode, starting Redpanda again in non-recovery mode leaves these partitions in a deactivated state. Disabled partitions and topics return a "Replica Not Available" message for Kafka API requests.
+
+The examples below assume that the Admin API port is `8081`.
+
+==== Disable a specific partition of a topic
+
+// TODO: Confirm whether these are new API endpoints and that they are available
```
kbatuigas: Have these Admin API endpoints been merged? Will the admin API specs [here](https://github.com/redpanda-data/redpanda/tree/dev/src/v/redpanda/admin/api-doc) be updated?

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
++
+--
+```bash
+rpk cluster partitions list --topic <topic_name> --disabled
+```
+--
+====
+
+=== Exit recovery mode
+
```
kbatuigas: After exiting recovery mode, are there any things that one may look out for when doing a normal startup for the cluster again?

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
+= Recovery Mode
+:description: 
+
+Recovery mode allows you to start Redpanda and form a stable cluster when it fails due to a system crash, OOM errors, or other issues that make a normal startup infeasible. 
+
+In recovery mode, cluster configuration changes and other manual administrative actions are permitted so you can repair and recover the failed cluster. Communication between producers and consumers, as well as topic reads and writes, are disabled.
```
ztlpn: The  wording is a bit incorrect here

> communication between producers and consumers

Producers and consumers don't communicate directly - they do it via Redpanda!

>  topic reads and writes, are disabled

These and consuming/producing are synonyms i.e. producers do topic writes, consumers do topic reads.

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
ztlpn: Hmm, maybe this is because of [my comment](https://github.com/redpanda-data/redpanda/pull/14579#issuecomment-1788133326)? Yes, it [has been merged](https://github.com/redpanda-data/redpanda/blob/c166c3b2bc777f7c53cdb92197cf20e65bd612ca/src/v/config/node_config.cc#L169).

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+== About recovery mode
+
+Recovery mode starts a cluster with limited functionality and deactivates normal traffic and data operations for a cluster. This provides a more stable environment for troubleshooting issues and restoring the cluster to a usable state. 
+
+Recovery mode helps to prevent triggering a storm of partition leader elections and replication (such as increased learner replica recovery load after a long broker downtime) that typically occur on startup, especially after some downtime. 
+
+Normal traffic and especially malicious traffic can disrupt availability for all users, including admin users. Recovery mode helps avoid placing the cluster under a heavy load of Kafka API and HTTP Proxy connections as you are working to resuscitate the cluster.
```
ztlpn: We should also mention that redpanda won't load user partitions in recovery mode (i.e. will not touch user partition data on disk)

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
ztlpn: Well, enabling it via rpk requires local access to machines running redpanda. I'm not sure what the requirements for enabling it in k8s are. @alejandroEsc can you comment?

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+== Requirements
+
+Recovery mode can only be entered during initial cluster formation as brokers start up, and not while they are already running.
+
+== Use recovery mode
+
+To use recovery mode, first configure the node property to enable recovery mode when a broker starts. When enabled, you can also disable partitions, if the cluster failure is localized to just a few partitions.
```
ztlpn: > you can also disable partitions, if the cluster failure is localized to just a few partitions

we should add "and then restart the cluster normally", otherwise the purpose of disabling partitions is not clear (in recovery mode all user partitions are disabled anyway)

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+=== Enable recovery mode
+
+. Run the `rpk redpanda mode` command to set the `recovery_mode_enabled` broker configuration property to `true`.
+// TODO: confirm whether rpk command is available
```
ztlpn: The rpk command has been merged as well: https://github.com/redpanda-data/redpanda/pull/14431

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+Recovery mode starts a cluster with limited functionality and deactivates normal traffic and data operations for a cluster. This provides a more stable environment for troubleshooting issues and restoring the cluster to a usable state. 
+
+Recovery mode helps to prevent triggering a storm of partition leader elections and replication (such as increased learner replica recovery load after a long broker downtime) that typically occur on startup, especially after some downtime. 
+
+Normal traffic and especially malicious traffic can disrupt availability for all users, including admin users. Recovery mode helps avoid placing the cluster under a heavy load of Kafka API and HTTP Proxy connections as you are working to resuscitate the cluster.
+
+== Requirements
+
+Recovery mode can only be entered during initial cluster formation as brokers start up, and not while they are already running.
```
ztlpn: "Initial cluster formation" usually refers to the time when nodes first join the newly formed cluster. If so, this wording is incorrect, already existing cluster can enter recovery mode as well, it just needs to be restarted.

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+// TODO: required flags
++
+[,bash] 
+---- 
+rpk 
+----
++
+. Start the cluster. You can run `rpk cluster health` to check whether the cluster has entered recovery mode.
+
+All brokers in the cluster must enter recovery mode together. If a cluster starts and a broker that doesn't have recovery mode enabled attempts to join, the broker will exit and return an error.
```
ztlpn: This has been simplified - we allow single nodes to enter recovery mode as well. We should write here that starting a mixed-mode cluster is possible, but not recommended.

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
ztlpn: [yes](https://github.com/redpanda-data/redpanda/blob/10d3d8287bcb0d130bc21d290d1466c7a7fa3ba6/src/v/redpanda/admin/api-doc/cluster.json#L70-L194)

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+All private Redpanda topics such as `__consumer_offsets` are accessible. While data in user-created topics is not available, you can manage the topics metadata.
+
+You can perform the following operations in recovery mode:
+
+* Edit controller metadata
+** Edit cluster configurations
```
ztlpn: configurations -> configuration properties (these: https://docs.redpanda.com/current/reference/cluster-properties/)

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+** Enable and disable features
```
ztlpn: I don't think we document our feature flag APIs so better to omit this

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
```diff
@@ -0,0 +1,203 @@
...
+* Tiered storage cache management
+* Compaction
+
+You can use the Admin API when the cluster is in recovery mode. The HTTP Proxy and Schema Registry APIs are not available.
+
+=== Disable partitions
+
+You may find that problems preventing normal cluster startup are often isolated to certain partitions or topics. In these situations, you can use the Admin API to manage thees partitions. You can disable partitions at the topic level, or individual partition level. 
+
+After you disable these partitions and exit recovery mode, starting Redpanda again in non-recovery mode leaves these partitions in a deactivated state. Disabled partitions and topics return a "Replica Not Available" message for Kafka API requests.
```
ztlpn: message -> error code

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
ztlpn: can't think of anything

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
kbatuigas: @ztlpn Is this referring to a situation where you are using disable partition outside of recovery mode? That is: a) cluster fails due to problem in specific partition -> b) disable the specific partition -> c) restart cluster normally (no need for recovery mode)?

ztlpn: yes, exactly

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
kbatuigas: @ztlpn For the rpk command, is there a specific command and flags that should be used?

Also is there a specific rpk command as well to "exit" recovery mode?

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
kbatuigas: @ztlpn would you mind helping to provide a specific example or case where you would use recovery mode **plus** disable specific partitions? Will there ever be a reason to do both together?

ztlpn: Can't think of anything - while a node is in recovery mode, all user partitions are disabled. No point in disabling them once more.

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
ztlpn: I think it is just `rpk redpanda mode recovery` to set the flag and `rpk redpanda mode production/development` to unset it (depending on whether you want to exit to the production mode or development mode).

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
+* Partition and leader balancers
+* Tiered storage object storage housekeeping
+* Tiered storage cache management
+* Compaction
+
+You can use the Admin API when the cluster is in recovery mode. The HTTP Proxy and Schema Registry APIs are not available.
+
+== Disable partitions
+
+You may find that problems preventing normal cluster startup are often isolated to certain partitions or topics. In these situations, you can use the Admin API to manage these partitions. You can disable partitions at the topic level, or individual partition level.
```
ztlpn: Admin API -> Admin API or rpk

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
+* Edit controller metadata
+** Edit cluster configuration properties
+** Add and remove users and ACLs
+** Add new brokers to the cluster
+** Modify topic properties
+** Delete topics
+** Delete WASM transforms
+* Edit consumer group metadata
+
+Node-wide and cluster-wide processes are disabled as they may disrupt recovery operations:
```
ztlpn: IMO thls list belongs more to the "about recovery mode" section

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
+=== Disable a specific partition of a topic
+
+// TODO: Confirm whether these are new API endpoints and that they are available
+[tabs]
+====
+Curl::
++
+--
+```bash
+curl -X POST -d '{"disabled": true}' http://localhost:8081/cluster/partitions/<namespace>/<topic_name>/<partition_id>
```
ztlpn, line 107: I guess we should mention that the namespace for user topics will be `kafka`. Or just insert it into the urls directly.

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
+= Recovery Mode
+:description: 
+
+Recovery mode allows you to start Redpanda and form a stable cluster when it fails due to a system crash, OOM errors, or other issues that make a normal startup infeasible. 
+
+In recovery mode, cluster configuration changes and other manual administrative actions are permitted so you can repair and restore the failed cluster. Topic reads and writes are disabled.
+
+== About recovery mode
```
ztlpn: This section looks good content-wise but feels a bit disjointed. Should we add some wording tying these together? E.g. preventing election storm is the direct consequence of not loading partitions. And preventing kafka traffic is a different but also important aspect of recovery mode. Also disabling some internal processes (listed in another section) can be added here as yet another aspect.

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+Recovery mode helps to prevent triggering a storm of partition leader elections and replication (such as increased learner replica recovery load after a long broker downtime) that may occur on startup. 
+
+Normal traffic and especially malicious traffic can disrupt availability for all users, including admin users. Recovery mode helps avoid placing the cluster under a heavy load of Kafka API and HTTP Proxy connections as you work to resuscitate the cluster.
+
+== Requirements
+
+To enable recovery mode, ensure that you have [rpk] installed, and you have local access to machines running Redpanda.
+
+Brokers can only enter recovery mode as it starts up, and not while it is already running. First set the broker configuration property to enable recovery mode, and then do a broker restart.
```
mattschumpert: grammar

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+. Run the `rpk redpanda mode` command to set the `recovery_mode_enabled` broker configuration property to `true`.
+// TODO: confirm whether rpk reference is missing this update
++
+[,bash] 
+---- 
+rpk redpanda mode recovery
+----
+
+. Start the cluster. You can run `rpk cluster health` to check whether the cluster has entered recovery mode.
```
mattschumpert, line 59: We should show the output of the cluster health command here @kbatuigas

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
++
+It is possible to start a mixed-mode cluster, where some individual brokers are in recovery mode while others are not, but Redpanda recommends starting all brokers in the cluster in recovery mode together.
```
mattschumpert: @ztlpn  I thought this is explicitly forbidden and prevented .

@kbatuigas

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+All private Redpanda topics such as `__consumer_offsets` are accessible. While data in user-created topics is not available, you can manage the topics metadata.
```
mattschumpert: grammar. which topics?

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+In recovery mode, you can also <<disable partitions,disable partitions>> if the cluster failure is localized to just a few partitions. If you disable specific partitions while in recovery mode, the disabled state persists even when you restart Redpanda again normally.
```
mattschumpert: not 'if' but 'to address situations where the cluster failure is localizes ....'

modules/manage/pages/recovery-mode.adoc
mattschumpert: Agreed.  But also @ztlpn  isn't 'edit controller metadata' the sum of the items below (and we could skip that bullet)?

mattschumpert: > just checking implementation.. @rockwotj is the wasm transforms internal topic part of the internal topics list such that this will work (deleting wasm transforms)?

it's in the controller, so I assume so, although we don't start the transform subsystem at all in node wide recovery mode. What are the requirements here? Just delete transforms?

modules/manage/pages/recovery-mode.adoc
mattschumpert: This list also belongs in the 'about recovery mode' section

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
+** Add new brokers to the cluster
+** Modify topic properties
+** Delete topics
+** Delete WASM transforms
+* Edit consumer group metadata
+
+Node-wide and cluster-wide processes are disabled as they may disrupt recovery operations:
+
+* Partition and leader balancers
+* Tiered storage object storage housekeeping
```
mattschumpert: 'Tiered storage housekeeping'

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
+= Recovery Mode
```
mattschumpert, line 1: Not sure where this is going in the overall doc tree, but I assume under 'manage'

modules/manage/pages/recovery-mode.adoc
mattschumpert: The second sentence in particular might be better placed near the end of this section after we finish with the background on 'why'

mattschumpert (approved):

modules/manage/pages/recovery-mode.adoc
ztlpn: Good point, it looks like this:
```
$ rpk cluster health
CLUSTER HEALTH OVERVIEW
=======================
Healthy:                          true
Unhealthy reasons:                []
Controller ID:                    0
All nodes:                        [0 1 2]
Nodes down:                       []
Nodes in recovery mode:           [0 1 2]
Leaderless partitions (0):        []
Under-replicated partitions (0):  []
```
(note the "Nodes in recovery mode:" field)

modules/manage/pages/recovery-mode.adoc
ztlpn: Current wording is correct and also what we agreed upon (discussions: [1](https://docs.google.com/document/d/1CyVv_ZfEm_K7WAlwD2U7VFok0cOoK8dbUT81HkBsBE8/edit?disco=AAAA6fcaGNY), [2](https://docs.google.com/document/d/1CyVv_ZfEm_K7WAlwD2U7VFok0cOoK8dbUT81HkBsBE8/edit?disco=AAAA41gA9P8)).

Rationale was that 1) enforcing the no-mixed-mode rule requires inter-node coordination that may not be reliable when the cluster is unhealthy and 2) there may be cases where this is actually useful (e.g. fixing just 1 node in a 20-node cluster). So enforcing the no-mixed-mode rule became the cluster operator's responsibility.

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+You can use the Admin API when the cluster is in recovery mode. The HTTP Proxy and Schema Registry APIs are not available.
+
+== Disable partitions
+
+You may find that problems preventing normal cluster startup are often isolated to certain partitions or topics. In these situations, you can use the Admin API to manage these partitions. You can disable partitions at the topic level, or individual partition level. 
+
+You can also disable partitions outside of recovery mode.
+
+If you disable partitions while in recovery mode, starting Redpanda again in non-recovery mode leaves these partitions in a deactivated state. Disabled partitions and topics return a "Replica Not Available" error code for Kafka API requests.
```
kbatuigas: @ztlpn @mattschumpert Does this section then become superfluous/irrelevant if there is [no use case](https://github.com/redpanda-data/docs/pull/113#discussion_r1412500072) for entering recovery mode and then *manually* disabling partitions?

modules/manage/pages/cluster-maintenance/recovery-mode.adoc
alejandroEsc: For k8s you need to have access to the cluster, and be able to run upgrades via helm against the namespace where your release exists. The upgrades are to set the flags in the configs that will enable/disable recovery mode.

alejandroEsc: For operator case: you need access to the cluster and access to the existing namespace and  Redpanda Resource to make the changes.

modules/manage/pages/recovery-mode.adoc
ztlpn: > But also @ztlpn isn't 'edit controller metadata' the sum of the items below (and we could skip that bullet)?

What is listed above under "edit controller metadata" are all user actions. What is listed below are internal Redpanda processes.

modules/manage/pages/recovery-mode.adoc
ztlpn: To clarify my linked comment, I don't see a use case for entering recovery mode, disabling partitions and *staying in recovery mode*. What is described in this line (i.e. starting in recovery, disabling, starting normally) is still useful because to disable partitions you need a cluster with healthy controller (and if you can't start a cluster because of a single partition, you'll *have* to start it in recovery mode first, otherwise you'll have no way to actually set the disabled bit for the partition).

kbatuigas: Very helpful, thank you for the explanation @ztlpn

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,218 @@
...
+To enable recovery mode, ensure that you have xref:get-started:intro-to-rpk.adoc[rpk] installed, and you have local access to machines running Redpanda.
+
+A broker can only enter recovery mode as it starts up, and not while it is already running. You first set the broker configuration property to enable recovery mode, and then do a broker restart.
+
+== Use recovery mode
+
+To use recovery mode, first configure the broker to enter recovery mode when it starts.
+
+. Run the `rpk redpanda mode` command to set the `recovery_mode_enabled` broker configuration property to `true`.
+// TODO: confirm whether rpk reference is missing this update
```
JakeSCahill: It's available in `v23.3.1-rc2`

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,218 @@
...
+* Partition and leader balancers
+* Tiered storage housekeeping
+* Tiered storage cache management
+* Compaction
+
+You can use the Admin API when the cluster is in recovery mode. The HTTP Proxy and Schema Registry APIs are not available.
+
+With normal Kafka traffic paused, and certain node-wide and cluster-wide processes disabled, you have a more stable environment for troubleshooting issues and restoring the cluster to a usable state. 
+
+== Requirements
```
JakeSCahill:
```suggestion
== Prerequisites
```

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,218 @@
+= Recovery Mode
+:description: Recovery mode starts Redpanda in "metadata only" mode and disables partitions so you can repair a failed cluster.
```
JakeSCahill: Do we need to explain metadata-only mode? Or, should we just omit it?

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+Recovery mode allows you to start Redpanda and form a stable cluster when it fails due to a system crash, OOM errors, or other issues that make a normal startup infeasible.
```
JakeSCahill:
```suggestion
Recovery mode allows you to repair and restore a failed cluster that cannot start normally due to issues such as a system crash or out-of-memory (OOM) errors. In recovery mode, Redpanda limits functionality to cluster configuration changes and other manual administrative actions so that you can repair the cluster. 
```

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,218 @@
+= Recovery Mode
+:description: Recovery mode starts Redpanda in "metadata only" mode and disables partitions so you can repair a failed cluster.
+
+Recovery mode allows you to start Redpanda and form a stable cluster when it fails due to a system crash, OOM errors, or other issues that make a normal startup infeasible. 
+
+In recovery mode, cluster configuration changes and other manual administrative actions are permitted so you can repair and restore the failed cluster. Topic reads and writes are disabled.
```
JakeSCahill: I think this paragraph should be removed. The suggestion above covers all the important points. And, this should be added to the list of disabled actions in 'About recovery mode':

> Topic reads and writes are disabled.

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+== About recovery mode
+
+Recovery mode starts a cluster with limited functionality and deactivates normal traffic and data operations for a cluster. Redpanda does not load user-managed partitions on disk. This helps prevent triggering a storm of partition leader elections and replication (such as increased learner replica recovery load after a long broker downtime) that may occur on startup. 
+
+Normal traffic and especially malicious traffic can disrupt availability for all users, including admin users. Recovery mode helps avoid placing the cluster under a heavy load of Kafka API and HTTP Proxy connections as you work to resuscitate the cluster.
```
JakeSCahill, line 25: What is 'normal traffic' here? This sounds like we disable connections to the Kafka API and HTTP Proxy. But no mention of Schema Registry.

Later on, we say:

> You can use the Admin API when the cluster is in recovery mode. The HTTP Proxy and Schema Registry APIs are not available...With normal Kafka traffic paused...

So all APIs are disabled apart from the Admin API?

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,218 @@
+= Recovery Mode
+:description: Recovery mode starts Redpanda in "metadata only" mode and disables partitions so you can repair a failed cluster.
+
+Recovery mode allows you to start Redpanda and form a stable cluster when it fails due to a system crash, OOM errors, or other issues that make a normal startup infeasible. 
+
+In recovery mode, cluster configuration changes and other manual administrative actions are permitted so you can repair and restore the failed cluster. Topic reads and writes are disabled.
+
+== About recovery mode
```
JakeSCahill: I think this whole section should be renamed. It's listing enabled/disabled operations/functionality. It'd be easier to scan if we had the headings 'Disabled functionality' and 'Enabled functionality'.

modules/manage/pages/recovery-mode.adoc
kbatuigas: @JakeSCahill Is something like "...starts Redpanda with limited functionality and disables partitions..." preferable?

modules/manage/pages/recovery-mode.adoc
kbatuigas: @JakeSCahill Yes exactly. Hm, this is how it was written in the PRD:

> Recovery mode also intends to avoid the load of Kafka API/HTTP Proxy connections, because normal traffic and especially abusive/rogue traffic can prevent availability for all users, even the admin trying to resuscitate the cluster.

I took it to mean that Kafka API and HTTP Proxy in particular can see heavy enough usage to be disruptive when trying to troubleshoot/fix a cluster. I'm not sure why SR isn't specifically mentioned there, but yes it is not available in recovery mode either.

ztlpn: >  all APIs are disabled apart from the Admin API?

Kafka API is only partially disabled (fetch and produce requests return non-recoverable errors). The reason for this is that some administrative actions (like deleting topics) that are needed in recovery mode are only available via Kafka API.

ztlpn (approved):

modules/manage/pages/recovery-mode.adoc
JakeSCahill: Thanks @ztlpn - @kbatuigas we should clarify this in the list of enabled and disabled actions

kbatuigas: @JakeSCahill @ztlpn added to "disabled" list. "Delete topics" is already in the enabled list -- should "Kafka API" be specified with it as well?

JakeSCahill: Looks great. I would flip the two so that users can see what's enabled first.

JakeSCahill: It would be nice to specify which actions are enabled for the Kafka API

ztlpn: Right, we should state that certain Kafka API is partially disabled and certain administrative requests are still available (deleting topics, modifying topic properties, editing consumer group metadata) - this is from the list of enabled operations, maybe grouping them by API type is a good idea.

modules/manage/pages/recovery-mode.adoc
```diff
@@ -0,0 +1,223 @@
...
+
+With normal Kafka traffic restricted, and certain node-wide and cluster-wide processes disabled, you have a more stable environment for troubleshooting issues and restoring the cluster to a usable state. 
+
+== Prerequisites
+
+To enable recovery mode, ensure that you have xref:get-started:intro-to-rpk.adoc[rpk] installed, and you have local access to machines running Redpanda.
+
+A broker can only enter recovery mode as it starts up, and not while it is already running. You first set the broker configuration property to enable recovery mode, and then do a broker restart.
+
+== Use recovery mode
```
JakeSCahill:
```suggestion
== Start Redpanda in recovery mode
```

JakeSCahill (approved):
LGTM. Just one suggestion to clarify the heading.

PR #115, opened 2023-11-15
docs: Add Kubernetes compatibility matrix

JakeSCahill:
Review deadline: November 17

Fixes redpanda-data/documentation-private#2109

Preview: https://deploy-preview-115--redpanda-docs-preview.netlify.app/current/upgrade/k-compatibility/

Notes:

- ~TODO: We need a way to add new rows to these tables automatically when new major or minor versions of the charts are released.~


modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+| Redpanda Operator Helm Chart | Redpanda Operator | Redpanda Helm Chart
+
+| 0.4.x
+| 2.0.x
+| 5.6.x
+|===
+
+== Redpanda Console compatibility
+
+Redpanda Console is included as a subchart in the Redpanda Helm chart. The version of Redpanda Console linked in the Redpanda Helm chart is established when users add or update the Redpanda chart repository. This process means that the included Redpanda Console version can change with each update to the local chart repository.
```
alejandroEsc: The versions we support are defined to be 
```
    version: ">=0.5 <1.0"

```

and for connectors 

```
    version: ">=0.1.2 <1.0"

```

which means to update the version of console you may have to run `helm repo update` in your machine before deploying the chart. That being said, the operator tries to use the latest as it pull in chart data.

alejandroEsc (approved):
lgtm, just one note, may be useful or not, ill let you decide.

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
+= Kubernetes Compatibility
+:description: Find a compatibility matrix for versions of Redpanda, Redpanda Helm chart, and Redpanda Operator.
+
+This topic describes the compatibility for versions of the Redpanda software, Redpanda Kubernetes Operator, and the Helm charts. The tables identify the versions required and supported for running the specified version of Redpanda.
```
mattschumpert: @JakeSCahill let's use just 'Redpanda' when referring to the redpanda binary. It's all software.

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+include::partial$versioning.adoc[]
+
+The latest version of the Redpanda Helm chart supports not only the most recent Redpanda release but also the previous two feature releases, subject to these considerations:
+
+- It supports the two preceding feature releases of the same year.
+- In cases where any of these two preceding feature releases are from the preceding year, the Helm chart supports the most recent feature release of that year.
```
mattschumpert: @JakeSCahill @alejandroEsc No.  We support the 3 latest feature releases (current and 2 preceding releases) REGARDLESS of the calendar year.  The alternative does not make sense , is not aligned with our lifecycle policy and would be randomly unfair.

Please remove this statement.

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+- Support is conditional on the preceding releases still being actively supported. In cases where any of the preceding releases are no longer supported, the Helm chart does not support them.
+
+== Redpanda Helm chart compatibility
+
+The Redpanda Helm chart is tested against these version of:
+
+- Redpanda software
```
mattschumpert: 'Redpanda'

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+The latest version of the Redpanda Helm chart supports not only the most recent Redpanda release but also the previous two feature releases, subject to these considerations:
+
+- It supports the two preceding feature releases of the same year.
+- In cases where any of these two preceding feature releases are from the preceding year, the Helm chart supports the most recent feature release of that year.
+- Support is conditional on the preceding releases still being actively supported. In cases where any of the preceding releases are no longer supported, the Helm chart does not support them.
+
+== Redpanda Helm chart compatibility
+
+The Redpanda Helm chart is tested against these version of:
```
mattschumpert: 'versions'

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+- Redpanda software
+- Kubernetes and Helm
+
+[cols="4*m"]
+|===
+| Redpanda Helm Chart |Redpanda|Kubernetes|Helm
+
+| 5.6.x
+| 23.2.x, 23.1.x
```
mattschumpert: This doesn't seem to show 3 feature versions, which is the whole point .

@alejandroEsc  again we are supporting 22.3 through Dec 15th.  We need to list that here (and continue testing it) or it undermines the whole point above.

mattschumpert: Also @alejandroEsc my understanding is we also support the 3 most recent versions of K8s itself (e.g. for self-hosted customers). This is what I had heard from Joe.  I don't think we can say we support only 1 k8s version, as of course people have to do upgrades across 2 versions.

What's the current testing state and what would be the 3 k8s versions we're comfortable with as of now?

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+|1.27.3
+|3.11.1
+
+|===
+
+By default, the Redpanda Helm chart depends on cert-manager for enabling TLS.
+Make sure to install a version of the CRDs that supports the `certificates.cert-manager.io/v1` resource.
```
mattschumpert: @JakeSCahill, @alejandroEsc  it seems quite odd to have a compat matrix that then asks the user to ensure the CRD version matches something else, without giving that information here. 

I assume you mean OUR CRDs with this statement.  As a user the immediate question is : well, what versions of Redpanda K8s stuff support which versions of cert-manager?  And moreover, if there are Redpanda CRD versions that don't support certmanager v1 and by assumption not  at all, what are they?

We need to clean this up and simplify the message here. What is the call to action? Is there one?  if there's some limitartion we need to spell it all out here.  The 'matrix' is supposed to be THIS page.

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+See the https://github.com/cert-manager/cert-manager/releases[cert-manager release notes].
+
+== Redpanda Operator compatibility
+
+Each version of the Redpanda Operator Helm chart deploys a specific version of the Redpanda Operator software and the Redpanda Helm chart. The Redpanda Operator software is tied to the same version of the custom resource definitions (CRDs).
```
mattschumpert: @JakeSCahill suggest we drop 'software' here too, in reference to the Operator.

mattschumpert: @alejandroEsc is "Redpanda Operator Helm Chart' the correct term, or is it just 'Redpanda Helm Chart' (what we have previously been using). Consider that some customers are still just using the helm chart with NO operator

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+The CRDs define the schemas used to configure Redpanda clusters. Each set of CRDs supports specific versions of the Redpanda Helm chart because they are used to translate the `clusterSpec` field of Redpanda resources to the values in the Redpanda Helm chart.
```
mattschumpert: @JakeSCahill 'seems like implementation detail.   The point is each Operator version (CRDs version?) supports specific versions of the helm chart.

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+Although new releases of the Redpanda Operator are backward compatible with existing versions of the Redpanda Helm chart, new releases of the Redpanda Helm chart might introduce changes to the Helm values, which could require you to upgrade both the Redpanda Operator and the accompanying CRDs.
```
mattschumpert, line 29: @JakeSCahill @alejandroEsc ok this is helpful but .......so the conclusion/ best practice here is to _not_ upgrade the helm chart once you're using the Operator and to always use the latest version of the Redpanda Operator, which will in turn upgrade the helm chart?  

Let's spell out what we want people to do.

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+[cols="3*m"]
+|===
+| Redpanda Operator Helm Chart | Redpanda Operator | Redpanda Helm Chart
+
+| 0.4.x
```
mattschumpert: I'm ignorant about the 2 different charts here @alejandroEsc I thought it was one Redpanda Helm chart and the whole idea was you could migrate from Helm-only to operator because the chart was one and the same.....  but perhaps I'm reminded the migration is a bunch of copy paste and they're two things?  But if that's the case why would you need both to work together?
 
I'm missing something, but we definitely need to work on the naming here.

mattschumpert: @JakeSCahill why do we not have a whole table with all the recent versions (is it because this is still in draft)?

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+- Kubernetes and Helm
+
+[cols="4*m"]
+|===
+| Redpanda Helm Chart |Redpanda|Kubernetes|Helm
+
+| 5.6.x
+| 23.2.x, 23.1.x
+|1.27.3
+|3.11.1
```
mattschumpert: @JakeSCahill why do we not have a whole table with all the recent versions (is it because this is still in draft)?

modules/upgrade/pages/k-compatibility.adoc
```diff
@@ -0,0 +1,71 @@
...
+| 0.4.x
+| 2.0.x
+| 5.6.x
+|===
+
+== Redpanda Console compatibility
+
+Redpanda Console is included as a subchart in the Redpanda Helm chart. The version of Redpanda Console linked in the Redpanda Helm chart is established when users add or update the Redpanda chart repository. This process means that the included Redpanda Console version can change with each update to the local chart repository.
+
+Redpanda Console is backwards compatible with all versions of Redpanda.
```
mattschumpert: @JakeSCahill  This is good but it kinda starts to address overall platform compat matrix and not just k8s (perhaps this page should eventually graduate to that, where we address RPK and all the components)..

modules/upgrade/pages/k-compatibility.adoc
mattschumpert: versions we support in what version of what @alejandroEsc 

Again here @JakeSCahill  we need the whole table.

If in fact this is all keyed off the Redpanda Helm chart itself (the text is a little cryptic but I think that's what it means), then the Redpanda helm chart should be the first column the table (because IIUC thats the determining factor that drives everything?). It sounds like there's a 1:1 mapping between the Redpanda chart and the Console chart, but that you have some flex in Console versions. Is that right @alejandroEsc ?

modules/upgrade/pages/k-compatibility.adoc
Feediver1:
```suggestion
The Redpanda Helm chart is tested against these versions of:
```

Feediver1 (approved):
lgtm

mattschumpert (changes requested):

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: This is talking about cert-manager CRDs. Will clarify.

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: This is under the heading `Redpanda Operator compatibility` as it applies to users of the operator. It is the Redpanda Operator Helm chart, which is here: https://github.com/redpanda-data/helm-charts/tree/main/charts/operator

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: So something like this?

Consider the following best practices for upgrades:

- **Prioritize updating the Redpanda Operator**: Always use the latest version of the Redpanda Operator. Newer versions are backward compatible with existing versions of the Redpanda Helm chart, ensuring stability and access to the latest features and fixes.

- **Be cautious with Redpanda Helm chart upgrades**: New releases of the Redpanda Helm chart might introduce changes to the Helm values, which could require a corresponding upgrade of both the Redpanda Operator and the CRDs to ensure compatibility.

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: @mattschumpert these are the versions we currently test: https://github.com/redpanda-data/helm-charts/blob/479a2720c28af8f3e279eb323ceb1db1be3d56b5/.github/workflows/nightly.yaml#L51

It doesn't include 22.3

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: @RafalKorepta or @alejandroEsc WDYT?

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: @mattschumpert we don't have access to previous chart versions. The idea was to start from this version and automate updating these tables on each new major/minor version: https://github.com/redpanda-data/helm-charts/blob/main/charts/operator/Chart.yaml

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: The Helm chart repo only provides us with the current version. I could provide all recent versions but I'd need those from @RafalKorepta or @alejandroEsc

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: Agreed. It would be nice to have that too, but should be addressed in a separate issue. I added this detail about Console so that users know they don't need to worry so much about Console compatibility.

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: Working on the automation which will fill in the table

JakeSCahill: I need a way to determine which versions of the RP Helm chart each version of the RP Operator supports @RafalKorepta

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: I've added the whole table now which is autogenerated @mattschumpert 

The charts define a minimum required version of Helm/K8s so I'm using that reference for the matrix.

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: I have autogenerated the table using the console chart versions defined in the RP charts.

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: All charts publish the minimum required Kubernetes version. I have added a step to fetch this in our automation. It's currently ^1.21.0-0 for all versions.

modules/upgrade/pages/k-compatibility.adoc
JakeSCahill: @mattschumpert We don't currently have a strategy to define the versions of the RP Helm chart that operator supports.

We should discuss this so that the docs can reflect our strategy.

JakeSCahill: @alejandroEsc agreed we can say the latest n - 2 releases of the RP Helm chart.

modules/upgrade/partials/k-redpanda-chart-compatibility-matrix.adoc
```diff
@@ -0,0 +1,104 @@
+|===
+| Redpanda Helm Chart |Supported Redpanda Versions|Minimum Kubernetes Version|Minimum Helm Version
```
mattschumpert, line 2: @JakeSCahill shouldn't we at least recommend something higher than 1.21 here though (even if technically the chart is configured for it).

@alejandroEsc what should that recommendation be?

modules/upgrade/partials/k-operator-compatibility-matrix.adoc
```diff
@@ -0,0 +1,14 @@
+|===
+| Redpanda Operator Helm Chart | Redpanda Operator|CRD|Support Redpanda Helm chart
```
mattschumpert, line 2: @JakeSCahill 
Shouldn't the Operator version be in the first column?

WDYT @alejandroEsc ?

mattschumpert: Also @JakeSCahill  shouldn't this be 'Supported Redpanda Helm chart' not 'Support'

mattschumpert (approved):

modules/upgrade/partials/k-redpanda-chart-compatibility-matrix.adoc
alejandroEsc: The PDB v1beta became deprecated by 1.21+ and became unavailable at 1.25. Note that that the latest k8s release is 1.28 so i think we should support 1.25 and up and we should probably aim at supporting n-2 of releases. What we need to do is establish a rquirement company wide. What do we do for cloud? Also i am aware that certain providers cannot guarantee n-2 but we should probably support 1.25 onward for now.

alejandroEsc (approved):
lgtm

modules/upgrade/partials/k-redpanda-chart-compatibility-matrix.adoc
JakeSCahill: Added a note to recommend 1.25 as a minimum.

PR #118, opened 2023-11-16
Whole cluster restore

kbatuigas:
Preview: https://deploy-preview-118--redpanda-docs-preview.netlify.app/23.3/manage/whole-cluster-restore/ (Linux)

This content will also be single sourced for Kubernetes

Closes https://github.com/redpanda-data/documentation-private/issues/1978


modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
 
 == Disaster recovery
 
+With tiered storage enabled, you can restore your cluster, including its metadata, in the event of failure. This is a faster, less expensive alternative to active-active replication (for example, using xref:upgrade:migrate/data-migration.adoc[MirrorMaker 2]).
```
mattschumpert: @andrwng  I don't think we can claim 'faster' given MM2 is continuous replication.  I think we can say 'simpler, less expensive'.  Fair?

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+
+You can take the following steps to restore a failed cluster:
+
+. Start a new cluster
+. Recover metadata from a failed source cluster to the new cluster
```
mattschumpert: With the new command these 2 steps are actually one step , right @andrwng ?  So I think we can combine them.

ALso @kbatuigas  we're using the word 'restore' now instead of recovery, to eliminate confusion with recovery mode.

So everywhere we should be saying "whole cluster restore" and "restore"

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+. Recover topic data from the source cluster to the new cluster
+. Perform a rolling restart, if necessary
+. Direct application workload to the new cluster
+
+Use this recovery method to restore your application to the latest functional state as quickly as possible. Note that this method does not guarantee consistency at the application level. Restored transactions may not uphold atomicity. The consumer offset topic may not reflect point-in-time snapshots of topic data already consumed by applications. 
+
+=== Start a recovery target cluster
```
mattschumpert: lets just say 'target cluster' (to get away from 'recovery')

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+
+You can take the following steps to restore a failed cluster:
+
+. Start a new cluster
+. Recover metadata from a failed source cluster to the new cluster
+. Recover topic data from the source cluster to the new cluster
+. Perform a rolling restart, if necessary
+. Direct application workload to the new cluster
+
+Use this recovery method to restore your application to the latest functional state as quickly as possible. Note that this method does not guarantee consistency at the application level. Restored transactions may not uphold atomicity. The consumer offset topic may not reflect point-in-time snapshots of topic data already consumed by applications.
```
mattschumpert: @andrwng  shouldn't we add.... 'though it will reflect the latest available offset in the topic data being restored'. (meaning the consumer offsets topic here)

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+. Recover metadata from a failed source cluster to the new cluster
+. Recover topic data from the source cluster to the new cluster
+. Perform a rolling restart, if necessary
+. Direct application workload to the new cluster
+
+Use this recovery method to restore your application to the latest functional state as quickly as possible. Note that this method does not guarantee consistency at the application level. Restored transactions may not uphold atomicity. The consumer offset topic may not reflect point-in-time snapshots of topic data already consumed by applications. 
+
+=== Start a recovery target cluster
+
+Follow the steps to xref:deploy:deployment-option/cloud/index.adoc[provision a new cloud cluster]. Cloud account credentials (such as secrets or access keys) are not included in backed up metadata, and so you must manually configure these credentials for the new cluster.
```
mattschumpert: @kbatuigas this should not be assuming Cloud.  We should say something like "if using Redpanda Cloud, follow this doc.  If using Redpanda Enterprise, follow this doc (self-hosted deploy docs)"

mattschumpert: @andrwng should we link to HOW to configure cloud credentials as cluster properties?  And also that if using IAM roles this isn't necessary?

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+
+=== Cluster metadata recovery
```
mattschumpert: restore

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+Metadata from a failed cluster can be restored to a target cluster using an rpk command. 
+
+[,bash]
+----
+rpk 
+----
+
+The following metadata is included in the restore:
+
+- Topic definitions. If you have enabled tiered storage only for specific topics, non-tiered storage will not be recovered
```
mattschumpert: at the end: "topics without tiered storage enabled will not be recovered"

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+----
+rpk 
+----
+
+The following metadata is included in the restore:
+
+- Topic definitions. If you have enabled tiered storage only for specific topics, non-tiered storage will not be recovered 
+- Users and access control lists (ACLs)
+- Schemas
+- Consumer offsets
```
mattschumpert: should we clarify 'including the  __consumer_offsets topic" @andrwng ? how best to say this?

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+----
+
+The following metadata is included in the restore:
+
+- Topic definitions. If you have enabled tiered storage only for specific topics, non-tiered storage will not be recovered 
+- Users and access control lists (ACLs)
+- Schemas
+- Consumer offsets 
+- All cluster configurations, including your Redpanda license key
+- Transaction metadata, up to the highest committed transaction. In-flight transactions are treated as aborted and will not be included in recovery
```
mattschumpert: @andrwng just curious for myself this is just historical transactions (e.g. for audit purposes but otherwise not useful)?

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+
+NOTE: Committed transactions are not guaranteed to be atomic.
```
mattschumpert: @kbatuigas it almost feels like all 5 of these statements are bullet points to a note.  They're all gotchas we want to alert people of

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+
+The destination cluster must be empty; that is, the cluster cannot contain data other than node and cluster configuration properties. If metadata indicating usage is found in the destination (such as user-provided schemas, non-superusers, ACLs, consumer offsets, or user-created topics), the restore will fail and the rpk command will issue a warning.
+
+If duplicate metadata is found in the destination cluster, it will be overwritten by the restore.
```
mattschumpert: @andrwng to be clear here we only mean 'duplicate cluster configuration', right?

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
+
+Metadata recovery supports only one source cluster. It is not possible to consolidate multiple clusters into the target cluster. 
+
+=== Topic data recovery 
+
 When you create a topic, you can use remote recovery to download the topic data from object storage. You can also use remote recovery to restore a topic that was deleted from a cluster.
```
mattschumpert: @andrwng this seems like 2 ways of saying the same thing, no?  Maybe just "remote recovery downloads all topic data from object storage into local storage, which is useful if a single topic was accidentally deleted from a cluster".

BTw, is this true?  ARe we still actually downloading topic data to local storage or only downloading metadata (for single topic recovery)?

cc @kbatuigas

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,6 +1250,50 @@ endif::[]
...
 
 CAUTION: Remote recovery is not intended to be a fully-functional disaster recovery solution. It is only safe when no other clusters are still writing to the specified bucket or container. Recovery does not provide snapshot-style consistency: some partitions in some topics will be more up to date than others. For example, restoring a consumer offsets topic could have offsets that point ahead of the restored data.
```
mattschumpert: @kbatuigas we need to move this warning up to the top of the new Whole-Cluster Restore  section but remove the last sentence which is no longer true / not applicable.  

(Now that we have whole cluster restore this becomes the closest thing to DR but still has some o the caveats)

cc @andrwng

modules/manage/partials/tiered-storage.adoc
mattschumpert: Actually this part is kind of a duplicate to the blue info note further down (transaction atomicity etc) @kbatuigas perhaps we can find a way to combine

kbatuigas:
@andrwng Other questions (just to make sure we don't lose them):
- Are we able to provide specific steps or things for users to consider in order to determine if and when they should go the cluster metadata recovery route? 
- Do we also have steps and other information specific to Kubernetes deployments?

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1268,6 +1312,13 @@ To create a new topic using remote recovery and enable Tiered Storage on the new
 rpk topic create <topic_name> -c redpanda.remote.recovery=true -c redpanda.remote.write=true -c redpanda.remote.read=true
 ----
 
+=== Performing rolling restart
```
kbatuigas: @andrwng Is the content on [this page](https://docs.redpanda.com/current/upgrade/rolling-upgrade/#perform-a-rolling-upgrade) the same as a "rolling restart"? How do we know when this step is required as part of recovery?

modules/manage/partials/tiered-storage.adoc
kbatuigas: @mattschumpert @andrwng We also have [this doc](https://docs.redpanda.com/current/develop/consume-data/consumer-offsets/), let me know if I should just link to it here.

modules/manage/partials/tiered-storage.adoc
kbatuigas: @mattschumpert @andrwng Would there be a case where a user might want to do remote topic recovery even if a cluster hasn't failed? If so, should we keep "It is only safe when no other clusters are still writing to the specified bucket or container" as a caveat for only that scenario?

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1268,6 +1312,13 @@ To create a new topic using remote recovery and enable Tiered Storage on the new
...
+
+
+=== Redirect workload to new cluster
```
kbatuigas: @andrwng Are there specific steps to take here or a doc we could link to?

andrwng: I don't think there are docs, and this is fairly application specific. It amounts to changing their applications to use the addresses of the new cluster

andrwng: I'm not aware of any. It's pretty application specific and amounts to changing their application code to use the addresses of the new cluster

modules/manage/partials/tiered-storage.adoc
andrwng: Yeah, but without the "Upgrade your version" step

modules/manage/partials/tiered-storage.adoc
andrwng: Yea this isn't quite true anymore (not as a part of this effort, but I think these docs just never were updated)

There are some caveats, but we generally only download archival metadata, which allows the broker to serve cloud data segments on demand.

modules/manage/partials/tiered-storage.adoc
andrwng: Hmm, yeah I'm not quite sure what this is referring to. @kbatuigas could you clarify? Or maybe remove

modules/manage/partials/tiered-storage.adoc
andrwng: Hmm, yeah I'm not quite sure what this is referring to. @kbatuigas could you clarify? Or maybe remove

modules/manage/partials/tiered-storage.adoc
andrwng: Yeah, transactional metadata is baked into the storage layer and the client protocol. The restored metadata will fence off older transactions so they don't interfere with new transactions

modules/manage/partials/tiered-storage.adoc
andrwng: >'including the __consumer_offsets topic" @andrwng ? how best to say this?

Yeah perhaps we can link that doc, and refer to it as the "consumer offsets topic"

modules/manage/partials/tiered-storage.adoc
andrwng: Yeah, perhaps https://docs.redpanda.com/current/manage/tiered-storage/#set-up-tiered-storage ?

modules/manage/partials/tiered-storage.adoc
andrwng: Right, the rpk command will restore both data and metadata

modules/manage/partials/tiered-storage.adoc
andrwng: "Less expensive" definitely. I'm on the fence about "simpler" -- conceptually duplicating data to two clusters is pretty simple

modules/manage/partials/tiered-storage.adoc
andrwng: >And also that if using IAM roles this isn't necessary?

@mattschumpert actually I'm confused by this, can you clarify? What changes if IAM roles are configured? The recovery right now doesn't restore cloud storage credentials and configs, with the assumption that the cluster is already able to access the cloud.

This makes me think we should be restoring the [cloud_storage_credentials_source](https://docs.redpanda.com/current/reference/cluster-properties/#cloud-storage) property (we aren't now, but I can update that). Is that right?

modules/manage/partials/tiered-storage.adoc
kbatuigas: @mattschumpert @andrwng yes, I lifted it from the [PRD](https://docs.google.com/document/d/1840e92oD8CPifNqCMt7KN9Z2PH5Ej-mceSJN9WS3Olc/edit#heading=h.7i0ls36lt6ow) but happy to remove if appropriate

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,9 +1250,257 @@ endif::[]
...
+For Redpanda Cloud, follow the steps to xref:deploy:deployment-option/cloud/index.adoc[provision a new cloud cluster]. 
+Cloud account credentials (such as secrets or access keys) are not included in backed up metadata. You must manually configure these credentials for the new cluster.
+
+=== Whole cluster restore
+
+Data from a failed cluster can be restored to a target cluster using an rpk command. 
+
+[,bash]
+----
+rpk cluster storage restore start
```
kbatuigas: @andrwng Are there required parameters? Does the command need to specify the source and destination clusters?

Also, should the `stop` command be mentioned at any point in this doc? Would that go under the [post-restart section](https://github.com/redpanda-data/docs/pull/118/files#diff-8e053dde88b9dfc92939a73598f0d06dd718f88b36c08f136745e3cf97bcb061R1483) below?

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,9 +1250,257 @@ endif::[]
...
+
+NOTE: If you need to restore only a subset of topic data, consider using <<remote-recovery,remote recovery>> instead of a whole cluster restore.
+
+The destination cluster must be empty; that is, the cluster cannot contain data other than node and cluster configuration properties. If metadata indicating usage is found in the destination (such as user-provided schemas, non-superusers, ACLs, consumer offsets, or user-created topics), the restore will fail and the rpk command will issue a warning.
+
+If duplicate metadata is found in the destination cluster, it will be overwritten by the restore.
+
+Whole cluster restore supports only one source cluster. It is not possible to consolidate multiple clusters into the target cluster.
+
+=== Perform rolling restart
```
andrwng: This content isn't specific to tiered storage or disaster recovery, so it feels like it belongs in its own section, perhaps in the Cluster Maintenance section of the docs? Then we can just link to it from here

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,9 +1250,257 @@ endif::[]
...
+Data from a failed cluster can be restored to a target cluster using an rpk command. 
+
+[,bash]
+----
+rpk cluster storage restore start
+----
+
+The following metadata is included in the restore:
+
+- Topic definitions. If you have enabled tiered storage only for specific topics, topics without tiered storage enabled will not be restored
```
andrwng: > topics without tiered storage enabled will not be restored

In the current implementation, these topics are actually restored as empty.

@mattschumpert we chatted about this briefly but I don't think we concluded anything. Just want to make sure this is desired behavior (it's easy to fix).

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,9 +1250,257 @@ endif::[]
...
+[,bash]
+----
+rpk cluster storage restore start
+----
+
+The following metadata is included in the restore:
+
+- Topic definitions. If you have enabled tiered storage only for specific topics, topics without tiered storage enabled will not be restored 
+- Users and access control lists (ACLs)
+- Schemas
```
andrwng: Maybe link to https://docs.redpanda.com/current/manage/schema-reg/schema-reg-overview/?

Also it's a technicality, but there isn't anything particularly special about schemas from a restoration perspective -- the schemas topic is just a normal topic.

We should note that if a user wants their schemas to be archived and restored, they should configure the `_schemas` topic to be enabled with tiered storage

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,9 +1250,257 @@ endif::[]
...
+- The xref:develop:consume-data/consumer-offsets.adoc[consumer offsets topic]
+- All cluster configurations, including your Redpanda license key
```
andrwng: This isn't quite true, the list of ignored properties is here

https://github.com/redpanda-data/redpanda/blob/494689cfd1703cedc8c84d9bfbbbc203a8986025/src/v/cluster/cluster_recovery_reconciler.cc#L33-L48

The general theme of them is that we don't restore any configs that change the way the cluster interacts with tiered storage, since the cluster is already meant to be configured with tiered storage

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1250,9 +1250,257 @@ endif::[]
...
+- Transaction metadata, up to the highest committed transaction. In-flight transactions are treated as aborted and will not be included in the restore
+
+NOTE: If you need to restore only a subset of topic data, consider using <<remote-recovery,remote recovery>> instead of a whole cluster restore.
+
+The destination cluster must be empty; that is, the cluster cannot contain data other than node and cluster configuration properties. If metadata indicating usage is found in the destination (such as user-provided schemas, non-superusers, ACLs, consumer offsets, or user-created topics), the restore will fail and the rpk command will issue a warning.
```
andrwng: >If metadata indicating usage is found in the destination (such as user-provided schemas, non-superusers, ACLs, consumer offsets, or user-created topics), the restore will fail and the rpk command will issue a warning

@mattschumpert in the current implementation this isn't true. I opted towards making the operation retriable, meaning e.g. it's fine if the topic already exists, it just won't be restored and the restore will move on rather than failing. I can make the change to be more strict if that behavior is more intuitive.

modules/manage/partials/tiered-storage.adoc
andrwng: >@andrwng Are there required parameters? Does the command need to specify the source and destination clusters?

The main one is `-w`. I haven't implemented support for `--topic-name-pattern` but I think it's a bit simpler to leave it out for now.

@mattschumpert given this is whole cluster restore, should we just remove the topic pattern?

modules/manage/partials/tiered-storage.adoc
kbatuigas: This sounds good to me actually, any thoughts @mattschumpert ? I'd create a separate issue and PR for this, and it would be a blocker for this PR cause I'd need to merge that new one first before I can incorporate the new section in this PR. But it shouldn't take as long.

modules/manage/partials/tiered-storage.adoc
mattschumpert: @andrwng I think that's right.  When using IAM roles, if the credential source is set to IAM roles and the role has been configured you don't need to set credentials on the new cluster.

modules/manage/partials/tiered-storage.adoc
mattschumpert: @andrwng should we say 'duplicate cluster configuration' or not?  (duplicate config props will be overridden by restore, right)?

modules/manage/partials/tiered-storage.adoc
mattschumpert: @kbatuigas yes one might want to do remote recovery of individual **topics** even if a cluster hasn't failed (e.g. deleted data accidentally). But this is a different RPK command.

However the 'whole cluster recovery' (rpk cluster storage recovery) is different, and the last sentence is  incorrect as we implemented avoiding that consumer offsets problem in the feature.

modules/manage/partials/tiered-storage.adoc
mattschumpert: @andrwng im not sure why we need the topic pattern with cluster storage restore.

modules/manage/partials/tiered-storage.adoc
mattschumpert: Do you mean the Rolling restart section?  I have no idea why a rolling restart is being recommended in this doc.  It seems misplaced, right @andrwng ?

modules/manage/partials/tiered-storage.adoc
mattschumpert: right it is @andrwng . @kbatuigas we should clarify (again specifically for WHOLE cluster recovery), non-TS topics will be restored as empty

modules/manage/partials/tiered-storage.adoc
mattschumpert: @andrwng what happens with users, ACLs, offsets that already exist?

modules/manage/partials/tiered-storage.adoc
andrwng: Right, the cluster configuration is overridden by restore

modules/manage/partials/tiered-storage.adoc
andrwng: It's sometimes warranted (e.g. to ensure the cluster reflects newly restored configs)

modules/manage/partials/tiered-storage.adoc
andrwng: - Users that already exist aren't created again
- ACL bindings I believe are additive so repeated application is idempotent
- Offsets is a good question: right now it would override the existing offsets, which seems less palatable if users have already committed offsets with an existing group. An alternative would be to ignore the consumer group if it's already exists. Or we can fail the operation entirely -- I'm open to this, but just exploring of ways to make the restore as low-touch as possible, if possible.

kbatuigas: @andrwng in the current implementation will the restore still be considered "complete" when there's existing metadata and it gets overridden? Will the command issue some output or indication on what was overridden or ignored? And by "retriable" do you mean manually by the user, or does the process itself automatically retry?

modules/manage/partials/tiered-storage.adoc
andrwng: @mattschumpert just to make sure I understand, it's right to restore the `cloud_storage_credentials_source` or to leave it as is and not restore?

If IAM roles aren't configured before restoring, the cluster wouldn't be able to access the bucket and restore, right? I.e. isn't proper tiered storage credentials a prereq to running recovery? Or do most users specify an access key up front and then add to the IAM roles later?

modules/manage/partials/tiered-storage.adoc
andrwng: @kbatuigas

>the current implementation will the restore still be considered "complete" when there's existing metadata and it gets overridden?

The restore will be considered complete and metadata will be overwritten (in the case of offsets and configs) or added to the cluster (in all other cases).

>Will the command issue some output or indication on what was overridden or ignored? 

Not at the moment. The ways to examine updated metadata would be to use `rpk` to view the topics, cluster config, group metadata.

>And by "retriable" do you mean manually by the user, or does the process itself automatically retry?

I mean manually by a user: if for whatever reason the operation stops part way, it seems desirable to be able to run it again (rather than deleting everything from the cluster and then retrying)

andrwng: @mattschumpert any thoughts on this? https://github.com/redpanda-data/docs/pull/118#discussion_r1410170175 Happy to make the change either way.

modules/manage/partials/tiered-storage.adoc
andrwng: I chatted a bit with @abhijat and I think the original stance is sound: cloud credentials (whether that be IAM or keys) should be configured before restoring. If this isn't configured in the first place, a restore wouldn't be possible because we'd be unable to access the bucket.

modules/manage/partials/tiered-storage.adoc
andrwng: >An alternative would be to ignore the consumer group if it's already exists.

Discussed with @mattschumpert and I will make a change before the release. In this way, most metadata updates will be additive and effectively idempotent (though cluster config restore will overwrite existing cluster configs).

Let's still document that we strongly recommend that the starting state of the cluster before restoring _should_ be empty (no topics or anything, no on-going produces, consumes, commits, transactions, etc).

kbatuigas:
Alias `rpk cluster storage restore` in this [docs PR](https://github.com/redpanda-data/docs/pull/162/files#diff-8fbf23782c2a82c4ab95ba4ff5fdcf7fc0361216957bec130c1e0df2855b340b)

mattschumpert:
@kbatuigas we should document ONLY restore (not recovery) in the main DR section of 23.3 docs.  'rpk cluster storage recovery' is there only for backwards compat purposes (can still be in rpk reference as deprecated)

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1248,11 +1248,83 @@ IMPORTANT: Do not set an object storage property to an empty string `""` or to `
 
 endif::[]
 
-== Disaster recovery
+== Whole cluster restore for disaster recovery
```
andrwng: I think it's probably worth mentioning the flip side of metadata recovery, i.e. that Redpanda will upload metadata periodically, in addition to topic data. I'm not sure if this is the right section to do it, but there are some cluster configs worth bringing up in that regard:

- `enable_cluster_metadata_upload_loop`: enables metadata uploads, required for whole cluster recovery to be performed.
- `cloud_storage_cluster_metadata_upload_interval_ms`: time interval to wait between cluster metadata uploads.
- `controller_snapshot_max_age_sec`: this isn't necessarily directly related to uploads of metadata, but it does affect how fresh the uploaded metadata can be.

There's also the `cloud_storage_attempt_cluster_recovery_on_bootstrap` config option that is mostly there for the purposes of automating recovery in k8s (cc @JakeSCahill when you're back), but it's still probably worth describing the behavior here.

Perhaps the order of this section could be something like
- description of metadata that is relevant to whole cluster restore
- configs relevant to uploading
- steps to restore

WDYT?

andrwng (commented):
I think the existing content looks pretty good, just a suggestion on some configs and ordering.

modules/manage/partials/tiered-storage.adoc
mattschumpert: Absolutely we need to mention this

modules/reference/pages/cluster-properties.adoc
kbatuigas, line 1: @JakeSCahill Do we have a separate page where we can add `cloud_storage_attempt_cluster_recovery_on_bootstrap` as Andrew [mentioned](https://github.com/redpanda-data/docs/pull/118#discussion_r1419830714), or do they go on this page as well?

modules/manage/partials/tiered-storage.adoc
JakeSCahill: @andrwng what does `cloud_storage_attempt_cluster_recovery_on_bootstrap` do exactly? And is it K8s only? Sounds like it's not, but not sure on the steps for using this config.

modules/reference/pages/cluster-properties.adoc
JakeSCahill: This whole page is single-sourced. Anything that isn't wrapped in a conditional will appear in both the Linux page and the K8s page.

I'll add the K8s content to this PR.

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,176 @@
...
+ifndef::env-kubernetes[]
+For self-hosted, follow the steps to xref:deploy:deployment-option/self-hosted/manual/index.adoc[deploy a new cluster].
+
+For Redpanda Cloud, follow the steps to xref:deploy:deployment-option/cloud/index.adoc[provision a new cloud cluster].
+Cloud account credentials (such as secrets or access keys) are not included in backed up metadata. You must manually configure these credentials for the new cluster.
+
+NOTE: Make sure to configure the target cluster with the same Tiered Storage settings as the failed source cluster.
+endif::[]
+
+=== Manage metadata uploads
```
andrwng: IMO this section is better suited to be perhaps the first section, right after describing the various metadata. It's not a part of the sequence of steps to restore, but rather a prerequisite to have enabled in the previous cluster in the first place.

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,176 @@
...
+** `cloud_storage_access_key`
+** `cloud_storage_secret_key`
+** `cloud_storage_region`
+** `cloud_storage_bucket`
+** `cloud_storage_api_endpoint`
+** `cloud_storage_credentials_source`
+** `cloud_storage_trust_file`
+** `cloud_storage_backend`
+** `cloud_storage_credentials_host`
+** `cloud_storage_azure_storage_account``
```
andrwng: nit: extra `

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,176 @@
...
+CAUTION: Whole cluster restore is not a fully-functional disaster recovery solution. It does not provide snapshot-style consistency. Some partitions in some topics will be more up-to-date than others. Committed transactions are not guaranteed to be atomic.
+
+NOTE: If you need to restore only a subset of topic data, consider using xref:manage:tiered-storage.adoc#remote-recovery[remote recovery] instead of a whole cluster restore.
+
+The following metadata is included in a whole cluster restore:
+
+* Topic definitions. If you have enabled Tiered Storage only for specific topics, topics without Tiered Storage enabled will be restored empty.
+* Users and access control lists (ACLs).
+* xref:manage:schema-reg/schema-reg-overview.adoc[Schemas]. To ensure that your schemas are also archived and restored, you must also enable Tiered Storage for the `_schemas` topic.
+* The xref:develop:consume-data/consumer-offsets.adoc[consumer offsets topic].
```
andrwng: It's somewhat implied from the "CAUTION" blurb above, but it may also be worth mentioning somewhere (maybe here?):

Upon restoring, because we don't provide snapshot style consistency across all partitions, some restored committed consumer offsets may be truncated to a lower value than what was in the original cluster, in order to keep offsets at or below the partitions' restored high watermarks.

modules/manage/partials/whole-cluster-restore.adoc
kbatuigas: @andrwng @JakeSCahill I ended up moving this per Andrew's suggestion, and made a change to the heading as well. Feel free to update if it makes more sense to stick to "Manage metadata uploads" or even change it entirely

JakeSCahill: @andrwng so should this section be part of the Tiered Storage topic so that users can know when configuring TS that if they ever need to do a cluster restore, they should configure these settings?

Then in the prereqs of the cluster restore topic, we can link to that section?

andrwng: Hmm good question. I wouldn't be against it, though it'd be worth mentioning the kind of metadata that is uploaded. The above list is fine, but there are a few details more relevant to the restore in it that ought to be trimmed down if we duplicate that info in another section.

kbatuigas: @JakeSCahill @andrwng Making sure I'm on the same page: isn't whole cluster restore already available to TS users without having to manually preconfigure these three settings? Since `enable_cluster_metadata_upload_loop` is set to `true` by default, and the other two properties are set with default values as well, then those properties don't have to be touched until users need to tweak the way whole cluster restore works for them. Am I on the right track? 

In that case could we consider flipping what Jake mentioned -- place a little blurb in a suitable place in the TS doc instead, and from there link out to this whole cluster restore doc?

JakeSCahill: If you're restoring from a failed cluster. I assume you can't tweak those values, so they'd need to be considered when setting up TS. @andrwng is that right?

andrwng: >In that case could we consider flipping what Jake mentioned -- place a little blurb in a suitable place in the TS doc instead, and from there link out to this whole cluster restore doc?

This sounds reasonable to me too, but I don't have a strong opinion either way for this, though I think I lean towards this because metadata recovery feels a bit different from what I think of when I think of tiered storage (it's data, not metadata, though not sure if that is how a customer would think about it).

>If you're restoring from a failed cluster. I assume you can't tweak those values, so they'd need to be considered when setting up TS. @andrwng is that right?

That's right. In general, a customer who wants to use cluster restore wouldn't touch the configs. The configs are prerequisites to running through a cluster restore.

JakeSCahill:
@mattschumpert should the disaster recovery section have an index page with a discussion of the options, then include recovery mode as an option as well as whole cluster restore?

Disaster Recovery
|
|---Whole Cluster Restore
|
|---Recovery Mode

Recovery Mode is currently under Manage > Recovery Mode

Edit: We can handle this in a separate issue: https://github.com/redpanda-data/documentation-private/issues/2112

JakeSCahill:
@mattschumpert is `rpk cluster storage recovery` being renamed to `rpk cluster storage restore`? We don't have any beta docs for that command yet. cc @Deflaimun

mattschumpert:
It has been renamed already and the docs must be updated as well

mattschumpert:
@JakeSCahill recovery mode isn't so much for a disaster as misconfiguration so I don't think it belongs in a dr section

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,183 @@
...
+======
+
+- `storage.tiered`: Make sure to configure the target cluster with the same Tiered Storage settings as the failed source cluster.
+- `config.cluster.cloud_storage_attempt_cluster_recovery_on_bootstrap`: Automate cluster restore in Kubernetes. Setting to `true` is recommended when using an automated method for deployment. When bootstrapping a cluster with a given bucket, make sure that any previous cluster using the bucket is fully destroyed; otherwise Tiered Storage subsystems may interfere with each other.
+endif::[]
+
+ifndef::env-kubernetes[]
+For self-hosted, follow the steps to xref:deploy:deployment-option/self-hosted/manual/index.adoc[deploy a new cluster].
+
+For Redpanda Cloud, follow the steps to xref:deploy:deployment-option/cloud/index.adoc[provision a new cloud cluster].
```
mattschumpert: @kbatuigas  I'm not sure we should be saying anything about cloud as cloud is seeking to automate this on behalf of customers, and for Dedicated im not sure customers even have direct access to the credentials required to restore.  @bpraseed  shouldn't we exclude cloud from Whole Cluster Restore docs for now?

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,183 @@
...
+You can restore data from a failed cluster to a target cluster using the xref:reference:rpk/rpk-cluster/rpk-cluster-storage-recovery.adoc[`rpk cluster storage restore`] command. The wait flag `-w` tells the command to poll the status of the restore process and then exit when completed.
+
+[,bash]
+----
+rpk cluster storage restore start -w
+----
+
+=== Perform a rolling restart
+
+To ensure that the cluster reflects newly-restored configuration properties, perform a xref:manage:cluster-maintenance/rolling-restart.adoc[rolling restart] after the restore process.
```
mattschumpert: @andrwng a rolling restart shouldn't be required here, right?.

cc @kbatuigas

mattschumpert (changes requested):

JakeSCahill:
> It has been renamed already and the docs must be updated as well

Looks like we have `rpk` updates ready here: https://github.com/redpanda-data/docs/pull/192

Will update the commands in this doc.

modules/manage/partials/whole-cluster-restore.adoc
andrwng: It is required if any of the properties are restored that require a restart. If anything, we should update this to make it more obvious that this isn't always necessary

mattschumpert: Right, @kbatuigas  we should update this to say, depending on the cluster configuration included, a rolling restart may be required and to check rpk cluster config status to determine if a restart is required.

modules/manage/partials/whole-cluster-restore.adoc
JakeSCahill: Praseed confirmed on Slack that we will control this.

JakeSCahill:
```suggestion
```

mattschumpert (approved):

modules/manage/partials/whole-cluster-restore.adoc
andrwng: Also, I think I saw it before but I don't anymore: we should also mention that cloud storage should be configured on the cluster _before_ running through the restore command.

modules/manage/partials/whole-cluster-restore.adoc
JakeSCahill: Done ✅

andrwng (approved):
LGTM!

modules/manage/pages/kubernetes/k-whole-cluster-restore.adoc
```diff
@@ -0,0 +1,6 @@
+= Whole Cluster Restore for Disaster Recovery in Kubernetes
```
Feediver1, line 1: Curious to know why docs is using "Whole" here? Isn't "Cluster Restore for Disaster Recovery..." equivalent? Do we have partial cluster restore? I could not find that...Seems like "Whole" here is superfluous. Is this simply a carryover from internal eng speak?

Feediver1: Why not just call this Disaster Recovery?

modules/reference/pages/cluster-properties.adoc
```diff
@@ -240,6 +264,18 @@ Enable automatic partition rebalancing when new nodes are added.
 
 ---
 
+=== enable_cluster_metadata_upload_loop
+
+Enables cluster metadata uploads. Required for xref:manage:tiered-storage.adoc#whole_cluster_restore_for_disaster_recovery[whole cluster restore].
```
Feediver1: This link takes me to the Tiered Storage topic, not Whole Cluster Restore for Disaster Recovery.

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,210 @@
...
+** `cloud_storage_azure_adls_endpoint`
+** `cloud_storage_azure_adls_port`
+
+== Manage source metadata uploads
+
+By default, Redpanda uploads cluster metadata to object storage periodically. You can manage metadata uploads for your source cluster, or disable them entirely, with the following cluster configuration properties:
+
+* xref:reference:cluster-properties.adoc#enable_cluster_metadata_upload_loop[`enable_cluster_metadata_upload_loop`]: Enable metadata uploads. This property is enabled by default and is required for whole cluster restore.
+* xref:reference:cluster-properties.adoc#cloud_storage_cluster_metadata_upload_interval_ms[`cloud_storage_cluster_metadata_upload_interval_ms`]: Set the time interval to wait between metadata uploads.
+* xref:reference:cluster-properties.adoc#controller_snapshot_max_age_sec[`controller_snapshot_max_age_sec`]: Maximum amount of time that can pass before Redpanda attempts to take a controller snapshot after a new controller command appears. This property affects how fresh the uploaded metadata can be.
```
Feediver1:
```suggestion
* xref:reference:cluster-properties.adoc#controller_snapshot_max_age_sec[`controller_snapshot_max_age_sec`]: Maximum amount of time that can pass before Redpanda attempts to take a controller snapshot after a new controller command appears. This property affects how current the uploaded metadata can be.
```

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,210 @@
...
++
+Example output when a restart is required:
++
+[,bash]
+----
+NODE  CONFIG-VERSION  NEEDS-RESTART  INVALID  UNKNOWN
+1     4               true           []       []
+----
+
+. If a restart is required, perform a xref:manage:cluster-maintenance/rolling-restart.adoc[rolling restart].
```
Feediver1, line 184: Link takes you back to top of page/beginning of topic, not rolling restart.

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,210 @@
...
++
+Example output when a restart is required:
++
+[,bash]
+----
+NODE  CONFIG-VERSION  NEEDS-RESTART  INVALID  UNKNOWN
+1     4               true           []       []
+----
+
+. If a restart is required, perform a xref:manage:kubernetes/k-rolling-restart.adoc[rolling restart].
```
Feediver1, line 213: Link took me to top of page, not rolling restart topic.

Feediver1 (approved):
Please check to be sure the links are working as intended.

modules/reference/pages/cluster-properties.adoc
JakeSCahill: Fixed ✅ thanks

modules/manage/partials/whole-cluster-restore.adoc
JakeSCahill: We need a rebase off `main`. That doc was merged into `main`.

modules/manage/partials/whole-cluster-restore.adoc
JakeSCahill: We need a rebase off `main`. That doc was merged into `main`.

modules/manage/pages/kubernetes/k-whole-cluster-restore.adoc
JakeSCahill: Kat has plans for a DR guide: https://github.com/redpanda-data/documentation-private/issues/2112

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,210 @@
...
+You can restore data from a source cluster to a target cluster using the xref:reference:rpk/rpk-cluster/rpk-cluster-storage-restore.adoc[`rpk cluster storage restore`] command.
+
+. Restore data from the source cluster:
++
+[,bash]
+----
+rpk cluster storage restore start -w
+----
++
+The wait flag ()`-w`) tells the command to poll the status of the restore process and then exit when completed.
```
JakeSCahill:
```suggestion
The wait flag (`-w`) tells the command to poll the status of the restore process and then exit when completed.
```

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -0,0 +1,217 @@
...
+:link-recovery-mode: manage:kubernetes/k-recovery-mode.adoc
+endif::[]
+ifndef::env-kubernetes[]
+:link-tiered-storage: manage:tiered-storage.adoc
+:link-recovery-mode: manage:recovery-mode.adoc
+endif::[]
+
+With xref:{link-tiered-storage}[Tiered Storage] enabled, you can use whole cluster restore to restore data from a failed cluster (source cluster), including its metadata, onto a new cluster (target cluster). This is a simpler and cheaper alternative to active-active replication, for example with xref:upgrade:migrate/data-migration.adoc[MirrorMaker 2]. Use this recovery method to restore your application to the latest functional state as quickly as possible.
+
+NOTE: You cannot use whole cluster restore if the source cluster is in xref:{link-cluster-restore}[recovery mode].
```
JakeSCahill:
```suggestion
NOTE: You cannot use whole cluster restore if the target cluster is in xref:{link-cluster-restore}[recovery mode].
```

PR #127, opened 2023-11-23
Update to remove the image tag  version for operator helm command

bhataprameya:
- Update to remove the image tag  version for operator helm command

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -7,6 +7,7 @@ nav:
 asciidoc:
   attributes:
     full-version: 23.2.3
+    operator-version: 2.1.1
     latest-release-commit: '9eefb907c'
     supported-kubernetes-version: 1.21
     supported-helm-version: 3.6.0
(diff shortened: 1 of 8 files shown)


JakeSCahill:
Thanks for your contribution @bhataprameya. You're right, the current tag should be updated. However, I believe the advice is to always start with the latest version of the operator, so I think it may be best to just remove the line where we set the image tag of the operator chart.

@alejandroEsc or @RafalKorepta WDYT?

RafalKorepta:
> However, I believe the advice is to always start with the latest version of the operator, so I think it may be best to just remove the line where we set the image tag of the operator chart.

I agree. It's easier when documentation doesn't have constantly moving target.

bhataprameya:
Updated The PR to remove the image tag

JakeSCahill (approved):
Thank you @bhataprameya 🎉

PR #138, opened 2023-11-29
Add OIDC docs for K8s and restructure authentication topic

JakeSCahill:
Review deadline: 4 December

@BenPope please review the structure and detail in 'Authentication for the HTTP APIs'.
@alejandroEsc please review the K8s content: https://deploy-preview-138--redpanda-docs-preview.netlify.app/23.3/manage/kubernetes/security/authentication/sasl-kubernetes/

We are single-sourcing, so content that only applies to Linux is nested within `ifndef::env-kubernetes[]endif::[]` and content that only applies to K8s is nested within `ifdef::env-kubernetes[]endif::[]`.

Fixes redpanda-data/documentation-private#1992 fixes redpanda-data/documentation-private#2122

- Added OIDC docs for Kubernetes.
- Restructured the authentication docs according to redpanda-data/documentation-private#2122.
- Single-sourced content between Linux and Kubernetes authentication topics to remove duplication.

Preview: https://deploy-preview-138--redpanda-docs-preview.netlify.app/23.3/manage/security/authentication/

Related issues:

- [x] The Kubernetes docs will not work correctly until this issue is fixed: https://github.com/redpanda-data/team-kubernetes-internal/issues/14
- [ ] Basic auth for the Admin API isn’t working correctly in the Helm chart. See https://github.com/redpanda-data/helm-charts/issues/848.


modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1583 @@
...
+    - address: 0.0.0.0
+      port: 9092
+      name: sasl_listener
+      authentication_method: sasl
+----
+endif::[]
+
+==== Enable SASL with TLS encryption
+
+SASL provides authentication, but not encryption. To provide encryption, you can enable TLS in addition to SASL. See xref:{tls-doc}[]
```
Feediver1:
```suggestion
SASL provides authentication, but not encryption. To provide encryption, you can enable TLS in addition to SASL. See xref:{tls-doc}[].
```

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1583 @@
...
+
+Redpanda's implementation of OIDC adheres to the Client Credentials Flow defined in
+https://datatracker.ietf.org/doc/html/rfc6749#section-4.4[OAuth 2.0 RFC 6749, section 4.4^]
+in which a client obtains an access token from the authorization server, and provides this
+access token to Redpanda, either using SASL/OAUTHBEARER for the Kafka API, or an HTTP
+Authorization (Bearer) header.
+
+The access token is a _bearer token_. A bearer token is used for authentication and
+authorization in web applications and APIs, and holds user credentials, usually in the form
+of random strings of characters. bearer tokens are generated based on protocols and specifications
```
Feediver1:
```suggestion
of random strings of characters. Bearer tokens are generated based on protocols and specifications
```

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1583 @@
...
+
+. Specify the amount of time keys from the `jwks_uri` are cached:
++
+[,bash]
+----
+rpk cluster config set oidc_keys_refresh_interval 3600
+----
+endif::[]
+
+If you're configuring authentication for the first time, you may need to
```
Feediver1: Shouldn't this come very early on in the section/topic--before configuring authN?

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1583 @@
...
+** `c` is optional. It can be either `/L` or `/U` to make the match lowercase or uppercase.
+* `DEFAULT` The principal name is used as the local user name. If the principal has more than one component or is not in the default realm, then the conversion fails.
++
+Examples of a Kerberos UPN without a host (`jdoe@EXAMPLE.COM`) and with a host (`jdoe/host@EXAMPLE.COM`):
++
+|===
+| Translation | `jdoe@EXAMPLE.COM` | `jdoe/host@EXAMPLE.COM`
+| `[1:$1@$0]`
+| `jdoe@EXAMPLE.COM`
+| Rule does not matchfootnote:two-components[Rule does not match because there are two components in principal name `jdoe/host@EXAMPLE.COM`.]
```
Feediver1: So the actual notes for 1 and 2 here appear at the very end of the topic. If you encounter them there, it is a bit puzzling for sure, and just not a good user experience with this content. Is there any way you can position right underneath this table?

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1583 @@
...
+In this case, Redpanda supports only SCRAM.
++
+To add support for Kerberos, append the `sasl_mechanisms` property with the value `GSSAPI`:
++
+[,bash]
+----
+rpk cluster config set sasl_mechanisms '["SCRAM","GSSAPI"]'
+----
+
+If you're configuring authentication for the first time, you may need to
```
Feediver1: Place earlier? Commented previously on this.

Feediver1 (approved):
Marked a couple typos, and requested that the footnotes appear under the table rather than at the very end of the AuthN topic (out of context). Othwerwise, looks fantastic.

modules/manage/partials/authentication.adoc
JakeSCahill: We could do something custom. But in this case, I think it's simpler to not have footnotes and just include the reasons in the table cells.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+| HTTP Proxy (PandaProxy)
+|
+- Basic authentication
+- OIDC
+
+| Schema Registry
+|
+- Basic authentication
+- OIDC
+|===
```
BenPope, line 34: I'm not a fan of the scrollbar on this table.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+- Ensure that host names are fully qualified domain names (FQDN).
+- Ensure that each broker has a http://web.mit.edu/Kerberos/krb5-latest/doc/admin/conf_files/krb5_conf.html[Kerberos configuration file^] set to use Active Directory or another corporate key distribution center (KDC). Default is at `/etc/krb5.conf`.
+- Ensure that the KDC has a valid Kerberos service principal name (SPN) for each broker in the form `primary/<FQDN>@<REALM>`.
+- Ensure that each broker has a keytab containing the SPN for that broker. This must be located at an identical file path on each Redpanda broker. Default is `/var/lib/redpanda/redpanda.keytab`.
+endif::[]
+
+If you want to configure OIDC for any of the APIs:
+
+* In an enterprise setting, partner with your security organization to register the client ID and client secret with for your IdP.
+* To use OIDC with clients, you must register the client credentials.
```
BenPope: I think this deserves more explanation. I think this is suggesting that the client application requires that there are application credentials registered with the IdP.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+ifdef::env-kubernetes[]
+You must have the following:
+
+* Kubernetes cluster. Ensure you have a running Kubernetes cluster, either locally, such as with minikube or kind, or remotely.
+
+* https://kubernetes.io/docs/tasks/tools/#kubectl[Kubectl^]. Ensure you have the `kubectl` command-line tool installed and configured to communicate with your cluster.
+endif::[]
+ifndef::env-kubernetes[]
+If you want to configure Kerberos for the Kafka API:
```
BenPope: There's some tension between having prerequisites upfront, and deferring specifics that won't matter to most users. I'd tend towards pushing this to SASL/GSSAPI section.

It's more awkward with OIDC though, since it appears in both SASL and HTTP.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+To enable authentication, choose from the following options:
+
+* Enable SASL authentication for all Kafka listeners and basic authentication for the HTTP Proxy an Schema Registry APIs. Use this method if you haven't already enabled authentication. This method does not require you to restart the cluster.
++
+[,bash]
+----
+rpk cluster config set enable_sasl true
+----
+
+* Enable SASL authentication but don't set an authentication method for any of the listeners. This method requires you to restart the cluster.
```
BenPope: This is the option that DOES require an `authentication_method` node configuration.

This option allows configuring non-sasl AuthN on one or more listeners:
```yaml
    - address: 127.0.0.2
      port: 19092
      name: mtls_identity
      authentication_method: mtls_identity
```

[This page](https://deploy-preview-138--redpanda-docs-preview.netlify.app/23.3/manage/security/listener-configuration/#mixed-mode-authentication-with-multiple-listeners) describes `authentication_method`.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+ifdef::env-kubernetes[]
+When authentication is enabled in the Redpanda Helm chart, SASL authentication is enabled for the Kafka API listeners and basic authentication is enabled for the HTTP API listeners. Enabling authentication for listeners ensures that only authorized users can interact with your Redpanda cluster.
+
+Both superusers and SCRAM users can authenticate to these APIs using their respective credentials.
+endif::[]
+ifndef::env-kubernetes[]
+To enable authentication, choose from the following options:
+
+* Enable SASL authentication for all Kafka listeners and basic authentication for the HTTP Proxy an Schema Registry APIs. Use this method if you haven't already enabled authentication. This method does not require you to restart the cluster.
```
BenPope: Use this option if you only require SASL. Do not set an `authentication_method` on the listener.

This option also enables AuthZ, so you'll want a superuser before doing this, or you'll lock yourself out of the cluster (I don't know if there is any protection against this).

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
++
+[,bash]
+----
+rpk cluster config set kafka_enable_authorization true
+----
+
+For details about these cluster configurations, see xref:manage:cluster-maintenance/cluster-property-configuration.adoc[Cluster configuration properties].
+endif::[]
+
+=== Create superusers
```
BenPope: This should come before enabling AuthZ (which can happen via `enable_sasl`).

We should probably have a mechanism to test the superuser before enabling AuthZ.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+rpk cluster config set superusers ['<superuser-username>']
+----
++
+This can be a new user or an existing user. For example, if you use the superuser named `admin`, then Redpanda allows the `admin` user to do anything, but Redpanda does not create the `admin` user.
+
+. Create the superuser:
++
+[,bash]
+----
+rpk acl user create <superuser-username> \
```
BenPope: It's probably worth mentioning the `--mechanism` flag here, too.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+. Specify the username of a superuser.
++
+[,bash]
+----
+rpk cluster config set superusers ['<superuser-username>']
+----
++
+This can be a new user or an existing user. For example, if you use the superuser named `admin`, then Redpanda allows the `admin` user to do anything, but Redpanda does not create the `admin` user.
+
+. Create the superuser:
```
BenPope: For posterity: It's possible to have a superuser that's not a SCRAM user., i.e., it can be provided by an external mechanism such as OIDC. This is because  `superusers` is actually checking principals, not SCRAM users.

However, there's a current limitation in that rpk can only speak HTTP Basic AuthN to the admin API, which actually requires a SCRAM user.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+- <<scram,SASL/SCRAM>>
+
+- <<oidc,SASL/OAUTHBEARER>> (OpenID Connect, also known as OIDC)
+
+ifndef::env-kubernetes[- <<kerberos, SASL/GSSAPI>> (Kerberos)]
+
+==== Enable SASL
+
+To enable SASL authentication for the Kafka API, set the `authentication_method` property of the Kafka listeners to `sasl`.
```
BenPope, line 478: This is _required_ if `kafka_enable_authorization` was used.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+ifdef::env-kubernetes[The Redpanda Helm chart sets the `authentication_method` property to `sasl` for all Kafka listeners by default when you <<enable-authentication, enable authentication>>.]
+
+ifndef::env-kubernetes[If you <<enable-authentication, enabled authentication with `enable_sasl=true`>>, Redpanda sets `authentication_method` to `sasl` for the Kafka listeners by default.]
```
BenPope: It's not really "by default", it's "implicitly", so setting `authentication_method` is not required.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+- `SCRAM-SHA-256`
+- `SCRAM-SHA-512`
+
+By default, SCRAM users don't have any permissions in the cluster. Only superusers can grant permissions to new users through ACLs.
+
+. To create the user `myuser` with a password `changethispassword`, run xref:reference:rpk/rpk-acl/rpk-acl-user-create.adoc[`rpk acl user create`]:
++
+```bash
+rpk acl user create myuser -p 'changethispassword'
```
BenPope: `--mechanism SCRAM-SHA-256` (256 is default, but setting it to 512 here would mean updating the working example below).

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+[[schema-and-http-to-redpanda]]
+===== Configure Schema Registry and HTTP Proxy to connect to Redpanda with SASL
+
+Schema Registry and HTTP Proxy connect to Redpanda over the Kafka API. For the Kafka username and password, Redpanda uses ephemeral credentials internal to the cluster. Ephemeral credentials are regular SCRAM credentials, but they're only stored in memory and are lost when a broker restarts. When the Schema Registry or HTTP Proxy start up, they broadcast an ephemeral credential to other brokers over the internal RPC. If authentication fails to a particular broker, new ephemeral credentials are sent to that broker, and the service reconnects.
+
+Schema Registry and HTTP Proxy support only the SASL/SCRAM mechanism.
+
+ifndef::env-kubernetes[]
+You can override the ephemeral credentials and manually configure Schema Registry and HTTP Proxy to connect to Redpanda with SASL.
```
BenPope: Doing this will also require setting up suitable ACLs for the PP/SR user.

I think it should be clearer that setting it manually is only recommended when an administrator wants to lock them down with specific ACLs, rather than using Basic Auth(or OIDC) Authentication on the HTTP APIs.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+      key_file: broker.key
+      cert_file: broker.crt
+      truststore_file: ca.crt
+      enabled: true
+      require_client_auth: false
+----
+endif::[]
+
+[[scram]]
+==== SASL/SCRAM
```
BenPope, line 528: This requires `SCRAM` to exist in the `sasl_mechanisms` cluster configuration
```bash
rpk cluster config set sasl_mechanisms '["SCRAM"]' -X admin.hosts=localhost:9644
```
That is the default.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+NOTE: OpenID Connect (OIDC) authentication requires an xref:get-started:licenses.adoc[Enterprise license].
+To upgrade, contact https://redpanda.com/try-redpanda?section=enterprise-trial[Redpanda sales^].
+
+When you enable https://openid.net/developers/how-connect-works/[OIDC^], Redpanda and
+Redpanda Console can delegate the authentication process to an external identity provider
+(IdP) such as Okta, Azure AD, or on-premise Active Directory Federation Service (AD FS).
+With OIDC enabled, Redpanda does not need to manage user credentials directly,
+but can instead rely on the trusted authentication capabilities of established IdPs.
+
+Redpanda's implementation of OIDC provides SASL/OAUTHBEARER support for the Kafka API, and enables
```
BenPope: `s/enables/supports/`

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+  "name": "Jane Doe",
+  "locale": "en_US",
+  "email": "jdoe@redpanda.com",
+  "preferred_username": "jdoe@redpanda.com",
+  "given_name": "Jane",
+  "family_name": "Doe",
+  "zoneinfo": "America/Los_Angeles",
+  "email_verified": true
+}
+----
```
BenPope: We don't access `/userinfo`, the principal is extracted from the claims in the JWT described above.
```suggestion
Applications that connect to Redpanda exchange their client-id and secret for an access token and forward it to Redpanda via `SASL/OAUTHBEARER` or HTTP `Authorization: Bearer` header.
```

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+URL uses the default value:
++
+[,bash]
+----
+rpk cluster config set oidc_discovery_url 'https://auth.prd.cloud.redpanda.com/.well-known/openid-configuration'
+----
++
+This sets the OIDC-specific configuration
+for all clusters in the `.bootstrap.yaml` (located in the same directory as the `redpanda.yaml`). The OIDC
+discovery URL is read upon the first startup after configuration, and ignored on all subsequent startups.
```
BenPope: In order to set cluster config, the cluster must be up, so even if it did set the value in `.bootstrap.yaml` (it doesn't), it wouldn't be read, as noted.

It's possible to set this in `.boostrap.yaml`, I guess, as is any cluster config.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+Use `rpk cluster config set` for any subsequent modifications of the discovery URL.
+
+. Specify the intended recipient of the token:
++
+[,bash]
+----
+rpk cluster config set oidc_token_audience 'redpanda'
+----
+
+. Specify the principal mapping, which is a JSON path that extracts a principal from any claim in the access token payload:
```
BenPope: It's a bit more comprehensive than that now: https://github.com/redpanda-data/redpanda/pull/14772

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+[[kerberos]]
+==== GSSAPI (Kerberos)
+
+NOTE: Kerberos authentication requires an xref:get-started:licenses.adoc[Enterprise license]. To upgrade, contact https://redpanda.com/try-redpanda?section=enterprise-trial[Redpanda sales^].
+
+To configure Kerberos authentication, use a keytab, which contains credentials for the service.
+
+. <<enable-authentication, Enable SASL authentication>> if it's not already enabled.
+
+. Enable <<SASL/SCRAM>> for the Kafka API.
```
BenPope: It might be easier to combine this with step 8 (enabling GSSAPI)

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+| `+RULE:.*O=([^,]+).*/$1/L+`
+| `redpanda`
+
+| `+RULE:.*O=([^,]+),OU=([^,]+),.*,C=([^,]+)/$1-$2-$3/L+`
+| `redpanda-engineering-uk`
+
+| `DEFAULT`
+| `CN=www.redpanda.com,O=Redpanda,OU=Engineering,L=London,S=England,C=UK`
+|===
```
BenPope, line 1138: Not a fan of this scrollbar either, it only hides half of a row.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+  --set config.cluster.admin_api_require_auth=true
+----
+
+====
+--
+======
+endif::[]
+
+ifndef::env-kubernetes[]
+When you <<enable-authentication, enable authentication>> with `enable_sasl=true`, the Schema Registry API and the HTTP Proxy API are configured with basic authentication by default.
```
BenPope: This isn't true.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+To enable basic authentication for the Admin API:
+
+. Enable authentication for the Admin API:
```
BenPope, line 1278: I've been reading these docs for so long now, but I wonder if it's worth noting that the user probably wants to have enabled a superuser before doing this command.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1600 @@
...
+schema_registry:
+  schema_registry_api:
+    address: "localhost"
+    port: 8081
+    authentication_method: none
+----
+
+If authorization is enabled, connections to this listener use the anonymous user.
+
+To disable authentication on the Kafka API, disable `kafka_enable_authorization` and set `authentication_method` to `none` for all listeners.
```
BenPope: Or, depending on how you did it, `enable_sasl` = false

BenPope: It might be worth splitting this slightly:
`kafka_enable_authorization` is for AuthZ
`authentication_method` is for AuthN

modules/manage/partials/authentication.adoc
JakeSCahill: You still need to set `authentication_method`? I thought we say basic auth is enabled here: https://github.com/redpanda-data/docs/pull/138#discussion_r1412023588

BenPope: `enable_sasl` enables sasl AuthN and AuthZ for Kafka API. It does not affect the the HTTP listeners, which must be configured with an `authentication_method` in the node config (`redpanda.yaml`).

modules/manage/partials/authentication.adoc
BenPope: This setting does not enable basic auth for HTTP listeners.

modules/manage/partials/authentication.adoc
JakeSCahill: Will fix in a separate PR. This is defined in our UI, which is a separate project.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1661 @@
...
+- Enable SASL authentication for all Kafka listeners:
++
+Use this method if you haven't already enabled authentication and you want to apply SASL authentication to all Kafka listeners. This approach does not require you to restart the cluster.
++
+[source,bash]
+----
+rpk cluster config set enable_sasl true
+----
++
+NOTE: This command implicitly enables authorization. If you want to disable authorization, you can use the `kafka_enable_authorization` cluster configuration property.
```
BenPope: I think the note should go above the potentially dangerous command.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1661 @@
...
+
+==== Enable SASL
+
+To enable SASL authentication for the Kafka API, set the `authentication_method` property of the Kafka listeners to `sasl`.
+
+ifdef::env-kubernetes[The Redpanda Helm chart sets the `authentication_method` property to `sasl` for all Kafka listeners by default when you <<enable-authentication, enable authentication>>.]
+
+ifndef::env-kubernetes[If you <<enable-authentication, enabled authentication with `enable_sasl=true`>>, Redpanda implicitly sets `authentication_method` to `sasl` for the Kafka listeners.]
+
+ifndef::env-kubernetes[If you <<enable-authentication, enabled authentication with `kafka_enable_authorization=true`>>, you must enable SASL for the Kafka API.]
```
BenPope: This jump point could be more specific. #enable-sasl-authentication

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1661 @@
...
+
+ifndef::env-kubernetes[- <<kerberos, SASL/GSSAPI>> (Kerberos)]
+
+==== Enable SASL
+
+To enable SASL authentication for the Kafka API, set the `authentication_method` property of the Kafka listeners to `sasl`.
+
+ifdef::env-kubernetes[The Redpanda Helm chart sets the `authentication_method` property to `sasl` for all Kafka listeners by default when you <<enable-authentication, enable authentication>>.]
+
+ifndef::env-kubernetes[If you <<enable-authentication, enabled authentication with `enable_sasl=true`>>, Redpanda implicitly sets `authentication_method` to `sasl` for the Kafka listeners.]
```
BenPope: This jump point could be more specific. #enable-sasl-authentication

It's only implicitly set if it's not explicitly set.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1661 @@
...
+To check if SASL/SCRAM is enabled:
+
+[,bash]
+----
+rpk cluster config get sasl_mechanisms
+----
+
+You should see `SCRAM` in the output.
+
+If SASL/SCRAM is not enabled, enable it:
```
BenPope:
```suggestion
If SASL/SCRAM is not enabled, enable it by setting or appending to the list:

File: modules/manage/partials/authentication.adoc [hunk 32]
```diff
@@ -0,0 +1,1661 @@
...
++
+```bash
+rpk acl user create myuser \
+  -p 'changethispassword' \
+  --mechanism SCRAM-SHA-256
+```
++
+TIP: Enclose passwords in single quotes to avoid conflicts with special characters. Enclosing characters in single quotes preserves the literal value of each character.
+
+. Use the xref:reference:rpk/rpk-acl/rpk-acl-create.adoc[`rpk acl create`] command to grant `create` and `describe` permissions to `myuser` in the cluster:
```
BenPope: Do we have a page we can link the describes ACLs? E.g., resources (e.g., cluster, topic, group) and permissions (e.g., create, describe).

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1661 @@
...
+```bash
+rpk topic create myfirsttopic \
+  -X user=myuser \
+  -X pass='changethispassword' \
+  -X sasl.mechanism=SCRAM-SHA-256
+```
+
+[TIP]
+====
+You can also use environment variables to set the user credentials instead of flags:
```
BenPope: I suspect that profiles are recommended now, although ENV is also useful. Not sure these alternative configuration mechanisms aren't more noise than signal in this part of the doc.

modules/manage/partials/authentication.adoc
BenPope: I think documenting the mapping is worthwhile.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1661 @@
...
+This registration process involves creating a new application entry in the IdP's management console.
+During this process, you'll specify details about your application,
+such as the type of application, the callback URLs, and any other required information as per your IdP's requirements.
+In an enterprise environment, OIDC integration typically requires coordination with your organization's security team.
+
+. <<enable-authentication, Enable SASL authentication>> if it's not already enabled.
+
+. Configure xref:manage:security/authorization.adoc#acls[ACLs] for your users so they can access Redpanda resources.
++
+NOTE: If you enabled OIDC authentication for the Admin API, you must also <<create-superusers,create a SCRAM superuser>> so that you can use `rpk` to create ACLs. `rpk` supports only basic authentication for the Admin API. See <<Authentication for the HTTP APIs>>.
```
BenPope: This is nested under the Kafka API section, it's probably more relevant down in #oidc-http.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1661 @@
...
+
+. Specify the discovery URL of your identity provider (IdP). The following IdP
+URL uses the default value:
++
+[,bash]
+----
+rpk cluster config set oidc_discovery_url 'https://auth.prd.cloud.redpanda.com/.well-known/openid-configuration'
+----
+
+. Specify the intended recipient of the token:
```
BenPope: recipient isn't described elsewhere:
```suggestion
. Specify the intended audience of the token:

File: modules/manage/partials/authentication.adoc [hunk 36]
```diff
@@ -0,0 +1,1661 @@
...
+
+NOTE: If the broker's certificate is signed by a well-known, trusted CA, and you're confident about the integrity of your system's CA trust store, you don't need the `--cacert` flag.
+endif::[]
+
+For all available endpoints, see xref:api:ROOT:pandaproxy-schema-registry.adoc[].
+
+[[oidc-http]]
+=== OIDC
+
+If you've configured the <<oidc, SASL/OAUTHBEARER>> mechanism for the Kafka API, you can configure the HTTP APIs to authenticate users with the OIDC bearer token. By using OIDC for all HTTP APIs, you centralize credentials and provide a password-free,
```
BenPope: You can configure OIDC for the HTTP APIs without having configured it for Kafka API. (the `oidc_*` cluster configs apply here, too.)

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1661 @@
...
+pandaproxy:
+  pandaproxy_api:
+  - address: "localhost"
+    port: 8082
+    authentication_method: http_basic
+schema_registry:
+  schema_registry_api:
+    address: "localhost"
+    port: 8081
+    authentication_method: http_basic
```
BenPope, line 1305: The nesting here doesn't make it clear that `authentication_method` is also required for OIDC.

Also, can we have the pretty highlight one the `authentication_method` lines?

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1661 @@
...
+
+Following are additional validation claims (JWT properties) that are included in the access token:
+
+* `alg`: The signature algorithm. The extension point in the JWT header is the signature algorithm used to sign the token, and cannot contain the value `none`.
+* `aud`:  Audience. Must match the configuration specified in `oidc_token_audience`. Cannot contain the value `none`.
+* `kid`: Key identifier. Must match _any_ of the public JWK listed in the `jwks_uri` endpoint.
+* `exp`: Expiration. The timestamp listed is greater than current time. Must validate within acceptable bounds of the value specified in `oidc_clock_skew_tolerance`. A clock skew tolerance period may be configured by an Admin to account for clock drift between Redpanda and the OIDC Identity Provider (IdP).
+* `iss`: Issuer. Must exactly match the `issuer` property of the JSON returned from the URL specified in `oidc_discovery_url`.
+* `scope`: Scope. Must include the value `openid`.
+* `sub`: Subject. Must match exactly the Principal ID found through the principal mapping.
```
BenPope: `sub` is a default provided claim, the RFC states: 
> The "sub" (subject) claim identifies the principal that is the subject of the JWT.

This is one claim that can be mapped to a Redpanda principal.

modules/manage/partials/authentication.adoc
JakeSCahill: Yes, added.

modules/manage/partials/authentication.adoc
```diff
@@ -770,7 +758,7 @@ Following are additional validation claims (JWT properties) that are included in
 * `exp`: Expiration. The timestamp listed is greater than current time. Must validate within acceptable bounds of the value specified in `oidc_clock_skew_tolerance`. A clock skew tolerance period may be configured by an Admin to account for clock drift between Redpanda and the OIDC Identity Provider (IdP).
 * `iss`: Issuer. Must exactly match the `issuer` property of the JSON returned from the URL specified in `oidc_discovery_url`.
 * `scope`: Scope. Must include the value `openid`.
-* `sub`: Subject. Must match exactly the Principal ID found through the principal mapping.
+* `sub`: Subject. This default claim identifies the principal subject. In Redpanda, it can be mapped to a recognized Redpanda principal for secure authentication and authorization.
```
BenPope: Any claim can be mapped to a redpanda principal, this is just the default, "$.sub".

modules/manage/partials/authentication.adoc
```diff
@@ -912,19 +898,48 @@ URL uses the default value:
...
+* Extract principal from the email field (`$.user_info.email/([^@]+)@.*/$1/L`): This rule captures the username part of the email before the `@` symbol and converts it to lowercase. The resulting principal is `user`.
+* Extract principal with domain validation (`$.user_info.email/([^@]+)@example.com/$1/L`): This rule is similar to the previous one but only applies if the email domain matches `example.com`. The resulting principal is `user` if the domain matches, otherwise, the mapping fails.
++
+To apply a principal mapping rule in Redpanda, use the following command:
 +
 [,bash]
 ----
 rpk cluster config set oidc_principal_mapping '$.sub'
-rpk cluster config set oidc_principal_mapping '$.email'
+rpk cluster config set oidc_principal_mapping '$.user_info.email/([^@]+)@.*/$1/L'
```
BenPope: This seems a bit too nested.

BenPope (commented):
This page is SO much better now!

modules/manage/partials/authentication.adoc
JakeSCahill: Fixed

BenPope (approved):

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1695 @@
...
+
+helm_ref:auth.sasl[]
+
+==== Use a Secret resource
+
+To use a Secret resource to store superuser credentials:
+
+. Create a file in which to store the credentials.
++
+NOTE: Make sure to include an empty line at the end of the file.
```
alejandroEsc, line 89: no longer required, you can remove this note.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1695 @@
...
+- `<superuser-password>`: The superuser's password.
+- `<superuser-authentication-mechanism>`: The authentication mechanism. Valid values are `SCRAM-SHA-256` or `SCRAM-SHA-512`.
++
+Or, leave this placeholder empty to set it to the default authentication mechanism. The default is `SCRAM-SHA-512`. This default is applied to all superusers that don't include an explicit authentication mechanism.
+
+. Use the file to create a Secret resource in the same namespace as your Redpanda cluster.
++
+[,bash]
+----
+kubectl create secret generic redpanda-superusers --namespace <namespace> --from-file=superusers.txt
```
alejandroEsc:
```suggestion
kubectl --namespace <namespace> create secret generic redpanda-superusers --from-file=superusers.txt
```

I understand if there is a formal style for the <substitutions> but having the namespace after kubectl helps mentally understand the command.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1695 @@
...
+```bash
+helm upgrade --install redpanda redpanda/redpanda --namespace <namespace> --create-namespace \
+  --values sasl-enable.yaml --reuse-values
+```
+
+--set::
++
+[,bash,lines=5-12]
+----
+helm upgrade --install redpanda redpanda/redpanda --namespace <namespace> --create-namespace \
```
alejandroEsc, line 851: We really need to stop asking users to use `--set`. The user completely looses their history and so will have to add these again in the future or use the flag to use previous values, and that can cause difficulties upgrading.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1695 @@
...
+    cert_file: broker.crt
+    truststore_file: ca.crt
+    enabled: true
+  scram_username: <username>
+  scram_password: <password>
+  sasl_mechanism: SCRAM-SHA-256
+----
+endif::[]
+
+[[oidc]]
```
alejandroEsc, line 687: I think what is difficult about this document is how large it is. Can we determine a way to split off these concepts a bit. For example we can make a point about what is authorization (be given permissions to do things) and how to set this up vs what is authentication (being able to determine identity of agent trying to perform actions ) and how to set these up. I think logically dividing this document will lower the complexity for the users.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1695 @@
...
+* `iss`: Issuer. Must exactly match the `issuer` property of the JSON returned from the URL specified in `oidc_discovery_url`.
+* `scope`: Scope. Must include the value `openid`.
+* `sub`: Subject. This default claim identifies the principal subject. While `sub` is the default mapping (`$.sub`) in Redpanda, any claim within the JWT can be mapped to a Redpanda principal.
+
+===== Enable OIDC
+
+. Register a client application with your IdP.
++
+A client application, in this context, refers to any application or service that will authenticate against the Redpanda cluster using OIDC.
+This registration process involves creating a new application entry in the IdP's management console.
```
alejandroEsc:
```suggestion
This registration process involves creating a new entry in the IdP's management console for the application, sometimes called a client.
```

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1695 @@
...
+During this process, you'll specify details about your application,
+such as the type of application, the callback URLs, and any other required information as per your IdP's requirements.
+In an enterprise environment, OIDC integration typically requires coordination with your organization's security team.
```
alejandroEsc, line 769: We should mention and make it very clear that the IDP only interacts with the application by giving it access to endpoints (via discovery url) that it can use to verify tokens. The IDP can then create tokens that are authenticated against the application. It may seem obvious but the fact that the interaction with a token agains the application is the whole point is a bit lost with all the details here.

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1695 @@
...
+    enabled: true
+----
+
+endif::[]
+
+== Authentication for the HTTP APIs
+
+The following HTTP APIs support basic authentication and OIDC authentication:
+
+- Admin API
```
alejandroEsc, line 1183: not kafka right?

modules/manage/partials/authentication.adoc
```diff
@@ -0,0 +1,1695 @@
...
+====
+--values::
++
+.`enable-sasl.yaml`
+[,yaml,lines=7]
+----
+auth:
+  sasl:
+    enabled: true
+  secretRef: redpanda-superusers
```
alejandroEsc:
```suggestion
    secretRef: redpanda-superusers
```

alejandroEsc (approved):
a few comments, some for consideration but i will not block.

modules/manage/partials/authentication.adoc
JakeSCahill: We do recommend using the values file. We just provide `--set` for UX when you're just trying things out.

modules/manage/partials/authentication.adoc
JakeSCahill: Agreed. For this release, we're fixing the structure and adding OIDC. I've created a separate issue to handle the split: https://github.com/redpanda-data/documentation-private/issues/2149

PR #153, opened 2023-12-04
Admin API reference: partition recovery

kbatuigas:
Add new Admin API endpoint for feature referred to as partition recovery. This feature allows you to recover a "stuck" partition. Please refer to preview: https://deploy-preview-153--redpanda-docs-preview.netlify.app/api/admin-api.html#tag/Debugging/operation/force_update_partition_replicas

Partially resolves https://github.com/redpanda-data/documentation-private/issues/2105 (a separate PR to be opened to update content in `main`).

### DIFF
--- a/modules/ROOT/attachments/admin-api.yaml
+++ b/modules/ROOT/attachments/admin-api.yaml
@@ -3,7 +3,6 @@ info:
   title: Redpanda Admin API
   version: 1.0.0
   description: |
-  
     Use the Admin API to administer and manage Redpanda.
 
     ---
@@ -43,7 +42,7 @@ paths:
       responses:
         200:
           description: Brokers' configurations response
-          externalDocs: 
+          externalDocs:
             url: https://docs.redpanda.com/docs/reference/node-properties/
           content:
             application/json:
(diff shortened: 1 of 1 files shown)


modules/ROOT/attachments/admin-api.yaml
```diff
@@ -1049,7 +1049,35 @@ paths:
         required: false
         schema:
           type: integer
-          format: int64 
+          format: int64
+  /debug/partitions/{namespace}/{topic}/{partition}/force_replicas:
+    post: 
+      tags: 
+      -  Debugging
```
ztlpn: This is more related to "fixing" than "debugging" (i.e. understanding what went wrong). Therefore I think this should be a separate tag named something like "escape hatches" or "disaster recovery" if we have it.

modules/ROOT/attachments/admin-api.yaml
```diff
@@ -1049,7 +1049,35 @@ paths:
         required: false
         schema:
           type: integer
-          format: int64 
+          format: int64
+  /debug/partitions/{namespace}/{topic}/{partition}/force_replicas:
```
ztlpn, line 1051: Request body is required as well (array of replica assignments)

modules/ROOT/attachments/admin-api.yaml
JakeSCahill: I think this was tagged 'Debugging' because the endpoint is under the `/debug` path.

Are there any other endpoints that should be tagged for disaster recovery?

JakeSCahill: Maybe a better place for now is 'Partitions' since that tag has the description 'Manage partitions of a Redpanda cluster.'

ztlpn (approved):

Feediver1 (approved):
lgtm

PR #157, opened 2023-12-05
Include double dash syntax for `rpk cluster config set`

vuldin:
Setting a cluster parameter to `-1` doesn't work in most cases:

```
> rpk cluster config set delete_retention_ms "-1"
Error: unknown shorthand flag: '1' in -1
```

But a workaround is to use the double dash syntax, which this PR includes in our docs.

### DIFF
--- a/modules/reference/pages/rpk/rpk-cluster/rpk-cluster-config-set.adoc
+++ b/modules/reference/pages/rpk/rpk-cluster/rpk-cluster-config-set.adoc
@@ -39,13 +39,13 @@ help' for detail or '-X list' for terser detail.
 
 [NOTE]
 ====
-Setting parameters to non-number values (such as string values with a `-`) can be problematic for some terminals due to how POSIX flags are parsed from the command. For example, this command may not work from some terminals:
+Setting properties to non-number values (such as setting string values with `-`) can be problematic for some terminals due to how POSIX flags are parsed. For example, the following command may not work from some terminals:
 
 ```
 rpk cluster config set delete_retention_ms -1
 ```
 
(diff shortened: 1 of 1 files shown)


r-vasquez (approved):

modules/reference/pages/rpk/rpk-cluster/rpk-cluster-config-set.adoc
```diff
@@ -37,3 +37,17 @@ help' for detail or '-X list' for terser detail.
 |-v, --verbose |- |Enable verbose logging.
 |===
 
+[NOTE]
+====
+Setting parameters to non-number values (such as string values with a `-`) can be problematic for some terminals due to how POSIX flags are parsed from the command. For example, this command may not work from some terminals:
```
micheleRP:
```suggestion
Setting properties to non-number values (such as setting string values with `-`) can be problematic for some terminals due to how POSIX flags are parsed. For example, the following command may not work from some terminals:
```

modules/reference/pages/rpk/rpk-cluster/rpk-cluster-config-set.adoc
```diff
@@ -37,3 +37,17 @@ help' for detail or '-X list' for terser detail.
...
+
+```
+rpk cluster config set delete_retention_ms -1
+```
+
+This can be resolved by using an alternate syntax that includes a `--` to disable parsing for all subsequent characters:
```
micheleRP:
```suggestion
Workaround: Use `--` to disable parsing for all subsequent characters. For example:
```

micheleRP (approved):
Thanks @vuldin! Please see if you agree with my suggested edits before you merge!

PR #177, opened 2023-12-13
Remove DOMAIN environment variable

JakeSCahill:
- The Helm + Operator steps were missing the DOMAIN environment variable, resulting in incorrect domains being appended to the `/etc/hosts` file. To simplify the tutorial, this PR removes the environment variable in favor of hard-coding the domain.
- The Decommission controller does not currently work if you deploy the Helm chart with Argo CD: https://github.com/redpanda-data/helm-charts/issues/932
- The indentation of `statefulset.budget` was incorrect in the deployment doc.

### DIFF
--- a/modules/deploy/pages/deployment-option/self-hosted/kubernetes/kubernetes-deploy.adoc
+++ b/modules/deploy/pages/deployment-option/self-hosted/kubernetes/kubernetes-deploy.adoc
@@ -551,8 +551,8 @@ For smooth and uninterrupted updates, use the default `RollingUpdate` strategy. 
 statefulset:
   updateStrategy:
     type: "RollingUpdate"
-    budget:
-      maxUnavailable: 1
+  budget:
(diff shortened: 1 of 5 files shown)


Deflaimun:
What about using an editable variable?

voutilad:
An alternative idea... Maybe we keep `DOMAIN` so the docs are useful for folks looking to customize it, but the usage of it will default to the one we have in our examples?

For instance, we should be able to change usage of `${DOMAIN}` to `${DOMAIN:-customredpandadomain.local}`. Most shells will properly substitute `customredpandadomain.local` in this case if `DOMAIN` is not set in the environment. (We should check though that this will work with Helm or whatever.)

For folks just copy/pasting things, they should all "just work" and folks setting `DOMAIN` to something custom, the same snippets _should_ just work.

greg-szrama (approved):
Small change, looks good!

PR #179, opened 2023-12-13
Explain CPU utilization metrics 

kbatuigas:
### Preview
https://deploy-preview-179--redpanda-docs-preview.netlify.app/current/manage/monitoring/#cpu-usage

Adds more information to [CPU utilization](https://docs.redpanda.com/current/manage/monitoring/#cpu-usage) section regarding why cores may always look busy and which metrics should be monitored to determine actual Redpanda process utilization.

The section currently contains a NOTE admonition, but the community member who opened the [thread](https://redpandacommunity.slack.com/archives/C01AJDUT88N/p1699851597910539) in Slack may not have seen the admonition right away nor realized that it answers their question albeit partially. I've added the additional info to the same block but changed it to a TIP. With that said the block is now rather long so I am also open to going a different direction in terms of the format.

Closes https://github.com/redpanda-data/documentation-private/issues/2108

### DIFF
--- a/modules/manage/partials/monitor-redpanda.adoc
+++ b/modules/manage/partials/monitor-redpanda.adoc
@@ -287,9 +287,7 @@ While CPU utilization at the host-level might appear high (for example, 99-100% 
 
 This high host-level CPU utilization happens because Redpanda uses Seastar, which runs event loops on every core (also referred to as a _reactor_), constantly polling for the next task. This process never blocks and will increment clock ticks. It doesn't necessarily mean that Redpanda is busy.
 
-Use xref:reference:public-metrics-reference.adoc#redpanda_cpu_busy_seconds_total[`redpanda_cpu_busy_seconds_total`] (public metrics) or xref:reference:internal-metrics-reference.adoc#vectorized_reactor_utilization[`vectorized_reactor_utilization`] (internal metrics) to monitor the actual Redpanda CPU utilization. When these particular metrics indicate close to 100% utilization over a given period of time, you will then also start to see higher produce and consume <<latency,latency>>.
-
-If achieving very low latency consume is not a priority, consider tuning the xref:reference:tunable-properties.adoc#fetch_reads_debounce_timeout[`fetch_reads_debounce_timeout`] property. The default setting at 1ms means that Redpanda pings the partition once per millisecond of the given client poll cycle. Setting the duration at a higher value, for example up to 10ms, can help reduce CPU utilization.
+Use xref:reference:public-metrics-reference.adoc#redpanda_cpu_busy_seconds_total[`redpanda_cpu_busy_seconds_total`] (public metrics) or xref:reference:internal-metrics-reference.adoc#vectorized_reactor_utilization[`vectorized_reactor_utilization`] (internal metrics) to monitor the actual Redpanda CPU utilization. When these particular metrics indicate close to 100% utilization over a given period of time, make sure to also monitor produce and consume <<latency,latency>> as they may then start to increase as a result of resources becoming overburdened.
 ====
 
 ==== Memory allocated
--- a/modules/manage/partials/monitor-redpanda.adoc
+++ b/modules/manage/partials/monitor-redpanda.adoc
@@ -283,11 +283,11 @@ rate(redpanda_cpu_busy_seconds_total[5m])
 
 [TIP]
 ====
-While CPU utilization at the host-level can typically appear high (for example, 99-100% utilization) when I/O events like message arrival occur, the actual Redpanda process utilization is likely low, in contrast to what system-level metrics might indicate such as those provided by the `top` command. 
+While CPU utilization at the host-level might appear high (for example, 99-100% utilization) when I/O events like message arrival occur, the actual Redpanda process utilization is likely low. System-level metrics such as those provided by the `top` command can be misleading.  
 
 This high host-level CPU utilization happens because Redpanda uses Seastar, which runs event loops on every core (also referred to as a _reactor_), constantly polling for the next task. This process never blocks and will increment clock ticks. It doesn't necessarily mean that Redpanda is busy.
 
-Use `redpanda_cpu_busy_seconds_total` (public metrics) or xref:reference:internal-metrics-reference.adoc#vectorized_reactor_utilization[`vectorized_reactor_utilization`] (internal metrics) to monitor the actual Redpanda CPU utilization. When these particular metrics approach 100% utilization, you will then also start to see higher produce and consume <<latency,latency>>.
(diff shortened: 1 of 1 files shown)


modules/manage/partials/monitor-redpanda.adoc
```diff
@@ -281,11 +281,15 @@ To detect unexpected idling, you can query the rate of change as a percentage of
...
+
+Use `redpanda_cpu_busy_seconds_total` (public metrics) or xref:reference:internal-metrics-reference.adoc#vectorized_reactor_utilization[`vectorized_reactor_utilization`] (internal metrics) to monitor the actual Redpanda CPU utilization. When these particular metrics approach 100% utilization, you will then also start to see higher produce and consume <<latency,latency>>.
 
-This high host-level CPU utilization happens because Redpanda uses Seastar and constantly polls and increments clock ticks, and it doesn't necessarily mean that Redpanda is busy.
+If achieving very low latency consume is not a priority, consider tuning the xref:reference:tunable-properties.adoc#fetch_reads_debounce_timeout[`fetch_reads_debounce_timeout`] property. The default setting at 1ms means that Redpanda pings the partition once per millisecond of the given client poll cycle. Setting the duration at a higher value, for example up to 10ms, can help reduce CPU utilization.
```
hcoyote: Might also add something like:

Increasing above 10ms is not recommended as this will provide diminishing returns at the expense of further increasing average message consume latency.

hcoyote (changes requested):

modules/manage/partials/monitor-redpanda.adoc
hcoyote: btw, looks good otherwise.

modules/manage/partials/monitor-redpanda.adoc
```diff
@@ -281,11 +281,15 @@ To detect unexpected idling, you can query the rate of change as a percentage of
 rate(redpanda_cpu_busy_seconds_total[5m])
 ----
 
-[NOTE]
+[TIP]
 ====
-While CPU utilization at the host-level can typically appear high (for example, 99-100% utilization) when I/O events like message arrival occur, the actual Redpanda process utilization is likely low and can be monitored with `redpanda_cpu_busy_seconds_total`.
+While CPU utilization at the host-level can typically appear high (for example, 99-100% utilization) when I/O events like message arrival occur, the actual Redpanda process utilization is likely low, in contrast to what system-level metrics might indicate such as those provided by the `top` command.
```
greg-szrama:
```suggestion
While CPU utilization at the host-level might appear high (for example, 99-100% utilization) when I/O events like message arrival occur, the actual Redpanda process utilization is likely low. System-level metrics such as those provided by the `top` command can be very misleading. 
```

modules/manage/partials/monitor-redpanda.adoc
```diff
@@ -281,11 +281,15 @@ To detect unexpected idling, you can query the rate of change as a percentage of
...
+While CPU utilization at the host-level can typically appear high (for example, 99-100% utilization) when I/O events like message arrival occur, the actual Redpanda process utilization is likely low, in contrast to what system-level metrics might indicate such as those provided by the `top` command. 
+
+This high host-level CPU utilization happens because Redpanda uses Seastar, which runs event loops on every core (also referred to as a _reactor_), constantly polling for the next task. This process never blocks and will increment clock ticks. It doesn't necessarily mean that Redpanda is busy.
+
+Use `redpanda_cpu_busy_seconds_total` (public metrics) or xref:reference:internal-metrics-reference.adoc#vectorized_reactor_utilization[`vectorized_reactor_utilization`] (internal metrics) to monitor the actual Redpanda CPU utilization. When these particular metrics approach 100% utilization, you will then also start to see higher produce and consume <<latency,latency>>.
```
greg-szrama: Recommend adding a hyperlink to the `redpanda_cpu_busy_seconds_total` metric. Also, you might consider rephrasing this. The CPU Busy Seconds metric is an incrementing count of busy seconds, so it might be best monitored over a time slice to see heavy utilization, while the reactor utilization is more in line with showing a % utilizaion.

greg-szrama (approved):
A few points for consideration but otherwise looks good!

modules/manage/partials/monitor-redpanda.adoc
StephanDollberg: I don't think we should include this property at all. We don't want people to set properties which they don't really understand. This should be reserved for CS etc to apply it in a targeted fashion where needed. 

We do now finally have a PR up which should make this unneeded and it looks likely that we can also backport this fix.

One of our general principles is that Redpanda should not require any changing of the tunables and they are kind of there as an escape hatch if things go wrong.

kbatuigas: @hcoyote If it sounds good, I'll remove this paragraph. We could replace it with other info too, if there are any other actions we might recommend to users?

modules/manage/partials/monitor-redpanda.adoc
kbatuigas: @greg-szrama I left the link out since we have it earlier in the section but on second thought, no harm in adding it here again. Yes, it should specify the "calculation" for utilization and not the count itself... I'll find a way to rephrase. Thanks!

modules/manage/partials/monitor-redpanda.adoc
hcoyote: @StephanDollberg  this goes back to our discussion about what the appropriate value is for fetch_reads_debounce_timeout out of the box. With cloud deploys, it's 5ms. With self-managed it's 1. And we've run into several cases already where 1 is probably overly aggressive on self-hosted. 

If we do follow through with the plan to align the defaults between self-managed/cloud, then I'm ok removing this (or maybe moving this to a deeper/nuanced discussion in a KB). But I think it's important for users to understand the mechanism here, as we constantly run into the "why does redpanda take so much cpu" question.

StephanDollberg: > If we do follow through with the plan to align the defaults between self-managed/cloud

Like I said above, I am positive that hopefully soon this issue resolved. 

> as we constantly run into the "why does redpanda take so much cpu" question.

I understand but that issue is way larger than just fetch debounce. 

I'd be happy with a more KB/known-issue style article (that we can easily deprecate/remove later) about this than just a blank recommendation.

mattschumpert: If we're backporting the fix that makes tuning it totally unnecessary, I would err on the side of not documenting this (at least not in the vanilla 'monitoring redpanda' page).  Perhaps in the config reference for tunables (which targets more advanced users), but the monitoring page at least should be for things that are mainstream

hcoyote (approved):

StephanDollberg (approved):

PR #194, opened 2023-12-19
Admin API endpoints for nodewise recovery

kbatuigas:
[Preview - GET /partitions/majority_lost](https://deploy-preview-194--redpanda-docs-preview.netlify.app/api/admin-api/#get-/partitions/majority_lost)
[Preview - POST /partitions/force_recover_from_nodes](https://deploy-preview-194--redpanda-docs-preview.netlify.app/api/admin-api/#post-/partitions/force_recover_from_nodes)

This feature was released in 23.3 and the main (conceptual) doc was merged through this PR: https://github.com/redpanda-data/documentation-private/issues/1982.


modules/ROOT/attachments/admin-api.yaml
```diff
@@ -584,6 +584,33 @@ paths:
...
               schema:
                 $ref: '#/components/schemas/partitions_local_summary'
+  /partitions/majority_lost:
+    get: 
+      tags: 
+      - Partitions
+      summary: Get partitions with lost majority
+      description: List of partitions with lost majority given an input set of dead nodes.
+      operationId: majority_lost
+      # responses:
```
kbatuigas: Double check - [spec](https://github.com/redpanda-data/redpanda/blob/dev/src/v/redpanda/admin/api-doc/partition.json#L366) indicates "void" but this should return something like an array? Is [this](https://github.com/redpanda-data/redpanda/pull/13943/files#diff-ba91dcae6799a5074c74533fcdb852af069b7fe3763469b3b2463730a8d5184eR703) the response schema?

modules/ROOT/attachments/admin-api.yaml
```diff
@@ -584,6 +584,33 @@ paths:
...
+          in: query
+          required: true
+          schema:
+            type: string   
+  /partitions/force_recover_from_nodes:
+    post: 
+      tags: 
+      - Partitions
+      summary: Force recover partitions
+      description: Force recover partitions from input list of nodes.
```
kbatuigas: This endpoint requires an input list but no parameters defined in [spec](https://github.com/redpanda-data/redpanda/blob/dev/src/v/redpanda/admin/api-doc/partition.json#L393)?

modules/ROOT/attachments/admin-api.yaml
ztlpn: Correct, "void" is a mistake, this endpoint returns an array of `ntp_with_majority_loss`

modules/ROOT/attachments/admin-api.yaml
ztlpn: Here is [the request body schema](https://github.com/redpanda-data/redpanda/blob/e9b93b9cd397862145e33a7e193b118c85495c57/src/v/redpanda/admin/partition.cc#L1019), there are no query parameters

modules/ROOT/attachments/admin-api.yaml
mattschumpert: @ztlpn can we update the spec to be accurate please?  This will be important because we are trying to automate Admin API doc generation from the spec moving forward. Thanks

ztlpn: Paging @bharathv as the main author :)

Although I would say that problems with keeping schemas for request bodies in sync with the code are bound to appear because they these schemas are not used anywhere (we do separate validation in the admin server code).

modules/ROOT/attachments/admin-api.yaml
```diff
@@ -584,6 +584,45 @@ paths:
...
+          content: 
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ntp_with_majority_loss'
+      parameters:
+        - name: dead_nodes
+          in: query
+          required: true
+          schema:
+            type: string
```
kbatuigas: @bharathv this spec indicates query parameter for this API call. The [force recover operation](https://github.com/redpanda-data/docs/pull/194/files#diff-7b60b751908bd5ea8c33bdf12dea54174f7cdb13a1e47904d78346751db72cf3R862) requires an input array of nodes in the API request body. Just want to double check that both are correct. Does the request URL look something like `/partitions/majority_lost?dead_nodes=1,3,5` ?

Feediver1:
What's the latest status of this PR @kbatuigas ?

modules/ROOT/pages/admin/{post}-{/partitions/force_recover_from_nodes}.adoc
kbatuigas, line 1: @JakeSCahill should this work to add a "slot" for this particular method? I see a slot element show up in the HTML source but it doesn't actually render.

JakeSCahill: I added a fix here https://github.com/redpanda-data/docs/pull/194/commits/ea0d3eae68bbc7fd96028557f7fbb96b3b07757a

The syntax is [`{method}-{path}`](https://rapidocweb.com/api.html#slots) for the directory name (without the curly braces).

In order to support this, I needed to adjust our UI here: https://github.com/redpanda-data/docs-ui/pull/217

You won't see the slot until the UI change is merged.

JakeSCahill (approved):

modules/ROOT/attachments/admin-api.yaml
bharathv: > /partitions/majority_lost?dead_nodes=1,3,5

yes you are correct, thats how the URL  looks. A GET with a request_body is not considered a standard practice in the implementations, so I had to use a query parameter.
https://github.com/redpanda-data/redpanda/pull/13943#issuecomment-1814103858

bharathv (approved):

PR #214, opened 2023-12-27
Update number of cores

JakeSCahill:
A report from community Slack. The example output shows 8 cores but the text says 12 cores.


micheleRP (approved):
lgtm

PR #223, opened 2024-01-04
docs: Audit logging in Kubernetes

JakeSCahill:
Review deadline: January 10

Documents how to use audit logging with the Helm chart and the Redpanda resource (with and without the Topic resource).

Preview: https://deploy-preview-223--redpanda-docs-preview.netlify.app/current/manage/kubernetes/security/k-audit-logging/

Related Helm chart issues:

- https://github.com/redpanda-data/helm-charts/issues/953
- https://github.com/redpanda-data/helm-charts/issues/955


RafalKorepta (approved):

Feediver1:
Topic and Secret should not be in caps--or is this a k8s convention? Even so, introduces an inconsistency within our docs, because we don't use initcap with either.

JakeSCahill:
> Topic and Secret should not be in caps--or is this a k8s convention? Even so, introduces an inconsistency within our docs, because we don't use initcap with either.

These are conventions in Kubernetes. Topic and Secret are resources. In Kubernetes, they are capitalised to differentiate them from the generic terms.

Feediver1 (approved):
lgtm

PR #236, opened 2024-01-10
Single partition recovery

kbatuigas:
[Deploy preview](https://deploy-preview-236--redpanda-docs-preview.netlify.app/current/manage/cluster-maintenance/partition-recovery/)

- Explain when and how to carry out partition recovery with the Admin API. (API doc updates were done in #153 ).
- This is also an accompanying doc to the nodewise recovery updates (separate PR) - an update will be needed to cross link the two docs.
- 
Questions for SME are added as comments in this PR

Closes https://github.com/redpanda-data/documentation-private/issues/2105

### DIFF
--- a/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
@@ -1,0 +1,30 @@
+= Partition recovery
+:description: Recover a single partition using the Admin API.
+
+You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can arise when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
+
+Partition recovery is carried out by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
+
+If you want to instead permanently remove failed brokers from a cluster, especially in situations where no surviving partition replicas remain, use nodewise recovery.  
+
+== Use the Admin API to recover a partition
+
+Make a POST request to the `/force_replicas` endpoint:
+
+[,bash]
+----
+curl -X POST http://localhost:9644/v1/debug/partitions/<namespace>/<topic>/<partition>/force_replicas \
+  -d '{
+    [
+        { "core": 0, "node_id": 0 } 
+    ]
+  }'
+----
+
+See xref:api:admin-api/#tag/Partitions/operation/force_update_partition_replicas[Redpanda Admin API] for a full reference.
+
+include::shared:partial$suggested-reading.adoc[]
+
+- https://redpanda.com/blog/raft-protocol-reconfiguration-solution/[Engineering a more robust Raft group reconfiguration^]
+
+
--- a/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
@@ -5,7 +5,7 @@ You can use the Redpanda Admin API to recover a partition that appears to be stu
 
 Partition recovery is carried out by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
 
-If you want to instead permanently remove failed brokers from a cluster, especially in situations where no surviving partition replicas remain, use nodewise recovery.  
+If you want to instead force recover partitions in bulk on a failed broker, use nodewise recovery.  
 
 == Use the Admin API to recover a partition
 
@@ -21,7 +21,8 @@ curl -X POST http://localhost:9644/v1/debug/partitions/<namespace>/<topic>/<part
   }'
 ----
 
-See xref:api:ROOT:admin-api.adoc#tag/Partitions/operation/force_update_partition_replicas[Redpanda Admin API] for a full reference.
+The request body must include the ID of the broker and the CPU core (shard ID). If there are `n` CPU cores on the machine, the value of `core` can be within the range `[0, n-1]`. You may use a random value within the range, or the least loaded shard.
+See the xref:reference:public-metrics-reference.adoc[public metrics reference] for metrics regarding CPU usage, and xref:api:ROOT:admin-api.adoc#tag/Partitions/operation/force_update_partition_replicas[Redpanda Admin API] for a full reference.
 
 include::shared:partial$suggested-reading.adoc[]
 
--- a/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
@@ -1,9 +1,9 @@
-= Partition recovery
+= Forced partition recovery
 :description: Recover a single partition using the Admin API.
 
 You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can arise when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
 
-Partition recovery is carried out by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
+Forced partition recovery is carried out by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Forced partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
 
 If you want to instead force recover partitions in bulk on a failed broker, use nodewise recovery.  
 
--- a/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
@@ -1,7 +1,7 @@
 = Forced partition recovery
 :description: Recover a single partition using the Admin API.
 
-You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can arise when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
+You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can occur when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
 
 Forced partition recovery is carried out by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Forced partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
 
--- a/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
@@ -21,7 +21,7 @@ curl -X POST http://localhost:9644/v1/debug/partitions/<namespace>/<topic>/<part
   }'
 ----
 
-See xref:api:admin-api/#tag/Partitions/operation/force_update_partition_replicas[Redpanda Admin API] for a full reference.
+See xref:api:ROOT:admin-api.adoc#tag/Partitions/operation/force_update_partition_replicas[Redpanda Admin API] for a full reference.
 
 include::shared:partial$suggested-reading.adoc[]
 
--- a/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
@@ -22,7 +22,7 @@ curl -X POST http://localhost:9644/v1/debug/partitions/<namespace>/<topic>/<part
 ----
 
 The request body must include the ID of the broker and the CPU core (shard ID). If there are `n` CPU cores on the machine, the value of `core` can be within the range `[0, n-1]`. You may use a random value within the range, or the least loaded shard.
-See the xref:reference:public-metrics-reference.adoc[public metrics reference] for metrics regarding CPU usage, and xref:api:ROOT:admin-api.adoc#tag/Partitions/operation/force_update_partition_replicas[Redpanda Admin API] for a full reference.
+See the xref:reference:public-metrics-reference.adoc[public metrics reference] for metrics regarding CPU usage, and xref:api:ROOT:admin-api.adoc#tag/Partitions/operation/force_update_partition_replicas[Redpanda Admin API] for additional detail.
 
 include::shared:partial$suggested-reading.adoc[]
 
--- a/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/partition-recovery.adoc
@@ -3,7 +3,7 @@
 
 You can use the Redpanda Admin API to recover a partition that is unavailable and has lost a majority of its replicas. This can occur when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
 
-Redpanda carries out forced partition recovery by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. 
+Redpanda performs forced partition recovery by promoting the best available replica to leader, making the partition available for produce and consume. Typically, Redpanda chooses the replica with the highest offset.  
 
 CAUTION: Forced partition recovery allows some potential data loss on the partition if the best available replica is out of sync, ending up in an unclean leader election. Use this operation with caution, and only when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
 
@@ -17,7 +17,7 @@ The following examples assume that partition 0 in topic `test` is unavailable an
 +
 [,bash]
 ----
-curl http://localhost:9644/v1/partitions/kafka/test | jq
+curl http://localhost:9644/v1/partitions/kafka/test
 ----
 +
 [,bash,role=no-copy]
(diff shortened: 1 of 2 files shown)


modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,30 @@
+= Partition recovery
+:description: Recover a single partition using the Admin API.
+
+You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can arise when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
+
+Partition recovery is carried out by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
+
+If you want to instead permanently remove failed brokers from a cluster, especially in situations where no surviving partition replicas remain, use nodewise recovery.
```
kbatuigas: Note to add nodewise recovery doc when available

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,30 @@
+= Partition recovery
+:description: Recover a single partition using the Admin API.
+
+You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can arise when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
+
+Partition recovery is carried out by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority.
```
kbatuigas: Please verify how RP picks the "best available" replica

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,30 @@
...
+
+If you want to instead permanently remove failed brokers from a cluster, especially in situations where no surviving partition replicas remain, use nodewise recovery.  
+
+== Use the Admin API to recover a partition
+
+Make a POST request to the `/force_replicas` endpoint:
+
+[,bash]
+----
+curl -X POST http://localhost:9644/v1/debug/partitions/<namespace>/<topic>/<partition>/force_replicas \
```
kbatuigas: Can we provide a better, working example?

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,30 @@
...
+  -d '{
+    [
+        { "core": 0, "node_id": 0 } 
+    ]
+  }'
```
kbatuigas, line 75: The request body schema requires an array of replica assignments. What is the definition of `core` and how do we find its values?

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
bharathv: lgtm.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
bharathv: > instead permanently remove failed brokers from a cluster

I'd rephrase this. node wise recovery is just bulk partition force recovery, it doesn't remove failed brokers. Instead of doing force recovery manually for every partition on the failed broker, node wise recovery provides a way to to perform that action in one go.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,30 @@
+= Partition recovery
```
bharathv: Forced partition recovery.

IMO force signifies that it is not a regular raft recovery.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
bharathv: core is the CPU core aka shard id. If there are n CPU cores on the machine, user has to pick a value in range [0, n-1], currently user has to pick a random value or find a least loaded shard (if they know what that is) and use that value.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
kbatuigas: I can change it, @mattschumpert any thoughts?

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,30 @@
...
+:description: Recover a single partition using the Admin API.
+
+You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can arise when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader.
```
Feediver1:
```suggestion
You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can occur when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
```

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
Feediver1: @greg-szrama should be reviewing this too.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,30 @@
...
+----
+curl -X POST http://localhost:9644/v1/debug/partitions/<namespace>/<topic>/<partition>/force_replicas \
+  -d '{
+    [
+        { "core": 0, "node_id": 0 } 
+    ]
+  }'
+----
+
+See xref:api:ROOT:admin-api.adoc#tag/Partitions/operation/force_update_partition_replicas[Redpanda Admin API] for a full reference.
```
Feediver1:
```suggestion
See xref:api:ROOT:admin-api.adoc#tag/Partitions/operation/force_update_partition_replicas[Redpanda Admin API] for additional detail.
```

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
mattschumpert: Agreed

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,31 @@
+= Partition recovery
+:description: Recover a single partition using the Admin API.
+
+You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can occur when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader.
```
mattschumpert: is 'appears to be stuck' the best way to describe it @bharathv ?

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
mattschumpert: @kbatuigas @bharathv this doesn't talk about specifying the actual partition ID which seems to be the most important thing!

I get that you have to specify the node and core (which seems awkward in itself), but what about the specific partitions.  We need to give an example with the actual partition ID you're using.

mattschumpert (changes requested):

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
kbatuigas: @mattschumpert Looks like the partition ID is specified in the request URL `/debug/partitions/{namespace}/{topic}/{partition}/force_replicas`. @bharathv can you please provide example values for the following?
- `namespace`
- `topic`
- `partition` 
- `node_id`

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
greg-szrama: Is this curl only or is there also an RPK command?

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,31 @@
+= Forced partition recovery
+:description: Recover a single partition using the Admin API.
+
+You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can occur when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
+
+Forced partition recovery is carried out by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Forced partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
+
+If you want to instead force recover partitions in bulk on a failed broker, use nodewise recovery.
```
greg-szrama: I'd rephrase this -- it's not so much when you want to force recover partitions as needing to do a recovery with whole dead nodes. Nodewise partition recovery force-recovers a node, letting partitions which have a majority be rebuilt while starting fresh partitions where there is no majority.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
bharathv: I think "leaderless because of lost majority of replicas" I think is the right way (please rephrase as you see fit).

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
bharathv: > Nodewise partition recovery force-recovers a node

Just to be clear it doesn't recover the node itself (perhaps you meant the same thing but just want to call out incase it is not obvious to someone), it just rebuilds the partition replicas on those dead nodes elsewhere on other available brokers.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
bharathv: there is no rpk command that I'm aware of

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
bharathv: Right, partition ID is in the URL,


Here is a working example, lets assume partition test/0 is leaderless because majority of replicas are gone. 

A user is expected to get the current node/shard assignments using the following command

```
curl http://$(hostname):9644/v1/partitions/kafka/test | jq

[
  {
    "ns": "kafka",
    "topic": "test",
    "partition_id": 0,
    "status": "done",
    "leader_id": 1,
    "raft_group_id": 1,
    "replicas": [
      {
        "node_id": 1,
        "core": 1
      },
      {
        "node_id": 3,
        "core": 1
      },
      {
        "node_id": 2,
        "core": 1
      }
    ]
  }
]
```


Lets assume nodes 2, 3 are dead and we want to replace them with nodes 4, 5, the command would look like this.


```
curl -X POST http://$(hostname):9644/v1/partitions/kafka/test/0/force_replicas  -H 'Content-Type: application/json' --data '[{"node_id": 1, "core": 1}, {"node_id": 4, "core": 1}, {"node_id": 5, "core": 0}]'
```

As you can see the input to the command is the new replica set that will force replace the current one. It includes {1, 1} (which ensures we recover data from it) and two new replica assignments {4, 1}, {5, 0} which will hydrate data from {1, 1}.

User has to pick least loaded cores and the system may later move around the replicas if they are imbalanced.

Clearly this a low level operation and has to be done with caution.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,31 @@
+= Forced partition recovery
+:description: Recover a single partition using the Admin API.
+
+You can use the Redpanda Admin API to recover a partition that appears to be stuck (for example, does not accept writes). This can occur when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
+
+Forced partition recovery is carried out by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Forced partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority.
```
greg-szrama:
```suggestion
Redpanda performs forced partition recovery by promoting the best available replica to leader, thereby making that partition available for produce and consume. The new leader is typically the replica with the highest offset. However, this allows some potential data loss on the partition if the best available replica is out of sync, resulting in an unclean leader election. Forced partition recovery should only be used when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
```

You have some passive voice in here, you might rephrase it a bit.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+== Use the Admin API to recover a partition
+
+The following examples assume that partition 0 in topic `test` is unavailable and its replicas cannot form a majority to elect a leader.
+
+. Call the xref:api:ROOT:admin-api.adoc#tag/Partitions/operation/get_topic_partitions[`/partitions/kafka/<topic>`] endpoint to determine the current broker and shard assignments of the partition replicas.
++
+[,bash]
+----
+curl http://localhost:9644/v1/partitions/kafka/test | jq
```
bharathv: nit: you can remove the `| jq` part, I used it to format the json

actual command is just

```
curl http://localhost:9644/v1/partitions/kafka/test
```

bharathv (approved):

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+Redpanda performs forced partition recovery by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. 
+
+CAUTION: Forced partition recovery allows some potential data loss on the partition if the best available replica is out of sync, ending up in an unclean leader election. Use this operation with caution, and only when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
+
+NOTE: If you want to instead force recover all partitions in bulk from a set of failed brokers, use nodewise recovery.
+
+== Use the Admin API to recover a partition
+
+The following examples assume that partition 0 in topic `test` is unavailable and its replicas cannot form a majority to elect a leader.
```
Feediver1, line 14:
```suggestion
The following examples are based on the assumption that partition 0 in topic `test` is unavailable and its replicas cannot form a majority to elect a leader.
```

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,86 @@
...
+      {
+        "node_id": 3,
+        "core": 1
+      }
+    ]
+  }
+]
+----
+
+. Suppose brokers 2 and 3 have failed and we want to replace them with brokers 4 and 5.
```
Feediver1:
```suggestion
. In this scenario, brokers 2 and 3 have failed and you want to replace them with brokers 4 and 5.
```

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,86 @@
+= Forced partition recovery
+:description: Recover a single partition using the Admin API.
+
+You can use the Redpanda Admin API to recover a partition that is unavailable and has lost a majority of its replicas. This can occur when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
+
+Redpanda performs forced partition recovery by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. 
+
+CAUTION: Forced partition recovery allows some potential data loss on the partition if the best available replica is out of sync, ending up in an unclean leader election. Use this operation with caution, and only when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
+
+NOTE: If you want to instead force recover all partitions in bulk from a set of failed brokers, use nodewise recovery.
```
Feediver1: Can you provide a link to the instructions for nodewise recovery?

Feediver1 (approved):
See updates in context

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
kbatuigas: This will be added when the doc is available cc @greg-szrama

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,86 @@
+= Forced partition recovery
+:description: Recover a single partition using the Admin API.
+
+You can use the Redpanda Admin API to recover a partition that is unavailable and has lost a majority of its replicas. This can occur when the partition replicas have lost glossterm:Raft[] consensus, for instance if brokers in a Raft group fail, preventing the group from reaching a majority and electing a new leader. 
+
+Redpanda performs forced partition recovery by promoting the best available replica to leader, allowing the partition to become available for produce and consume. Typically, Redpanda chooses the replica with the highest offset.
```
greg-szrama:
```suggestion
Redpanda performs forced partition recovery by promoting the best available replica to leader, making the partition available for produce and consume. Typically, Redpanda chooses the replica with the highest offset. 
```

Passive voice here

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+CAUTION: Forced partition recovery allows some potential data loss on the partition if the best available replica is out of sync, ending up in an unclean leader election. Use this operation with caution, and only when brokers have failed beyond recovery to the point that the remaining replicas cannot form a majority. 
+
+NOTE: If you want to instead force recover all partitions in bulk from a set of failed brokers, use xref:./nodewise-partition-recovery.adoc[nodewise recovery].
+
+== Use the Admin API to recover a partition
+
+The following examples are based on the assumption that partition 0 in topic `test` is unavailable and its replicas cannot form a majority to elect a leader.
+
+. Call the xref:api:ROOT:admin-api.adoc#tag/Partitions/operation/get_topic_partitions[`/partitions/kafka/<topic>`] endpoint to determine the current broker and shard assignments of the partition replicas.
```
greg-szrama, line 16: The example here is good, I'm just checking whether there's also an RPK command to do this. If so, let's reference it in this section as well.

greg-szrama (approved):
Looks good overall! Just one question in there and a small edit.

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
mattschumpert: I don't think so. @ztlpn ?

modules/manage/pages/cluster-maintenance/partition-recovery.adoc
```diff
@@ -0,0 +1,86 @@
...
+      {
+        "node_id": 3,
+        "core": 1
+      }
+    ]
+  }
+]
+----
+
+. In this scenario, brokers 2 and 3 have failed and you want to replace them with brokers 4 and 5.
```
mattschumpert: @kbatuigas maybe instead of 'replace them', 'we want to move the replicas from those nodes to brokers 4 and 5, which are healthy'

mattschumpert (approved):

mmaslankaprv (approved):

PR #243, opened 2024-01-11
Add max_transactions__per_coordinator to properties reference

Feediver1:
This PR fixes doc [issue-2097](https://github.com/redpanda-data/documentation-private/issues/2097).

While it's great we are adding this missing prop to the doc, would be ideal if we could also mention it in the appropriate contexts, along with any other specific guidance on its use. Thx.

[Preview](https://deploy-preview-243--redpanda-docs-preview.netlify.app/current/reference/cluster-properties/#max_transactions_per_coordinator)


modules/reference/pages/cluster-properties.adoc
```diff
@@ -1677,6 +1677,19 @@ Enable transactions (atomic writes).
...
 ---
 
+=== max_transactions_per_coordinator
+
+Specifies the maximum number of active transaction sessions (producers). When the threshold
+is exceeded, Redpanda terminates old sessions. When an idle producer corresponding to the terminated
+session wakes up and produces, its batches are rejected with the message `invalid producer epoch` or
+`invalid_producer_id_mapping`, depending on where it is in the transaction execution phase.
+
+*Default*: true
```
JFlath: The default is actually `std::numeric_limits<uint64_t>::max()`, i.e. `9223372036854775807` or functionally unlimited.

JFlath (changes requested):
I was actually under the impression the default was lower, I have my suspicions that we might want to actually lower this in future, it might be nice to have Core's take on things.

It's also important to call out that as this is *per coordinator*, the total number of txns in the cluster at any one time is `max_transactions_per_coordinator` * `transaction_coordinator_partitions` (which is now by default `50`). Perhaps `configuration.cc` should say the same actually - not sure what the preference is on having differences between docs and that file.

In terms of context, I think this is important to mention wherever we talk about running transactional workloads from clients. One particular gotcha is that this must be tuned to ~the expected number of *active* transactions that a user's clients will be running at any one time if they use clients where transaction ids are *not* reused.

That is, if you have this at default and then your clients make a new id for every transaction, you just keep adding more and never removing, and that bloats memory. By default I believe Flink works in this way.

JFlath (approved):

modules/develop/pages/transactions.adoc
```diff
@@ -19,18 +19,19 @@ By default, the `enable_transactions` cluster configuration property is set to t
...
 The two primary use cases for transactions are:
 
 * xref:develop:transactions.adoc#atomic-publishing-of-multiple-messages[Atomic (all or nothing) publishing of multiple messages]
 * xref:develop:transactions.adoc#exactly-once-stream-processing[Exactly-once stream processing]
 
+=== Transaction usage tips
+
+* A transaction can span partitions from different topics, and a topic can be deleted while there are active transactions on one or more of its partitions. In-flight transactions can detect deletion events, remove the deleted partitions (and related messages) from the transaction scope, and commit changes to the remaining partitions.
+* Ongoing transactions can prevent consumers from advancing. To avoid this, don't set transaction timeout (`transaction.timeout.ms` in Java client) to high values: the higher the timeout, the longer consumers may be blocked. By default, it's about a minute, but it's a client setting that depends on the client.
+* When upgrading, make sure to use maintenance mode with xref:manage:cluster-maintenance/rolling-upgrade.adoc[rolling upgrades].
```
micheleRP: You could instead make rolling upgrade a [glossterm link](https://deploy-preview-243--redpanda-docs-preview.netlify.app/current/reference/glossary/#rolling-upgrade) or have the link go direct to the [rolling upgrade section](https://deploy-preview-243--redpanda-docs-preview.netlify.app/current/upgrade/rolling-upgrade/#perform-a-rolling-upgrade)?

modules/develop/pages/transactions.adoc
```diff
@@ -19,18 +19,19 @@ By default, the `enable_transactions` cluster configuration property is set to t
...
+* When running transactional workloads from clients, it is recommended that you tune xref:reference:cluster-properties#max_transactions_per_coordinator[`max_transactions_per_coordinator`] to the expected number of active transactions that your clients are running at any given time (if your client transaction IDs are not reused). The total number of transactions in the cluster at any one time is `max_transactions_per_coordinator * transaction_coordinator_partitions` (default is 50). When the threshold is exceeded, Redpanda terminates old sessions. When an idle producer corresponding to the terminated session wakes up and produces, its batches are rejected with the message `invalid producer epoch` or `invalid_producer_id_mapping`, depending on where it is in the transaction execution phase.
```
micheleRP:
```suggestion
* When running transactional workloads from clients, tune xref:reference:cluster-properties#max_transactions_per_coordinator[`max_transactions_per_coordinator`] to the number of active transactions that you expect your clients to run at any given time (if your client transaction IDs are not reused). The total number of transactions in the cluster at any one time is `max_transactions_per_coordinator * transaction_coordinator_partitions` (default is 50). When the threshold is exceeded, Redpanda terminates old sessions. If an idle producer corresponding to the terminated session wakes up and produces, its batches are rejected with the message `invalid producer epoch` or `invalid_producer_id_mapping`, depending on where it is in the transaction execution phase.
```

modules/develop/pages/transactions.adoc
```diff
@@ -19,18 +19,19 @@ By default, the `enable_transactions` cluster configuration property is set to t
...
+Be aware that if you use the default and your clients create a new ID for every transaction, this will continue to accumulate and never remove, which will bloat memory.
```
micheleRP:
```suggestion
Be aware that if you keep the default as 50 and your clients create a new ID for every transaction, the total continues to accumulate, which bloats memory.
```
I don't completely understand this bullet. Not positive, but this is my suggested text.

micheleRP (approved):
left some suggestions, but looks good!

mmaslankaprv (approved):

modules/develop/pages/transactions.adoc
micheleRP:
```suggestion
* When upgrading, make sure to use maintenance mode with a glossterm:rolling upgrade[].
```

PR #246, opened 2024-01-12
New Admin API endpoints for OIDC

kbatuigas:
[Deploy preview ](https://deploy-preview-246--redpanda-docs-preview.netlify.app/api/admin-api.html#tag/Authentication)

- New OIDC endpoints
- Adds new Authentication tag

Closes https://github.com/redpanda-data/documentation-private/issues/1912

### DIFF
--- a/modules/ROOT/attachments/admin-api.yaml
+++ b/modules/ROOT/attachments/admin-api.yaml
@@ -285,7 +285,7 @@ paths:
     get:
       tags: 
         -  Authentication
-      summary: Verify identity of current user and resolve to a principal.
+      summary: Verify identity of current user and resolve to a principal
       description: Obtain the principal and details of the JWT passed in the bearer token.
       operationId: oidc_whoami   
       responses:
--- a/modules/ROOT/attachments/admin-api.yaml
+++ b/modules/ROOT/attachments/admin-api.yaml
@@ -281,6 +281,57 @@ paths:
         200:
           description: Delete user success
           content: {}
+  /security/oidc/whoami:
+    get:
+      tags: 
+        -  Authentication
+      summary: Verify identity of current user and resolve to a principal.
+      description: Obtain the principal and details of the JWT passed in the bearer token.
+      operationId: oidc_whoami   
+      responses:
+        200:
+          description: Successful response
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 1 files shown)


BenPope (approved):
LGTM. I wonder why `authentication` tag and not `security`, but I don't hold a strong opinion here, and I haven't checked if the path prefix matches the tag in any other part of the API.

kbatuigas:
> LGTM. I wonder why `authentication` tag and not `security`, but I don't hold a strong opinion here, and I haven't checked if the path prefix matches the tag in any other part of the API.

@BenPope I did consider using "Security," but I saw that we had a couple other endpoints in the `/security/` path that are under a different tag (`/security/users/` under "Users"), so I thought to follow suit and create a new tag. We can easily change this though; I'll run it by the rest of the team. Thanks!

modules/ROOT/attachments/admin-api.yaml
```diff
@@ -281,6 +281,57 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/oidc_whoami_response'            
+        401:
+          description: Unauthorized
+          content: {}      
+  /security/oidc/keys/cache_invalidate:
+    post:
+      tags: 
+        -  Authentication
+      summary: Reload the keys from the Identity Provider
```
Feediver1:
```suggestion
      summary: Reload the keys from the Identity Provider.
```

modules/ROOT/attachments/admin-api.yaml
```diff
@@ -281,6 +281,57 @@ paths:
...
+          description: Keys reload failed
+          content:
+            application/json:
+              schema: 
+                $ref: '#/components/schemas/oidc_keys_cache_invalidate_error_response'
+  /security/oidc/revoke:
+    post: 
+      tags: 
+        -  Authentication
+      summary: Reload the keys from the Identity Provider and disconnect clients
```
Feediver1:
```suggestion
      summary: Reload the keys from the Identity Provider and disconnect clients.
```

Feediver1: Added end punctuation because the first summary I came to had it. Regardless of the convention you choose, just be consistent.

Feediver1 (approved):
lgtm

PR #247, opened 2024-01-12
Rephrase 'no data loss'

kbatuigas:
"No data loss" needs to be rephrased as it is too broad and we typically mean "safe when using recommended settings". 

We already cover the difference between Kafka and RP for acks=all in this doc https://docs.redpanda.com/current/develop/produce-data/configure-producers/#acksall

Closes https://github.com/redpanda-data/documentation-private/issues/47


modules/get-started/pages/architecture.adoc
```diff
@@ -40,7 +40,7 @@ https://raft.github.io/[Raft^] uses a heartbeat mechanism to maintain leader aut
...
+If a follower triggers an election, but the incumbent leader subsequently springs back to life and starts sending data again, then it's too late. As part of the election process, the follower (now an election candidate) incremented the term and rejects requests from the previous term, essentially forcing a leadership change. If a cluster is experiencing wider network infrastructure problems that result in latencies above the heartbeat timeout, then back-to-back election rounds can be triggered. During this period, unstable Raft groups may not be able to form a quorum. This results in partitions rejecting writes, but previously written data will not be lost. Redpanda has a Raft-priority implementation that allows the system to settle quickly after network outages.
```
hcoyote: very minor tweak to the wording, otherwise +1 on everything. 

> but data previously written to disk will not be lost

this reinforces the fact that as long as we've gotten it down to disk, we should be safe.

hcoyote (approved):

modules/get-started/pages/architecture.adoc
micheleRP:
```suggestion
If a follower triggers an election, but the incumbent leader subsequently springs back to life and starts sending data again, then it's too late. As part of the election process, the follower (now an election candidate) incremented the term and rejects requests from the previous term, essentially forcing a leadership change. If a cluster is experiencing wider network infrastructure problems that result in latencies above the heartbeat timeout, then back-to-back election rounds can be triggered. During this period, unstable Raft groups may not be able to form a quorum. This results in partitions rejecting writes, but previously written data is not lost. Redpanda has a Raft-priority implementation that allows the system to settle quickly after network outages.
```

micheleRP (approved):
looks good with little suggestion

PR #258, opened 2024-01-18
Single-source command to install CRDs

JakeSCahill:
Fixes https://github.com/redpanda-data/documentation-private/issues/2190

Uses the [release version fetched from GitHub](https://github.com/redpanda-data/docs-extensions-and-macros/pull/28) in all install commands for the RP Operator chart and CRDs.

[See an example in the live preview](https://deploy-preview-258--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/self-hosted/kubernetes/k-production-deployment/#deploy-a-redpanda-cluster:~:text=Install%20the%20Redpanda%20Operator%20CRDs%3A).

### DIFF
--- a/modules/deploy/partials/kubernetes/install-crds.adoc
+++ b/modules/deploy/partials/kubernetes/install-crds.adoc
@@ -1,0 +1,14 @@
+ifdef::latest-operator-version[]
+[,bash,subs="attributes+"]
+----
+kubectl kustomize "https://github.com/redpanda-data/redpanda-operator//src/go/k8s/config/crd?ref={latest-operator-version}" \
+    | kubectl apply -f -
+----
+endif::[]
+ifndef::latest-operator-version[]
+[,bash]
+----
+kubectl kustomize "https://github.com/redpanda-data/redpanda-operator//src/go/k8s/config/crd" \
+    | kubectl apply -f -
+----
+endif::[]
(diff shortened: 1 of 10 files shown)


modules/deploy/partials/kubernetes/install-crds.adoc
```diff
@@ -0,0 +1,14 @@
...
+[,bash,subs="attributes+"]
+----
+kubectl kustomize "https://github.com/redpanda-data/redpanda-operator//src/go/k8s/config/crd?ref={latest-operator-version}" \
+    | kubectl apply -f -
+----
+endif::[]
+ifndef::latest-operator-version[]
+[,bash]
+----
+kubectl kustomize "https://github.com/redpanda-data/redpanda-operator//src/go/k8s/config/crd" \
```
chrisseto: Do you think it would be preferable to throw an error rather than falling back to the previous buggy template?

Alternatively, we could update this version to instead have a placeholder for the end user to fill out.

chrisseto (approved):
LGTM!

modules/deploy/partials/kubernetes/install-crds.adoc
JakeSCahill: Great idea. Will add the placeholder.

Feediver1:
so the purple highlighting is just to help reviewers can see that it is the latest version of RP Operator?

Feediver1 (approved):
lgtm

PR #259, opened 2024-01-18
Update cluster-properties.adoc

jrkinley:
Related to https://github.com/redpanda-data/helm-charts/issues/973.

Preview:
https://deploy-preview-259--redpanda-docs-preview.netlify.app/current/reference/cluster-properties

### DIFF
--- a/modules/reference/pages/cluster-properties.adoc
+++ b/modules/reference/pages/cluster-properties.adoc
@@ -192,8 +192,8 @@ AWS or GCP access key.
 
 Optional API endpoint.
 
-- For AWS, this can be left blank. It’s generated automatically using the region and bucket.
-- For GCS, use `storage.googleapis.com`.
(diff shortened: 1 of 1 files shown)


modules/reference/pages/cluster-properties.adoc
```diff
@@ -192,6 +192,9 @@ AWS or GCP access key.
 
 Optional API endpoint.
 
+- For AWS, this can be left blank. It’s generated automatically using the region and bucket.
+- For GCS, use `storage.googleapis.com`.
```
greg-szrama: Should this be GCS or GCP? I only ask because the above line is AWS where the product is S3 -- here GCS is the product on GCP.

greg-szrama (approved):
One point for clarification -- not sure if it needs altered or not so approving.

PR #261, opened 2024-01-19
Add new cluster config option minimum_topic_replications #2161

greg-szrama:
Resolves issue: https://github.com/redpanda-data/documentation-private/issues/2161

Added the property and references to it in several places.

[Property listing in cluster properties.](https://deploy-preview-261--redpanda-docs-preview.netlify.app/current/reference/cluster-properties/#minimum_topic_replications)

Cloud [create topic](https://deploy-preview-261--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/cloud/create-topic/)

Develop -> [config topics](https://deploy-preview-261--redpanda-docs-preview.netlify.app/current/develop/config-topics/#choose-the-replication-factor)

### DIFF
--- a/modules/deploy/pages/deployment-option/cloud/create-topic.adoc
+++ b/modules/deploy/pages/deployment-option/cloud/create-topic.adoc
@@ -19,7 +19,7 @@ After you create a cluster, you can create a topic for that cluster.
 | The number of partitions for the topic.
 
 | *Replication factor*
-| The number of partition replicas for the topic. Three replicas are recommended. The cluster property `minimum_topic_replication`, if set, provides a lower bound for this property.
+| The number of partition replicas for the topic. Three replicas are recommended. The value for this property must meet or exceed the cluster's xref:reference:cluster-properties.adoc#minimum_topic_replication[`minimum_topic_replication`] setting.
 
 | *Cleanup policy*
 | The policy that determines how to clean up old log segments. The default is *delete*.
--- a/modules/deploy/pages/deployment-option/cloud/create-topic.adoc
+++ b/modules/deploy/pages/deployment-option/cloud/create-topic.adoc
@@ -19,7 +19,7 @@ After you create a cluster, you can create a topic for that cluster.
 | The number of partitions for the topic.
 
 | *Replication factor*
-| The number of partition replicas for the topic. Three replicas are recommended.
+| The number of partition replicas for the topic. Three replicas are recommended. The cluster property `minimum_topic_replicas`, if set, provides a lower bound for this property.
 
 | *Cleanup policy*
 | The policy that determines how to clean up old log segments. The default is *delete*.
--- a/modules/deploy/pages/deployment-option/cloud/create-topic.adoc
+++ b/modules/deploy/pages/deployment-option/cloud/create-topic.adoc
@@ -19,7 +19,7 @@ After you create a cluster, you can create a topic for that cluster.
 | The number of partitions for the topic.
 
 | *Replication factor*
-| The number of partition replicas for the topic. Three replicas are recommended. The cluster property `minimum_topic_replicas`, if set, provides a lower bound for this property.
+| The number of partition replicas for the topic. Three replicas are recommended. The cluster property `minimum_topic_replication`, if set, provides a lower bound for this property.
 
 | *Cleanup policy*
 | The policy that determines how to clean up old log segments. The default is *delete*.
--- a/modules/develop/pages/config-topics.adoc
+++ b/modules/develop/pages/config-topics.adoc
@@ -45,7 +45,7 @@ To specify a replication factor of 3 for topic "`xyz`":
 rpk topic create xyz -r 3
 ----
 
-NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases. Administrators may set a lower boundary for replication factor through the cluster level `minimum_topic_replications` property.
+NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases. Administrators may set a lower boundary for replication factor through the cluster level `minimum_topic_replication` property.
 
 == Update topic configurations
 
--- a/modules/develop/pages/config-topics.adoc
+++ b/modules/develop/pages/config-topics.adoc
@@ -45,7 +45,7 @@ To specify a replication factor of 3 for topic "`xyz`":
 rpk topic create xyz -r 3
 ----
 
-NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases.
+NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases. Administrators may set a lower boundary for replication factor through the cluster level `minimum_topic_replications` property.
 
 == Update topic configurations
 
--- a/modules/develop/pages/config-topics.adoc
+++ b/modules/develop/pages/config-topics.adoc
@@ -45,7 +45,7 @@ To specify a replication factor of 3 for topic "`xyz`":
 rpk topic create xyz -r 3
 ----
 
-NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases. Administrators may set a lower boundary for replication factor through the cluster level `minimum_topic_replication` property.
+NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases. Administrators may set a minimum required replication factor for any new topic in the cluster through the cluster-level xref:reference:cluster-properties.adoc#minimum_topic_replication[`minimum_topic_replication`] property.
 
 == Update topic configurations
 
--- a/modules/reference/pages/cluster-properties.adoc
+++ b/modules/reference/pages/cluster-properties.adoc
@@ -1642,7 +1642,7 @@ Target replication factor for internal topics.
 
 Minimum permitted replication factor value for newly-created topics. Redpanda returns an `INVALID_REPLICATION_FACTOR` error on any attempt to create a topic with a replication factor less than this property.
 
-When changing the `minimum_topic_replications` parameter, any previously existing topics having a replication factor less than the new value will continue to work as normal. Redpanda will, however, issue a warning level log message at start up indicating the existence of topics with a replication factor less than the cluster's `minimum_topic_replications`. That error message will read, `Topic X has a replication factor less than specified minimum: {} < {}`.
+When changing the `minimum_topic_replication` parameter, any previously existing topics having a replication factor less than the new value will continue to work as normal. Redpanda will, however, issue a warning level log message at start up indicating the existence of topics with a replication factor less than the cluster's `minimum_topic_replication`. That error message will read, for example, `Topic X has a replication factor less than specified minimum: 1 < 3`.
 
 *Units*: minimum number of replicas per topic
 
--- a/modules/reference/pages/cluster-properties.adoc
+++ b/modules/reference/pages/cluster-properties.adoc
@@ -1638,7 +1638,7 @@ Target replication factor for internal topics.
 
 ---
 
-=== minimum_topic_replications
+=== minimum_topic_replication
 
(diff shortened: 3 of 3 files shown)


modules/reference/pages/cluster-properties.adoc
```diff
@@ -1638,6 +1638,20 @@ Target replication factor for internal topics.
 
 ---
 
+=== minimum_topic_replications
+
+Minimum permitted replication factor value for newly created topics. Redpanda issues an error for any attempt to create a topic with a replication factor less than the specified value. Any existing topics found to have a replication factor less than the specified value will still continue to work as normal. Redpanda will issue a warning level log message at start up indicating the topic's replication factor is less than the specified minimum.
```
Feediver1:
```suggestion
Minimum permitted replication factor value for newly-created topics. Redpanda returns an error for any attempt to create a topic with a replication factor less than the specified value. Any existing topics found to have a replication factor less than the specified value will continue to work as normal. Redpanda will issue a warning level log message at start up indicating the topic's replication factor is less than the specified minimum.
```

Feediver1: When you say "specified value", is that the default of 3? In other words, is the issue that users must stick with the default of 3? Or, can another value be set by a cluster admin and that is the value that is "specified"? I guess I'm wondering if "specified" and "default" are one and the same?

mattschumpert: @michael-redpanda what is the Kafka API error code/message that we return when this is violated at topic creation?  Would be good to document this.

Also what is the log message when exiting topics violate it.  Would be good to add this detail for users to follow. @greg-szrama

mattschumpert (approved):

modules/reference/pages/cluster-properties.adoc
Feediver1: +1 on documenting the msgs here.

michael-redpanda: > what is the Kafka API error code/message that we return when this is violated at topic creation? Would be good to document this.

The rc is "INVALID_REPLICATION_FACTOR" (38) - message is "Replication factor must be greater than or equal to specified minimum value"

> Also what is the log message when exiting topics violate it.
warning level on the cluster logger facility: "Topic X has a replication factor less than specified minimum: {} < {}"

Should be printed on startup and whenever the config value changes.

modules/deploy/pages/deployment-option/cloud/create-topic.adoc
```diff
@@ -19,7 +19,7 @@ After you create a cluster, you can create a topic for that cluster.
 | The number of partitions for the topic.
 
 | *Replication factor*
-| The number of partition replicas for the topic. Three replicas are recommended.
+| The number of partition replicas for the topic. Three replicas are recommended. The cluster property `minimum_topic_replicas`, if set, provides a lower bound for this property.
```
michael-redpanda:
```suggestion
| The number of partition replicas for the topic. Three replicas are recommended. The cluster property `minimum_topic_replication`, if set, provides a lower bound for this property.
```

modules/develop/pages/config-topics.adoc
```diff
@@ -45,7 +45,7 @@ To specify a replication factor of 3 for topic "`xyz`":
 rpk topic create xyz -r 3
 ----
 
-NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases.
+NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases. Administrators may set a lower boundary for replication factor through the cluster level `minimum_topic_replications` property.
```
michael-redpanda:
```suggestion
NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases. Administrators may set a lower boundary for replication factor through the cluster level `minimum_topic_replication` property.
```

modules/reference/pages/cluster-properties.adoc
```diff
@@ -1638,6 +1638,20 @@ Target replication factor for internal topics.
 
 ---
 
+=== minimum_topic_replications
```
michael-redpanda:
```suggestion
=== minimum_topic_replication
```

modules/reference/pages/cluster-properties.adoc
```diff
@@ -1638,6 +1638,20 @@ Target replication factor for internal topics.
...
+
+Minimum permitted replication factor value for newly-created topics. Redpanda returns an error on any attempt to create a topic with a replication factor less than this property. When changing the `minimum_topic_replications` parameter, any previously existing topics having a replication factor less than the new value will continue to work as normal. Redpanda will, however, issue a warning level log message at start up indicating the existence of a topic with a replication factor less than the cluster's `minimum_topic_replications`.
```
michael-redpanda:
```suggestion
Minimum permitted replication factor value for newly-created topics. Redpanda returns an error on any attempt to create a topic with a replication factor less than this property. When changing the `minimum_topic_replication` parameter, any previously existing topics having a replication factor less than the new value will continue to work as normal. Redpanda will, however, issue a warning level log message at start up indicating the existence of a topic with a replication factor less than the cluster's `minimum_topic_replication`.
```

mattschumpert (approved):

modules/reference/pages/cluster-properties.adoc
```diff
@@ -1638,16 +1638,20 @@ Target replication factor for internal topics.
...
+Minimum permitted replication factor value for newly-created topics. Redpanda returns an `INVALID_REPLICATION_FACTOR` error on any attempt to create a topic with a replication factor less than this property.
+
+When changing the `minimum_topic_replications` parameter, any previously existing topics having a replication factor less than the new value will continue to work as normal. Redpanda will, however, issue a warning level log message at start up indicating the existence of topics with a replication factor less than the cluster's `minimum_topic_replications`. That error message will read, `Topic X has a replication factor less than specified minimum: {} < {}`.
```
michael-redpanda:
```suggestion
When changing the `minimum_topic_replication` parameter, any previously existing topics having a replication factor less than the new value will continue to work as normal. Redpanda will, however, issue a warning level log message at start up indicating the existence of topics with a replication factor less than the cluster's `minimum_topic_replication`. That error message will read, `Topic X has a replication factor less than specified minimum: {} < {}`.
```

michael-redpanda (commented):
sorry two more `minimum_topic_replications` to `minimum_topic_replication`

modules/reference/pages/cluster-properties.adoc
michael-redpanda: Is there a better way of describing this warning message other than "{} < {}"?  Maybe just an example of "1 < 3" or something?

greg-szrama: Yeah, I'll just reword it as a representative example and put actual numbers in there. I also fixed those last minimum_topic_replications instances and made sure there weren't any more. Silly mistake to make, my apologies!

modules/deploy/pages/deployment-option/cloud/create-topic.adoc
```diff
@@ -19,7 +19,7 @@ After you create a cluster, you can create a topic for that cluster.
 | The number of partitions for the topic.
 
 | *Replication factor*
-| The number of partition replicas for the topic. Three replicas are recommended.
+| The number of partition replicas for the topic. Three replicas are recommended. The cluster property `minimum_topic_replication`, if set, provides a lower bound for this property.
```
Deflaimun: we could add a link to the property reference

michael-redpanda (approved):

Deflaimun:
on rpk topic create reference we have a message saying 

> The replication factor must be positive, odd and equal or less than the number of available brokers.

I assume this is also the case here, right?

modules/reference/pages/cluster-properties.adoc
```diff
@@ -1638,6 +1638,24 @@ Target replication factor for internal topics.
 
 ---
 
+=== minimum_topic_replication
+
+Minimum permitted replication factor value for newly-created topics. Redpanda returns an `INVALID_REPLICATION_FACTOR` error on any attempt to create a topic with a replication factor less than this property.
```
Deflaimun: What about something more direct? 
```suggestion
Minimum replication factor for topics. Changing this parameter only affects newly-created topics.

Redpanda returns an `INVALID_REPLICATION_FACTOR` error on any attempt to create a topic with a replication factor less than this property.
```

modules/deploy/pages/deployment-option/cloud/create-topic.adoc
JakeSCahill:
```suggestion
| The number of partition replicas for the topic. Three replicas are recommended. The specified number of replicas for a topic must meet or exceed the cluster's `minimum_topic_replication` setting, which defines the minimum required replication factor for any new topic in the cluster.
```

JakeSCahill: agree with Paulo on the link

modules/develop/pages/config-topics.adoc
```diff
@@ -45,7 +45,7 @@ To specify a replication factor of 3 for topic "`xyz`":
 rpk topic create xyz -r 3
 ----
 
-NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases.
+NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases. Administrators may set a lower boundary for replication factor through the cluster level `minimum_topic_replication` property.
```
JakeSCahill:
```suggestion
NOTE: The replication factor must be an odd number. Redpanda Data recommends a replication factor of 3 for most use cases. Administrators may set a minimum required replication factor for any new topic in the cluster through the cluster-level `minimum_topic_replication` property.
```

modules/reference/pages/cluster-properties.adoc
```diff
@@ -1638,6 +1638,24 @@ Target replication factor for internal topics.
...
+
+When changing the `minimum_topic_replication` parameter, any previously existing topics having a replication factor less than the new value will continue to work as normal. Redpanda will, however, issue a warning level log message at start up indicating the existence of topics with a replication factor less than the cluster's `minimum_topic_replication`. That error message will read, for example, `Topic X has a replication factor less than specified minimum: 1 < 3`.
```
JakeSCahill:
```suggestion
If you change the `minimum_topic_replication` setting, the replication factor of existing topics remains unchanged. But, Redpanda will log a warning on start-up with a list of any topics that have fewer replicas than this minimum. For example, you might see a message such as `Topic X has a replication factor less than specified minimum: 1 < 3`.
```

Deflaimun: What do we mean continue to work as normal?

Will they have their replication factor altered? If no, then changing the parameter only affects new topics. Then this passage might be superfluous no?  We can change to something along those lines


"Existing topics are not affected by this change.  Redpanda will, however, issue a warning level log message at start up indicating the existence of topics with a replication factor less than the cluster's `minimum_topic_replication`. That error message will read, for example, `Topic X has a replication factor less than specified minimum: 1 < 3`."

Tech question, can we modify the number of replication factor for a topic after its creation? If yes, we should document it here.

Something like "to modify the topic replication factor, see: bla"

greg-szrama: The point here was to make sure users know setting this won't update any other existing topics. If you have a topic with a replication factor of 1 and you set the minimum to 3, the previously existing topic will not be updated and it will continue functioning with a replication factor of 1. You'll just get error messages indicating the existence of a topic that breaks the configured lower bound.

JakeSCahill (approved):
Some suggestions but looks good to me 😄

modules/develop/pages/config-topics.adoc
JakeSCahill: should also be a link

modules/reference/pages/cluster-properties.adoc
Deflaimun: Yup. So I would change the passage that says `continue to work as normal` . Would include something along the lines of "modifying this parameter don't change the replication factor for existing topics"

PR #268, opened 2024-01-20
Documented node-wise partition recovery

greg-szrama:
Resolves: https://github.com/redpanda-data/documentation-private/issues/1982

[Preview](https://deploy-preview-268--redpanda-docs-preview.netlify.app/current/manage/cluster-maintenance/nodewise-partition-recovery/)

### DIFF
--- a/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
@@ -14,7 +14,7 @@ This feature has two prerequisites for recovering partitions:
 
 Executing a node-wise partition recovery with no nodes or partitions meetings these criteria will result in Redpanda performing no action.
 
-== Performing the recovery operation
+== Perform the recovery operation
 
 Node-wise partition recovery is performed via the RPK command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. You may lose data as a result of executing this operation.
 
--- a/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
@@ -15,7 +15,7 @@ The syntax for this command is as follows:
 
  rpk cluster partitions unsafe-recover --from-nodes 1,3,5
 
-The `--from-nodes` parameter accepts a comma-delineated list of dead node IDs you wish to recover the data from. The above example would perform recovery operations on nodes 1, 3, and 5. Redpanda will assess these nodes to identify which partitions lack majority. It will then create a plan to recover the impacted partitions and ask you to confirm it. You must respond `yes` to the prompt to continue with recovery.
+The `--from-nodes` parameter accepts a comma-delineated list of dead node IDs you wish to recover the data from. The above example would perform recovery operations on nodes 1, 3, and 5. Redpanda will assess these nodes to identify which partitions lack majority. It will then create a plan to recover the impacted partitions and prompt you to confirm it. You must respond `yes` to continue with recovery.
 
 You may also optionally add the `--dry` parameter to this command. This will perform a dry run and allow viewing the recovery plan with no risk to your cluster.
 
--- a/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
@@ -1,28 +1,31 @@
 = Node-wise Partition Recovery
-:description: Feature to recover dead nodes.
+:description: Feature to recover partitions that have lost a majority of replicas.
 
 Multi-node or entire-AZ failures (especially in cloud environments) and some forms of human error may result in ‘stuck’ partitions with fewer replicas than required to make a quorum. In these scenarios, some data loss is unavoidable but admins need to recover what they can and move on.
 
 IMPORTANT: This is a last ditch measure when all other recovery options have failed. In some cases, no remaining replicas may exist for the partitions on the dead nodes. This recovery method is used in scenarios where you are already experiencing data loss - the goal is to stop the loss of any more data.
 
-== Prerequisites
+== Assumptions
 
-This feature has two prerequisites for recovering partitions:
+This capability only attempts to recover data from partitions where the following conditions are met:
 
-* Your cluster must have one or more `dead` nodes.
+* Your cluster must have one or more irrecoverable nodes.
 * One or more partitions on these nodes must have no majority.
+* Some of the data on these partitions is no longer recoverable.
 
-Executing a node-wise partition recovery with no nodes or partitions meetings these criteria will result in Redpanda performing no action.
+Using this command allows you to unsafely recover at least a portion of your data using whatever replicas are remaining. If no partitions satisfy these criteria, the unsafe partition recovery function exits with no action performed.
 
 == Perform the recovery operation
 
-Node-wise partition recovery is performed via the RPK command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. You may lose data as a result of executing this operation.
+Node-wise partition recovery is performed via the `rpk` command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. You may lose some data which was not yet replicated to the surviving partition replicas as a result of executing this operation.
 
 The syntax for this command is as follows:
 
  rpk cluster partitions unsafe-recover --from-nodes 1,3,5
 
 The `--from-nodes` parameter accepts a comma-delineated list of dead node IDs you wish to recover the data from. The above example would perform recovery operations on nodes 1, 3, and 5. Redpanda will assess these nodes to identify which partitions lack majority. It will then create a plan to recover the impacted partitions and ask you to confirm it. You must respond `yes` to the prompt to continue with recovery.
+
+You may also optionally add the `--dry` parameter to this command. This will perform a dry run and allow viewing the recovery plan with no risk to your cluster.
 
 Once the recovery operation is started, you may monitor the status of its execution via the `rpk cluster partitions balancer-status` command. The recovery operation can take some time to complete, especially when a lot of data is involved, and this command allows you to monitor progress in real-time.
 
--- a/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
@@ -1,9 +1,9 @@
 = Node-wise Partition Recovery
 :description: Feature to recover partitions that have lost a majority of replicas.
 
-Multi-node or entire-AZ failures (especially in cloud environments) and some forms of human error may result in ‘stuck’ partitions with fewer replicas than required to make a quorum. In these scenarios, some data loss may be unavoidable but admins need to recover what they can and move on.
+Multi-node or entire-AZ failures (especially in cloud environments), and some forms of human error may result in ‘stuck’ partitions with fewer replicas than required to make a quorum. In such failure scenarios, some data loss may be unavoidable. This description of node-wise partition recovery helps admins understand what they can or cannot recover.
 
-IMPORTANT: This is a last ditch measure when all other recovery options have failed. In some cases, no remaining replicas may exist for the partitions on the dead nodes. This recovery method is used in scenarios where you are already experiencing data loss - the goal is to stop the loss of any more data.
+IMPORTANT: This is a last-ditch measure when all other recovery options have failed. In some cases, no remaining replicas may exist for the partitions on the dead nodes. This recovery method is intended for scenarios where you are already experiencing data loss - the goal is to stop the loss of additional data.
 
 Node-wise partition recovery allows you to unsafely recover at least a portion of your data using whatever replicas are remaining. All replicas are moved off the target nodes and allocated to healthy nodes. In one step, this process repairs partitions while draining the target nodes.
 
--- a/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
@@ -9,7 +9,7 @@ Node-wise partition recovery allows you to unsafely recover at least a portion o
 
 == Perform the recovery operation
 
-Node-wise partition recovery is performed via the `rpk` command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. You may lose some data which was not yet replicated to the surviving partition replicas as a result of executing this operation.
+You perform node-wise partition recovery using the `rpk` command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. As a result of executing this operation, you may lose some data that has not yet been replicated to the surviving partition replicas.
 
 The syntax for this command is as follows:
 
--- a/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
+++ b/modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
@@ -19,7 +19,7 @@ The `--from-nodes` parameter accepts a comma-delineated list of dead node IDs yo
 
 You may also optionally add the `--dry` parameter to this command. This will perform a dry run and allow viewing the recovery plan with no risk to your cluster.
 
-Once the recovery operation is started, you may monitor the status of its execution via the `rpk cluster partitions balancer-status` command. The recovery operation can take some time to complete, especially when a lot of data is involved, and this command allows you to monitor progress in real-time.
+Once the recovery operation is started, you may monitor the status of its execution using the `rpk cluster partitions balancer-status` command. The recovery operation can take some time to complete, especially when a lot of data is involved. This command allows you to monitor progress in real-time.
 
 == Example recovery operation
 Here's an example of the recovery process in action.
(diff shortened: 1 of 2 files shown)


modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+== Performing the recovery operation
+
+Node-wise partition recovery is performed via the RPK command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. You may lose data as a result of executing this operation.
+
+The syntax for this command is as follows:
+
+ rpk cluster partitions unsafe-recover --from-nodes 1,3,5
+
+The `--from-nodes` parameter accepts a comma-delineated list of node IDs you wish to recover. The above example would perform recovery operations on nodes 1, 3, and 5. Redpana will assess these nodes to identify which partitions lack majority. It will then create a plan to recover the impacted partitions and ask you to confirm it. You must respond `yes` to the prompt to continue with recovery.
```
r-vasquez:
```suggestion
The `--from-nodes` parameter accepts a comma-delineated list of dead node IDs you wish to recover the data from. The above example would perform recovery operations on nodes 1, 3, and 5. Redpanda will assess these nodes to identify which partitions lack majority. It will then create a plan to recover the impacted partitions and ask you to confirm it. You must respond `yes` to the prompt to continue with recovery.
```

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+Once the recovery operation is started, you may monitor the status of its execution via the `rpk cluster partitions balancer-status` command. The recovery operation can take some time to complete, especially when a lot of data is involved, and this command allows you to monitor progress in real-time.
```
r-vasquez: I would add a section to specify that a user can perform a dry run (`--dry`) to identify "the plan" (i.e, the partitions that lost majority due to the irrecoverable nodes).

Since this is an unsafe operation, the availability of this flag is very important.

r-vasquez (commented):
rpk commands look good to me, just a few suggestions.

I would love the review of @bharathv too, that's why I'm not marking it as approved. :+1:

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,45 @@
+= Node-wise Partition Recovery
+:description: Feature to recover dead nodes.
```
mattschumpert: 'Feature to recover partitions that have lost a majority of replicas'

This does not recover dead nodes. It recovers the data that was partially located on the dead nodes. the nodes themselves are gone and will never be recovered, neither physically (the instance) or logically (the node ID in redpanda).

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+Multi-node or entire-AZ failures (especially in cloud environments) and some forms of human error may result in ‘stuck’ partitions with fewer replicas than required to make a quorum. In these scenarios, some data loss is unavoidable but admins need to recover what they can and move on.
+
+IMPORTANT: This is a last ditch measure when all other recovery options have failed. In some cases, no remaining replicas may exist for the partitions on the dead nodes. This recovery method is used in scenarios where you are already experiencing data loss - the goal is to stop the loss of any more data.
+
+== Prerequisites
+
+This feature has two prerequisites for recovering partitions:
+
+* Your cluster must have one or more `dead` nodes.
```
mattschumpert: @ztlpn is this really even true? The dead nodes could have been removed from the cluster long ago and there are still partitions without majority that need to be repaired.

We need to remove the emphasis on dead nodes here.

There are also NO pre-requisites to using this feature. you can use it when nothing is wrong at all. It will just skip healthy partitions.  It basically SCANS the nodes for anything without majority and repairs these partitions. That's it.

Lets drop this 'prerequisite' section entirely, its not needed and a bit misleading

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,45 @@
...
+* One or more partitions on these nodes must have no majority.
+
+Executing a node-wise partition recovery with no nodes or partitions meetings these criteria will result in Redpanda performing no action.
```
mattschumpert: this statement refutes the whole prerequisite section above.

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+== Performing the recovery operation
+
+Node-wise partition recovery is performed via the RPK command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. You may lose data as a result of executing this operation.
```
mattschumpert: lose 'some data which was not yet replicated to the surviving partition replicas as a result of executing this operation'

besides that this looks good.

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
mattschumpert: 👍

mattschumpert (changes requested):

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
ztlpn: Summoning @bharathv as the SME

mattschumpert: Oops sorry @ztlpn

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
kbatuigas: We could double check with the rest of the team but I  don't think we use capitalization, and really only monospace with `rpk` in the docs.

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,45 @@
...
+== Prerequisites
+
+This feature has two prerequisites for recovering partitions:
+
+* Your cluster must have one or more `dead` nodes.
+* One or more partitions on these nodes must have no majority.
+
+Executing a node-wise partition recovery with no nodes or partitions meetings these criteria will result in Redpanda performing no action.
+
+== Performing the recovery operation
```
kbatuigas:
```suggestion
== Perform the recovery operation
```
imperative for headings

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
bharathv: > is this really even true? The dead nodes could have been removed from the cluster long ago and there are still partitions without majority that need to be repaired.

Technically that is not possible because decommissioning a dead node requires draining all replicas on it and a partition with lost majority cannot do it. That said, I agree that we need remove too much emphasis on the dead nodes here (see below).

> There are also NO pre-requisites to using this feature. you can use it when nothing is wrong at all. It will just skip healthy partitions. It basically SCANS the nodes for anything without majority and repairs these partitions. That's it.

@mattschumpert  A small clarification, when it scans the topic list, the decision of majority loss is based on the user input nodes. Example, partition p has replicas [0, 1, 2] and user says "unsafe recover 1,2", Redpanda knows that partition `p` lost majority because majority [1, 2] of its current replica assignment [0,1,2] are gone. In this process, the logic actually is based on the user input to recover partitions from nodes [1,2]. Without that the broker has no way of knowing whether nodes 1, 2 are dead or down temporarily or partitioned from the cluster etc.

So, I wouldn't say there are NO prereqs. The prereq here is that the nodes are irrecoverable (data is gone for whatever reason, dead, corruption, vanished etc) and the user has confirmed that with their tooling and the user has to communicate that to Redpanda with this command.


The way I'd phrase it is, if there is a majority loss of partitions because one or more nodes are no longer accessible and the data is irrecoverable, we can use this command to unsafely recover them using whatever replicas are remaining. The reason I think `dead nodes` is confusing is because there is no technical definition of what `dead` is. May be it is dead because someone shut it down but the data is recoverable once someone reboots it. So I think we should rather emphasis on "non recoverability" of data from those nodes.

wdyt.

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
bharathv: nit: Redpanda typo.

mattschumpert: Makes sense.  'prerequisite' is a strange term though. Nothing prevents you from using it anytime.  But it relies on an _assumption_ by the user that these nodes not longer exist or are unrecoverable.

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
greg-szrama: I reworded this as an assumptions section, per this feedback and the feedback below.

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+IMPORTANT: This is a last ditch measure when all other recovery options have failed. In some cases, no remaining replicas may exist for the partitions on the dead nodes. This recovery method is used in scenarios where you are already experiencing data loss - the goal is to stop the loss of any more data.
+
+== Assumptions
+
+This capability only attempts to recover data from partitions where the following conditions are met:
+
+* Your cluster must have one or more irrecoverable nodes.
+* One or more partitions on these nodes must have no majority.
+* Some of the data on these partitions is no longer recoverable.
```
mattschumpert: @bharathv is this last part really accurate?  For some of the partitions, you will still have the leader and NO data will be unrecoverable.   let's say in the future we have leader pinning and there are no leaders at all on the nodes you're recovering from.  You may have stuck partitions due to lost majority but may end up with all the data at the end.

mattschumpert: @greg-szrama 

IMO the way to explain this feature  is that it is intended for situations when partitions have lost majority for whatever reason, including irrecoverable nodes or user error.  

We say that up front (that it could be just human error). but then we say 'only where the following conditions are met', and not all of those 3 things must be true (e.g. in the case of human error the nodes might be lost, in the case of only follower replicas on the target nodes, there's no unrecoverable data from those partitions).

So I don't think the assumptions section is really helpful here. IMO we're overcomplicating the story. 

The command reviews all partitions from those nodes, detects any partitions on them that have lost majority and therefore may have unrecoverable data.

We also don't explain that we move all replicas OFF the target nodes in the process. So we are both repairing the partitions and also draining the target nodes of all their replicas in one step.

greg-szrama: I removed the section and added a paragraph to explain your last note. Let me know your thoughts.

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+== Perform the recovery operation
+
+Node-wise partition recovery is performed via the `rpk` command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. You may lose some data which was not yet replicated to the surviving partition replicas as a result of executing this operation.
+
+The syntax for this command is as follows:
+
+ rpk cluster partitions unsafe-recover --from-nodes 1,3,5
+
+The `--from-nodes` parameter accepts a comma-delineated list of dead node IDs you wish to recover the data from. The above example would perform recovery operations on nodes 1, 3, and 5. Redpanda will assess these nodes to identify which partitions lack majority. It will then create a plan to recover the impacted partitions and ask you to confirm it. You must respond `yes` to the prompt to continue with recovery.
```
bharathv: nit: 1, 3 and 5 (extra comma).

bharathv (approved):

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,40 @@
+= Node-wise Partition Recovery
+:description: Feature to recover partitions that have lost a majority of replicas.
+
+Multi-node or entire-AZ failures (especially in cloud environments) and some forms of human error may result in ‘stuck’ partitions with fewer replicas than required to make a quorum. In these scenarios, some data loss may be unavoidable but admins need to recover what they can and move on.
```
Feediver1:
```suggestion
Multi-node or entire-AZ failures (especially in cloud environments), and some forms of human error may result in ‘stuck’ partitions with fewer replicas than required to make a quorum. In such failure scenarios, some data loss may be unavoidable. This description of node-wise partition recovery helps admins understand what they can or cannot recover.
```

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+IMPORTANT: This is a last ditch measure when all other recovery options have failed. In some cases, no remaining replicas may exist for the partitions on the dead nodes. This recovery method is used in scenarios where you are already experiencing data loss - the goal is to stop the loss of any more data.
```
Feediver1:
```suggestion
IMPORTANT: This is a last-ditch measure when all other recovery options have failed. In some cases, no remaining replicas may exist for the partitions on the dead nodes. This recovery method for scenarios where you are already experiencing data loss - the goal is to stop the loss of additional data.
```

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+Node-wise partition recovery allows you to unsafely recover at least a portion of your data using whatever replicas are remaining. All replicas are moved off the target nodes and allocated to healthy nodes. This process repairs partitions while draining the target nodes all in one step.
```
Feediver1:
```suggestion
Node-wise partition recovery allows you to unsafely recover at least a portion of your data using whatever replicas are remaining. All replicas are moved off the target nodes and allocated to healthy nodes. In one step, this process repairs partitions while draining the target nodes.
```

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+== Perform the recovery operation
+
+Node-wise partition recovery is performed via the `rpk` command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. You may lose some data which was not yet replicated to the surviving partition replicas as a result of executing this operation.
```
Feediver1:
```suggestion
You perform node-wise partition recovery using the `rpk` command `rpk cluster partitions unsafe-recover`. This command includes an interactive prompt to confirm execution of the generated recovery plan as it is a destructive operation. When you trigger node-wise partition recovery, the partitions on the node are rebuilt in a best-effort basis. As a result of executing this operation, you may lose some data that has not yet been replicated to the surviving partition replicas.
```

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
Feediver1:
```suggestion
The `--from-nodes` parameter accepts a comma-delineated list of dead node IDs you wish to recover the data from. The above example would perform recovery operations on nodes 1, 3, and 5. Redpanda will assess these nodes to identify which partitions lack majority. It will then create a plan to recover the impacted partitions and prompt you to confirm it. You must respond `yes` to continue with recovery.
```

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+The syntax for this command is as follows:
+
+ rpk cluster partitions unsafe-recover --from-nodes 1,3,5
+
+The `--from-nodes` parameter accepts a comma-delineated list of dead node IDs you wish to recover the data from. The above example would perform recovery operations on nodes 1, 3, and 5. Redpanda will assess these nodes to identify which partitions lack majority. It will then create a plan to recover the impacted partitions and ask you to confirm it. You must respond `yes` to the prompt to continue with recovery.
+
+You may also optionally add the `--dry` parameter to this command. This will perform a dry run and allow viewing the recovery plan with no risk to your cluster.
+
+Once the recovery operation is started, you may monitor the status of its execution via the `rpk cluster partitions balancer-status` command. The recovery operation can take some time to complete, especially when a lot of data is involved, and this command allows you to monitor progress in real-time.
```
Feediver1:
```suggestion
Once the recovery operation is started, you may monitor the status of its execution using the `rpk cluster partitions balancer-status` command. The recovery operation can take some time to complete, especially when a lot of data is involved. This command allows you to monitor progress in real-time.
```

Feediver1 (approved):
Note that our Style guide prohibits the use of "via"--I've replaced with "using". Revised a few places to use more active voice.

modules/manage/pages/cluster-maintenance/nodewise-partition-recovery.adoc
```diff
@@ -0,0 +1,40 @@
+= Node-wise Partition Recovery
+:description: Feature to recover partitions that have lost a majority of replicas.
+
+Multi-node or entire-AZ failures (especially in cloud environments), and some forms of human error may result in ‘stuck’ partitions with fewer replicas than required to make a quorum. In such failure scenarios, some data loss may be unavoidable. This description of node-wise partition recovery helps admins understand what they can or cannot recover.
+
+IMPORTANT: This is a last-ditch measure when all other recovery options have failed. In some cases, no remaining replicas may exist for the partitions on the dead nodes. This recovery method is intended for scenarios where you are already experiencing data loss - the goal is to stop the loss of additional data.
+
+Node-wise partition recovery allows you to unsafely recover at least a portion of your data using whatever replicas are remaining. All replicas are moved off the target nodes and allocated to healthy nodes. In one step, this process repairs partitions while draining the target nodes.
```
mattschumpert, line 8: 👌

mattschumpert (approved):

PR #272, opened 2024-01-23
Updating compaction image

greg-szrama:
Fixes: https://github.com/redpanda-data/documentation-private/issues/2170

Preview: https://deploy-preview-272--redpanda-docs-preview.netlify.app/current/manage/cluster-maintenance/compaction-settings/#redpanda-compaction-overview


Deflaimun (approved):

andrwng (approved):
Pretty much looks good to me.

One suggestion, after looking at it with fresh eyes I was confused because I thought the dotted box meant that everything in the box was deleted, when really it's referring to only the first and third record. Perhaps we should copy the black box to also be included on the third record to avoid that confusion?

PR #276, opened 2024-01-23
add example to set property keys

Deflaimun:
Preview: 
https://deploy-preview-276--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/cloud/managed-connectors/converters-and-serialization

Not sure if this is the best place to put this type of content. If not, please let me know where we can put it

### DIFF
--- a/modules/deploy/pages/deployment-option/cloud/managed-connectors/converters-and-serialization.adoc
+++ b/modules/deploy/pages/deployment-option/cloud/managed-connectors/converters-and-serialization.adoc
@@ -80,3 +80,9 @@ It works for primitive types (numbers, boolean, string) and enum types.
 To use the converter, select the `Protobuf` option as a key or value message format.
 
 See also: xref:manage:schema-registry.adoc[Redpanda Schema Registry]
+
+== Set property keys
+
+Redpanda Connectors use a set of `<property.key>=<value>` to setup its properties. 
+
+For example if you want to set the property `topic.creation.enable` to `true`, use `topic.creation.enable=true` in the property settings page.
--- a/modules/deploy/pages/deployment-option/cloud/managed-connectors/converters-and-serialization.adoc
+++ b/modules/deploy/pages/deployment-option/cloud/managed-connectors/converters-and-serialization.adoc
@@ -83,6 +83,6 @@ See also: xref:manage:schema-registry.adoc[Redpanda Schema Registry]
 
 == Set property keys
 
-Redpanda Connectors use a set of `<property.key>=<value>` to setup its properties. 
+Redpanda Connectors use a set of `<property.key>=<value>` to set up its properties. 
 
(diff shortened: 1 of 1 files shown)


modules/deploy/pages/deployment-option/cloud/managed-connectors/converters-and-serialization.adoc
```diff
@@ -80,3 +80,9 @@ It works for primitive types (numbers, boolean, string) and enum types.
 To use the converter, select the `Protobuf` option as a key or value message format.
 
 See also: xref:manage:schema-registry.adoc[Redpanda Schema Registry]
+
+== Set property keys
+
+Redpanda Connectors use a set of `<property.key>=<value>` to setup its properties. 
+
+For example if you want to set the property `topic.creation.enable` to `true`, use `topic.creation.enable=true` in the property settings page.
```
bochenekmartin, line 88: Connector properties are configurable in a JSON editor. 
Do you think `topic.creation.enable=true` intuitive enough or `"topic.creation.enable": "true"` would be better?

bochenekmartin (commented):
> Not sure if this is the best place to put this type of content. If not, please let me know where we can put it

What issue we want to address with this PR? 
tbh, _property keys_ section describes some detailed aspect of connector configuration and we have multiple paged about how to create a connector, for each connector type

modules/deploy/pages/deployment-option/cloud/managed-connectors/converters-and-serialization.adoc
Deflaimun: @bochenekmartin the idea is to provide the information on how to setup the values. Something that we were missing in the current docs. While we do talk about the properties all the time, we don't have a generic example on how to set up the keys

Deflaimun: > Do you think topic.creation.enable=true intuitive enough or "topic.creation.enable": "true" would be better?

both are ok. We just have to use whatever we recommend that users will do. Do we have to encapsulate the values  in quotes?

modules/deploy/pages/deployment-option/cloud/managed-connectors/converters-and-serialization.adoc
```diff
@@ -80,3 +80,9 @@ It works for primitive types (numbers, boolean, string) and enum types.
 To use the converter, select the `Protobuf` option as a key or value message format.
 
 See also: xref:manage:schema-registry.adoc[Redpanda Schema Registry]
+
+== Set property keys
+
+Redpanda Connectors use a set of `<property.key>=<value>` to setup its properties.
```
Feediver1:
```suggestion
Redpanda Connectors use a set of `<property.key>=<value>` to set up its properties. 
```

Feediver1 (approved):
Thanks for proactively taking this on @Deflaimun !

PR #295, opened 2024-02-06
Update log_segment_ms min default

Feediver1:
Fixes https://github.com/redpanda-data/documentation-private/issues/2249

### DIFF
--- a/modules/reference/pages/cluster-properties.adoc
+++ b/modules/reference/pages/cluster-properties.adoc
@@ -1002,7 +1002,7 @@ The topic property xref:./topic-properties.adoc#segmentms[`segment.ms`] override
 
 *Default*: null
 
-*Range*: [60000 (60 sec), ...]
+*Range*: [600000 (10 minutes), ...]
 
 *Restart required*: no
 
--- a/modules/reference/pages/cluster-properties.adoc
+++ b/modules/reference/pages/cluster-properties.adoc
@@ -994,15 +994,15 @@ If `null`, the property is disabled, and traffic is not limited.
 
 === log_segment_ms
 
-Default lifetime of log segments. If `null`, the property is disabled, and no default lifetime is set. This property can also be set in the Kafka API using the Kafka-compatible alias, `log.roll.ms`.
+Default lifetime of log segments. If `null`, the property is disabled, and no default lifetime is set. Any value under 60 seconds (60000 ms) is rejected. This property can also be set in the Kafka API using the Kafka-compatible alias, `log.roll.ms`.
 
 The topic property xref:./topic-properties.adoc#segmentms[`segment.ms`] overrides the value of `log_segment_ms` at the topic level.
 
 *Units*: milliseconds
 
-*Default*: null
+*Default*: 1209600000 (2 weeks)
 
-*Range*: [600000 (10 minutes), ...]
+*Range*: [60000 (60 sec), ...]
 
 *Restart required*: no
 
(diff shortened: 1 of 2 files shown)


modules/reference/pages/tunable-properties.adoc
```diff
@@ -1870,9 +1870,9 @@ The lower bound for the topic property `segment.ms` and the cluster property `lo
 
 *Units*: milliseconds
 
-*Default*: 60000 (60 sec)
+*Default*: 600000 (10 minutes)
 
-*Range*: [60000 (60 sec), ...]
+*Range*: [600000 (10 minutes), ...]
```
dotnwat: The properties `log_segment_ms_min` and `log_segment_ms_max` don't really have a range in the sense that these two properties define the min and max values of the valid range of `log_segment_ms`. Technically they have a range that is more like `[0, ...)`.

dotnwat (commented):
oh, just wanted to say: these changes look fine. the update to the min/max properties seems like an update that is consistent with the existing content. the comment about the range of the min/max properties is just a drive-by comment that might want to be addressed.

modules/reference/pages/cluster-properties.adoc
```diff
@@ -1002,7 +1002,7 @@ The topic property xref:./topic-properties.adoc#segmentms[`segment.ms`] override
 
 *Default*: null
 
-*Range*: [60000 (60 sec), ...]
+*Range*: [600000 (10 minutes), ...]
```
andijcr: https://github.com/redpanda-data/redpanda/blob/0afd0bd1ab7a012149b460639fe525d82299992f/src/v/config/configuration.cc#L89-L98

The default value is 2 weeks (in milliseconds: 1209600000), and the range didn't change: any value under 60 seconds (in milliseconds 60000) is rejected.

the values of this prop haven't changed since the first release it landed on (23.1 if I remember correctly, but I might be wrong)

```suggestion
*Default*: 1209600000 (2 weeks)

*Range*: [60000 (60 sec), ...]
```

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1893,7 +1893,7 @@ The upper bound for the topic property `segment.ms` and the cluster property `lo
 
 *Default*: 31536000000 (365 days)
 
-*Range*: [60000 (60 sec), ...]
+*Range*: [600000 (10 minutes), ...]
```
andijcr: as Noah said, the range [0, ...]

andijcr (approved):
lgtm

modules/reference/pages/tunable-properties.adoc
micheleRP:
```suggestion
*Range*: [600000, ...]
```

micheleRP (approved):
suggested leaving the 10 minute explanation out of the range for better readability (especially since it's in the default, immediately above this line); otherwise, lgtm!

modules/reference/pages/tunable-properties.adoc
Feediver1:
```suggestion
*Range*: [0, ...]
```

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1870,9 +1870,9 @@ The lower bound for the topic property `segment.ms` and the cluster property `lo
 
 *Units*: milliseconds
 
-*Default*: 60000 (60 sec)
+*Default*: 600000 (10 minutes)
 
-*Range*: [60000 (60 sec), ...]
+*Range*: [600000, ...]
```
Feediver1:
```suggestion
*Range*: [0, ...]
```

PR #302, opened 2024-02-12
add github folder to shared branch

Deflaimun:


JakeSCahill (commented):
I think we also need the `build.yaml` workflow from `main` in this branch. The site build doesn't currently get triggered on pushes to the `shared` branch.

JakeSCahill (approved):
LGTM

PR #32, opened 2023-09-22
docs: Fix K8s SASL docs

JakeSCahill:
An issue was discovered by a community user in which setting `users` to `null` was not allowed when using the Redpanda Operator.

- Updated `users: null` to `users: []`.
- Updated rpk flags to use `-X`.
- Added line highlighting for code blocks.
- Improved descriptions of config properties.
- Added procedure for updating superuser secrets

PR #309, opened 2024-02-21
Fix auth.sasl.users set command

JakeSCahill:
We had a fix to correct the way we set `auth.sasl.users` to an empty list in the Redpanda resource, but it appears that we still need to use `null` when using the `--set` flag: https://github.com/redpanda-data/docs/pull/32


RafalKorepta (approved):

Feediver1 (approved):

greg-szrama (approved):
Simple enough fix.

PR #322, opened 2024-02-28
Document how to troubleshoot a crash loop backoff in Kubernetes

JakeSCahill:
Fixes https://github.com/redpanda-data/documentation-private/issues/2002
Fixes https://github.com/redpanda-data/documentation-private/issues/2061

### DIFF
--- a/modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
+++ b/modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
@@ -224,9 +224,9 @@ flux resume helmrelease <cluster-name> --namespace <namespace>
 //tag::crashloopbackoff[]
 === Crash loop backoffs
 
-If a broker crashes after startup, and if it gets stuck in a crash loop, it could produce progressively more stored state that uses more disk space and takes more time for each restart to process.
+If a broker crashes after startup, or gets stuck in a crash loop, it could produce progressively more stored state that uses additional disk space and takes more time for each restart to process.
 
-To prevent infinite crash loops, the Redpanda Helm chart sets the `crash_loop_limit` node property to 5. The crash loop limit is the number of consecutive crashes that can happen within one hour of each other. After Redpanda reaches this limit, it will not start until its internal consecutive crash counter is reset to zero. In Kubernetes, the Pod running Redpanda will be in a `CrashLoopBackoff` state until its internal consecutive crash counter is reset to zero.
+To prevent infinite crash loops, the Redpanda Helm chart sets the `crash_loop_limit` node property to 5. The crash loop limit is the number of consecutive crashes that can happen within one hour of each other. After Redpanda reaches this limit, it will not start until its internal consecutive crash counter is reset to zero. In Kubernetes, the Pod running Redpanda remains in a `CrashLoopBackoff` state until its internal consecutive crash counter is reset to zero.
 
 To troubleshoot a crash loop backoff:
 
--- a/modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
+++ b/modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
@@ -221,6 +221,49 @@ flux resume helmrelease <cluster-name> --namespace <namespace>
 ```
 //end::deployment-retries-exhausted[]
 
+//tag::crashloopbackoff[]
+=== Crash loop backoffs
+
+If a broker crashes after startup, and if it gets stuck in a crash loop, it could produce progressively more stored state that uses more disk space and takes more time for each restart to process.
+
+To prevent infinite crash loops, the Redpanda Helm chart sets the `crash_loop_limit` node property  to 5, which is the number of consecutive crashes that can happen within one hour of each other. After Redpanda reaches this limit, it cannot restart until its internal consecutive crash counter is reset to zero.
+
+To troubleshoot a crash loop backoff:
+
+. Check the Redpanda logs from the most recent crashes:
++
+[,bash]
+----
+kubectl logs <pod-name> --namespace <namespace>
+----
++
+NOTE: Kubernetes retains logs only for the current and the previous instance of a container. This limitation makes it difficult to access logs from earlier crashes, which may contain vital clues about the root cause of the issue. Given these log retention limitations, setting up a centralized logging system is crucial. Systems such as Loki or Fluentd can capture and store logs from all containers, ensuring you have access to historical data.
+
+. Resolve the issue that led to the crash loop backoff.
+
+. Reset the crash counter to zero to allow Redpanda to restart. You can do any of the following to reset the counter:
++
+- Update the redpanda.yaml configuration file. You can make changes to any of the following sections in the Redpanda Helm chart to trigger an update:
+* `config.cluster`
+* `config.node`
+* `config.tunable`
+
+- Delete the `startup_log` file in the broker's data directory.
++
+[,bash]
+----
+kubectl exec <pod-name> --namespace <namespace> -- rm /var/lib/redpanda/data/startup_log
+----
+
+- Wait one hour since the last crash. The crash counter resets after one hour.
+
+To avoid future crash loop backoffs and manage the accumulation of small segments effectively:
+
+* xref:manage:kubernetes/monitoring/k-monitor-redpanda.adoc[Monitor] the size and number of segments regularly.
+* Optimize your Redpanda configuration for segment management.
+* Consider implementing xref:manage:kubernetes/storage/tiered-storage/k-tiered-storage.adoc[Tiered Storage] to manage data more efficiently.
+//end::crashloopbackoff[]
+
 //tag::deployment-pod-pending[]
 === StatefulSet never rolls out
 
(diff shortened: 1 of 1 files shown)


modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
```diff
@@ -221,6 +221,49 @@ flux resume helmrelease <cluster-name> --namespace <namespace>
...
+- Update the redpanda.yaml configuration file. You can make changes to any of the following sections in the Redpanda Helm chart to trigger an update:
+* `config.cluster`
+* `config.node`
+* `config.tunable`
+
+- Delete the `startup_log` file in the broker's data directory.
++
+[,bash]
+----
+kubectl exec <pod-name> --namespace <namespace> -- rm /var/lib/redpanda/data/startup_log
```
chrisseto, line 255: This might not be feasible. If a Pod is crash looping, you can't ``exec` into it to delete the `startup_log` file. I've _sometimes_ been able to get something like this to work by wrapping this in a loop to run until a successful exit but the time window is quite small.

Maybe we should add a warning about the low likelihood of success?

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
```diff
@@ -221,6 +221,49 @@ flux resume helmrelease <cluster-name> --namespace <namespace>
 ```
 //end::deployment-retries-exhausted[]
 
+//tag::crashloopbackoff[]
+=== Crash loop backoffs
+
+If a broker crashes after startup, and if it gets stuck in a crash loop, it could produce progressively more stored state that uses more disk space and takes more time for each restart to process.
+
+To prevent infinite crash loops, the Redpanda Helm chart sets the `crash_loop_limit` node property  to 5, which is the number of consecutive crashes that can happen within one hour of each other. After Redpanda reaches this limit, it cannot restart until its internal consecutive crash counter is reset to zero.
```
chrisseto: I actually wonder if this is an accident. K8s will perform a backoff for us and as noted below getting out of this crash loop that Redpanda is now enforcing can be quite difficult. @RafalKorepta WDYT?

It may be worth noting here that the value can be set to `null` to disable this behavior.

chrisseto (approved):
LGTM!

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
JakeSCahill: I'm not sure we want to advertise disabling this. We explicitly set the default to 5 fairly recently in this PR: https://github.com/redpanda-data/helm-charts/pull/718

chrisseto: Ooooh, interesting! Thanks for the ref. In that case we may want to add in a custom delay to the Redpanda Pods so there's a window of time to allow users to delete the file. 🤔 I'll file us an issue.

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
```diff
@@ -221,6 +221,49 @@ flux resume helmrelease <cluster-name> --namespace <namespace>
...
+To troubleshoot a crash loop backoff:
+
+. Check the Redpanda logs from the most recent crashes:
++
+[,bash]
+----
+kubectl logs <pod-name> --namespace <namespace>
+----
++
+NOTE: Kubernetes retains logs only for the current and the previous instance of a container. This limitation makes it difficult to access logs from earlier crashes, which may contain vital clues about the root cause of the issue. Given these log retention limitations, setting up a centralized logging system is crucial. Systems such as Loki or Fluentd can capture and store logs from all containers, ensuring you have access to historical data.
```
RafalKorepta: NIT:

> Systems such as Loki or Fluentd

Loki could be consider as a centralized logging system, but Fluentd is only agent that pushes logs to external system. Maybe:
* https://www.elastic.co/observability/kubernetes-monitoring
* https://www.datadoghq.com/product/log-management/
* EFK stack https://www.digitalocean.com/community/tutorials/how-to-set-up-an-elasticsearch-fluentd-and-kibana-efk-logging-stack-on-kubernetes

could be a better example.

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
RafalKorepta:
```suggestion
To prevent infinite crash loops, the Redpanda Helm chart sets the `crash_loop_limit` node property to 5. The crash loop limit is the number of consecutive crashes that can happen within one hour of each other. After Redpanda reaches this limit, it will not start until its internal consecutive crash counter is reset to zero. In Kubernetes, the Pod running Redpanda will be in a `CrashLoopBackoff` state until its internal consecutive crash counter is reset to zero.
```

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
RafalKorepta: I agree with you @chrisseto when I was testing this I couldn't get this working as `execing` into container was impossible. Redpanda bail out in a matter of milliseconds. That's why I recommend to change `some` settings of the Redpanda.

Another option would be to access Redpanda data volume to remove mentioned file, but that might be difficult to achieve if you are not the admin of the Kubernetes cluster.

RafalKorepta (approved):
No blockers. This is an improvement that is worth merging 

LGTM

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
RafalKorepta: https://github.com/redpanda-data/core-internal/issues/820 :) I remember that Core team was asked to improve on that front.

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
JakeSCahill:
```suggestion
NOTE: Kubernetes retains logs only for the current and the previous instance of a container. This limitation makes it difficult to access logs from earlier crashes, which may contain vital clues about the root cause of the issue. Given these log retention limitations, setting up a centralized logging system is crucial. Systems such as https://grafana.com/docs/loki/latest/[Loki] or https://www.datadoghq.com/product/log-management/[Datadog] can capture and store logs from all containers, ensuring you have access to historical data.
```

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
```diff
@@ -221,6 +221,49 @@ flux resume helmrelease <cluster-name> --namespace <namespace>
...
+* `config.cluster`
+* `config.node`
+* `config.tunable`
+
+- Delete the `startup_log` file in the broker's data directory.
++
+[,bash]
+----
+kubectl exec <pod-name> --namespace <namespace> -- rm /var/lib/redpanda/data/startup_log
+----
```
JakeSCahill, line 256:
```suggestion
----
+
NOTE: It might be challenging to execute this command within a Pod that is in a `CrashLoopBackoff` state due to the limited time during which the Pod is available before it restarts. Wrapping the command in a loop might work.
```

RafalKorepta (approved):

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
chrisseto: > Another option would be to access Redpanda data volume to remove mentioned file, but that might be difficult to achieve if you are not the admin of the Kubernetes cluster.

This is quite tedious to do 😓

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
```diff
@@ -221,6 +221,51 @@ flux resume helmrelease <cluster-name> --namespace <namespace>
 ```
 //end::deployment-retries-exhausted[]
 
+//tag::crashloopbackoff[]
+=== Crash loop backoffs
+
+If a broker crashes after startup, and if it gets stuck in a crash loop, it could produce progressively more stored state that uses more disk space and takes more time for each restart to process.
```
Feediver1:
```suggestion
If a broker crashes after startup, or gets stuck in a crash loop, it could produce progressively more stored state that uses additional disk space and takes more time for each restart to process.
```

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
```diff
@@ -221,6 +221,51 @@ flux resume helmrelease <cluster-name> --namespace <namespace>
...
+
+To prevent infinite crash loops, the Redpanda Helm chart sets the `crash_loop_limit` node property to 5. The crash loop limit is the number of consecutive crashes that can happen within one hour of each other. After Redpanda reaches this limit, it will not start until its internal consecutive crash counter is reset to zero. In Kubernetes, the Pod running Redpanda will be in a `CrashLoopBackoff` state until its internal consecutive crash counter is reset to zero.
```
Feediver1:
```suggestion
To prevent infinite crash loops, the Redpanda Helm chart sets the `crash_loop_limit` node property to 5. The crash loop limit is the number of consecutive crashes that can happen within one hour of each other. After Redpanda reaches this limit, it will not start until its internal consecutive crash counter is reset to zero. In Kubernetes, the Pod running Redpanda remains in a `CrashLoopBackoff` state until its internal consecutive crash counter is reset to zero.
```

Feediver1 (approved):

PR #331, opened 2024-03-05
Restart is not required for cloud_storage_segment_upload_timeout_ms

vuldin:
This may not be the only parameter that improperly says a restart is required, but I know this one does not take a restart to be applied. Maybe we could figure out a way to automate syncing this list with Redpanda core.


Feediver1 (approved):

Deflaimun:
This will be hopefully fixed with #334 for all properties

PR #335, opened 2024-03-07
connectors: jdbc-source memory warning

bochenekmartin:


modules/deploy/pages/deployment-option/cloud/managed-connectors/create-jdbc-source-connector.adoc
```diff
@@ -21,6 +21,7 @@ The JDBC Source connector has the following limitations:
 ** PostgreSQL 8.2 and higher using the version 3.0 of the PostgreSQL® protocol
 ** SQLite
 ** SQL Server - Microsoft SQL versions: Azure SQL Database, Azure Synapse Analytics, Azure SQL Managed Instance, SQL Server 2014, SQL Server 2016, SQL Server 2017, SQL Server 2019
+* This connector is excessive in memory usage. Select the initial timestamp or initial value carefully, so the data set is not too big
```
tomasz-sadura: I would avoid stating "This connector is excessive in memory usage." in general. Maybe "Select the initial timestamp or initial value carefully, so the data set is not too big. Otherwise the memory usage can be too excessive."?

tomasz-sadura (approved):

modules/deploy/pages/deployment-option/cloud/managed-connectors/create-jdbc-source-connector.adoc
```diff
@@ -21,6 +21,7 @@ The JDBC Source connector has the following limitations:
 ** PostgreSQL 8.2 and higher using the version 3.0 of the PostgreSQL® protocol
 ** SQLite
 ** SQL Server - Microsoft SQL versions: Azure SQL Database, Azure Synapse Analytics, Azure SQL Managed Instance, SQL Server 2014, SQL Server 2016, SQL Server 2017, SQL Server 2019
+* Select the initial timestamp or initial value carefully, so the data set is not too big. Otherwise the memory usage can be too excessive.
```
Feediver1: Hmmm...this does not seem like a limitation. It seems more like configuration guidance. Perhaps this can go in the table where the specific configuration setting is discussed?

Also, how about:
To avoid excessive memory usage leading to a large data set, carefully select the initial timestamp or initial value.

bochenekmartin: Thanks @Feediver1 , I pushed a change in the next commit

Feediver1 (approved):

PR #336, opened 2024-03-07
Fix bootstrap procedure

JakeSCahill:
Fixes https://github.com/redpanda-data/documentation-private/issues/2295

Tested on Ubuntu.

Also single-sources duplicated information in the prod deployment and dev deployment guides.

Preview: https://deploy-preview-336--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/self-hosted/manual/production/production-deployment/#bootstrapping

### DIFF
--- a/modules/deploy/pages/deployment-option/self-hosted/manual/production/dev-deployment.adoc
+++ b/modules/deploy/pages/deployment-option/self-hosted/manual/production/dev-deployment.adoc
@@ -41,133 +41,24 @@ Redpanda uses the following default ports:
 | internal RPC
 |===
 
-== Install Redpanda
+include::deploy:partial$linux/install-redpanda.adoc[leveloffset=+1]
 
-Install Redpanda on each system you want to be part of your cluster. There are binaries available for Fedora/RedHat or Debian systems.
+include::deploy:partial$linux/install-console.adoc[leveloffset=+1]
 
-You can also install Redpanda using an xref:./production-deployment-automation.adoc[Ansible playbook].
+include::deploy:partial$linux/bootstrap.adoc[leveloffset=+1]
 
-[tabs]
-=====
-Fedora/RedHat::
-+
---
-[,bash]
-----
-curl -1sLf 'https://dl.redpanda.com/nzc4ZYQK3WRGd9sy/redpanda/cfg/setup/bash.rpm.sh' | \
-sudo -E bash && sudo yum install redpanda -y
-----
+include::deploy:partial$linux/start-redpanda.adoc[leveloffset=+1]
 
---
-Debian/Ubuntu::
-+
---
-[,bash]
-----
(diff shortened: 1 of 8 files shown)


supahcraig:
In this section we have example private IPs given, should those be the "parameterized" <self-IP> like we have for the public IPs further down?

```
redpanda:
  kafka_api:
    - address: 192.168.1.100  # Private IP for internal traffic
      name: internal
      port: 9092
    - address: 0.0.0.0  # Listen on all interfaces for external traffic
      name: external
      port: 9093
  advertised_kafka_api:
    - address: 192.168.1.100  # Advertise internal IP to internal clients
  ```

**Comment by supahcraig** [comment]
Also this line is technically false (as bootstrapping works today), which is why we had to touch these documents in the first place.    Well...maybe not _false_, but misleading.   It puts those values into the yaml, but doesn't put any of them into the correct spots.

```
After bootstrapping Redpanda, your /etc/redpanda/redpanda.yaml configuration file will be updated to include your seed servers and listener addresses.
```

**Comment by supahcraig** [comment]
The format of the yaml given here is not how the actual output looks:

```
redpanda:
    data_directory: /var/lib/redpanda/data
    empty_seed_starts_cluster: false
    seed_servers:
        - host:
            address: <seed-server-ips>
            port: 33145
```

The actual output looks like this, with the seed servers on separate lines.  Or maybe you were just being concise with your example?

```
redpanda:
    data_directory: /var/lib/redpanda/data
    empty_seed_starts_cluster: false
    seed_servers:
        - host:
            address: <seed server #1 IP>
            port: 33145
        - host:
            address: <seed server #2 IP>
            port: 33145
        - host:
            address: <seed server #3 IP>
            port: 33145
```

JakeSCahill:
> Also this line is technically false (as bootstrapping works today), which is why we had to touch these documents in the first place. Well...maybe not _false_, but misleading. It puts those values into the yaml, but doesn't put any of them into the correct spots.
> 
> ```
> After bootstrapping Redpanda, your /etc/redpanda/redpanda.yaml configuration file will be updated to include your seed servers and listener addresses.
> ```

It does include the values passed to `--self` as the listeners though, right? It just doesn't update the advertised listeners.

JakeSCahill:
> In this section we have example private IPs given, should those be the "parameterized" like we have for the public IPs further down?
> 
> ```
> redpanda:
>   kafka_api:
>     - address: 192.168.1.100  # Private IP for internal traffic
>       name: internal
>       port: 9092
>     - address: 0.0.0.0  # Listen on all interfaces for external traffic
>       name: external
>       port: 9093
>   advertised_kafka_api:
>     - address: 192.168.1.100  # Advertise internal IP to internal clients
> ```

To keep this doc succinct, I've removed this section and linked out to our doc on configuring listeners.

r-vasquez (approved):

Feediver1 (approved):
Thanks @JakeSCahill !

PR #344, opened 2024-03-13
Serverless homepage update

Deflaimun:
Resolves https://github.com/redpanda-data/documentation-private/issues/2290

Preview:
https://deploy-preview-344--redpanda-docs-preview.netlify.app/current/home

**Should only be merged on March 18th**


micheleRP:
@Deflaimun I thought Serverless was replacing Console (instead of Tiered Storage)? This is fine with me, but I have a couple suggestions: 

- Swap the **Serverless** and **Quickstart** tiles, so Serverless is the top left tile. 
- Replace the **Redpanda Console** tile with **Redpanda Cloud** (link [here](https://docs.redpanda.com/current/deploy/deployment-option/cloud/cloud-overview/)). We have a new overview page with the Serverless PR, so that page will be much more helpful. Tile text could be: A complete streaming data platform delivered as a fully-managed service. 
- Then I'd swap the location of the **Redpanda Cloud** tile with the **Use rpk** tile, so the top 3 tiles are all easy intro-type things. Actually, for the top row, I think the order should be: Serverless, Redpanda Cloud, Quickstart

@Feediver1 is that OK with you?

micheleRP (approved):
suggest changing "on the Kubernetes platform" to "on Kubernetes", and then it looks good!

PR #377, opened 2024-03-30
Clarify when controller snapshots are enabled with older, upgraded clusters

Feediver1:
Fixes https://github.com/redpanda-data/documentation-private/issues/2189

Preview in [Partition leadership elections](https://deploy-preview-377--redpanda-docs-preview.netlify.app/current/get-started/architecture/#partition-leadership-elections)
Preview in [Upgrades](https://deploy-preview-377--redpanda-docs-preview.netlify.app/current/upgrade/rolling-upgrade/#limitations)


modules/get-started/pages/architecture.adoc
```diff
@@ -47,7 +47,9 @@ If a follower triggers an election, but the incumbent leader subsequently spring
...
+Each broker has a snapshot file stored in the controller log directory, such as `/var/lib/redpanda/data/redpanda/controller/0_0/snapshot`. The controller partition is replicated by a Raft group that includes all cluster brokers, and the controller snapshot is the Raft snapshot for this group. Snapshots are hydrated when a broker joins the cluster or restarts. 
+
+The controller snapshots feature was released in Redpanda version 23.2. When older clusters are upgraded to a version supporting controller snapshots, this feature is disabled by default. To enable controller snapshots on these clusters, contact https://support.redpanda.com/hc/en-us[Redpanda Support^]. After controller snapshots is enabled, it will persist across cluster restarts and future upgrades. New clusters created subsequently automatically enable controller snapshots.
```
ztlpn: Actually starting from 23.3 this is no longer true - the feature is enabled by default for all clusters, both new and upgraded. To summarize:
* 23.1 -> 23.2 upgrade: disabled
* new 23.2 cluster: enabled
* 23.3 cluster (both new and upgraded): enabled

Feediver1: So this guidance only applies in version 23.2. Sounds like I need to update the 23.2 version for this point.

modules/upgrade/partials/rolling-upgrades/upgrade-limitations.adoc
```diff
@@ -22,4 +22,4 @@ Upgrade the Remote Read Replica cluster first, ensuring it's on the same version
 When upgrading to Redpanda 23.2, metadata from object storage is not synchronized until all brokers in the cluster are upgraded. If you need to force a mixed-version cluster to sync read replicas, transfer partition leadership to brokers running the original version.
 
 * *Controller snapshots*:
-xref:get-started:architecture.adoc#controller-partition-and-snapshots[Controller snapshots] are disabled in upgraded clusters. To re-enable them, contact link:https://support.redpanda.com/hc/en-us[Redpanda Support^].
\ No newline at end of file
+Starting in Redpanda version 23.2, when older clusters are upgraded to a version supporting controller snapshots, this feature is enabled by default.  After controller snapshots is enabled, it persists across cluster restarts and future upgrades. New clusters created subsequently automatically enable controller snapshots.
```
ztlpn: Is this the final wording? Still a bit confusing to me.

> Starting in Redpanda version 23.2, when older clusters are upgraded to a version supporting controller snapshots, this feature is enabled by default

First, this is true for 23.3, not 23.2. Second, "a version supporting controller snapshots" _is_ 23.2 and later versions. The text hints that they are somehow different.

> New clusters created subsequently automatically enable controller snapshots.

Subsequently to what? The previous sentence talks about enabling controller snapshots so it is probably that. But you can't enable anything without a cluster which is a contradiction with this cluster being new.

modules/get-started/pages/architecture.adoc
Feediver1:
```suggestion
Each broker has a snapshot file stored in the controller log directory, such as `/var/lib/redpanda/data/redpanda/controller/0_0/snapshot`. The controller partition is replicated by a Raft group that includes all cluster brokers, and the controller snapshot is the Raft snapshot for this group. Snapshots are hydrated when a broker joins the cluster or restarts. Snapshots are enabled by default for all clusters, both new and upgraded.
```

Feediver1: Given that we do not go back and update older versions, seems like this is all a moot point. Updating per new feedback.

modules/upgrade/partials/rolling-upgrades/upgrade-limitations.adoc
Feediver1:
```suggestion
Starting in Redpanda version 23.3, controller snapshots are enabled by default, and persist across cluster restarts and future upgrades. 
```

Feediver1:
Closing this PR because it is so old/oout of date that it is quicker to re-submit than resolve 12+ merge conflicts.

PR #387, opened 2024-04-03
Fix Okta sample url in config example

voutilad:
The example url is borked. This makes it look like what someone would actually need to use.


kbatuigas (approved):
Thank you for spotting this, @voutilad ! I'll go ahead and merge.

PR #396, opened 2024-04-05
[24.1] Data transform logs stored in topic

kbatuigas:
[Deploy preview](https://deploy-preview-396--redpanda-docs-preview.netlify.app/24.1/develop/data-transforms/run-transforms/#view-data-transform-logs)

Resolves https://github.com/redpanda-data/documentation-private/issues/2231


rockwotj (approved):

mattschumpert (approved):
@kbatuigas 2 things come to mind.

1. We should also show how to access/view this log from within Redpanda Console (e.g. a screenshot showing how you go to the Message Viewer, tick 'show internal topics' and then bring up this internal _redpanda.transform_logs, similar to the way to do Connectors logs viewing.
2. This page is fine, but in the rpk transform logs reference (separate PR I think, we don't show an example of --head or --tail (in which you specify the number of log lines). Reading the reference this threw me a bit. I think the flags are all for --since / --until but this wasn't clear to me.

Thanks

JakeSCahill (approved):

PR #397, opened 2024-04-05
[24.1] Add RBAC Admin API endpoints

kbatuigas:
[Deploy preview
](https://deploy-preview-397--redpanda-docs-preview.netlify.app/api/admin-api.html#tag--Roles)

- Add new "Roles" tag
- Add new RBAC Admin API endpoints, six under "Roles" tag and one under "Users" tag

Resolves https://github.com/redpanda-data/documentation-private/issues/2287

### DIFF
--- a/modules/ROOT/attachments/admin-api.yaml
+++ b/modules/ROOT/attachments/admin-api.yaml
@@ -221,7 +221,7 @@ paths:
           required: false
           schema:
             type: string
-        - description: Applies to `user` parameter. Ignored if `user` is unset
+        - description: Applies to `principal` parameter. Ignored if `principal` is not set
           in: query
           name: principal_type
           required: false
@@ -5102,7 +5102,7 @@ tags:
       Manage partitions of a Redpanda cluster.
 
   - name: Tiered Storage
-    description: Manage tiered storage.
+    description: Manage Tiered Storage.
   - name: Transactions
   - name: Users
     description: Manage SCRAM (Salted Challenge Response Authentication Mechanism) users.
--- a/modules/ROOT/attachments/admin-api.yaml
+++ b/modules/ROOT/attachments/admin-api.yaml
@@ -206,6 +206,191 @@ paths:
         200:
           description: Raft leadership transfer success
           content: {}
+  /v1/security/roles:
+    get:
+      parameters:
+        - description: Return only roles whose names match this substring
+          in: query
+          name: filter
+          required: false
+          schema:
+            type: string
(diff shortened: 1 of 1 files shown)


modules/ROOT/attachments/admin-api.yaml
```diff
@@ -206,6 +206,191 @@ paths:
...
+          required: false
+          schema:
+            type: string
+        - description: Return only roles which are assigned to this user
+          in: query
+          name: principal
+          required: false
+          schema:
+            type: string
+        - description: Applies to `user` parameter. Ignored if `user` is unset
```
kbatuigas: @oleiman Not sure which `user` parameter is being referred to here -- is it in a different endpoint?

modules/ROOT/attachments/admin-api.yaml
```diff
@@ -4831,10 +5102,12 @@ tags:
       Manage partitions of a Redpanda cluster.
 
   - name: Tiered Storage
-    description: Manage
+    description: Manage tiered storage.
   - name: Transactions
   - name: Users
     description: Manage SCRAM (Salted Challenge Response Authentication Mechanism) users.
+  - name: Roles
+    description: Role object and role membership management.
```
kbatuigas, line 5110: @oleiman Instead of having a separate "Members" tag I opted to have just the "Roles" tag and add roles and membership endpoints under there. Let me know if that doesn't make sense.

modules/ROOT/attachments/admin-api.yaml
oleiman: typo. should be `principal` 😕

modules/ROOT/attachments/admin-api.yaml
oleiman: Ya no problem at all.

modules/ROOT/attachments/admin-api.yaml
```diff
@@ -4831,10 +5102,12 @@ tags:
       Manage partitions of a Redpanda cluster.
 
   - name: Tiered Storage
-    description: Manage
+    description: Manage tiered storage.
```
Feediver1: Given that it is an Enterprise licensed feature, this should be "Manage Tiered Storage."

Feediver1 (approved):
one minor suggestion...otherwise LGTM

oleiman (commented):
lgtm aside from s/user/principal/ in the parameter descriptions, as discussed

oleiman (approved):
lgtm!

PR #398, opened 2024-04-05
add config to api

Deflaimun:


.github/ISSUE_TEMPLATE/02_new_feature.yml
```diff
@@ -0,0 +1,47 @@
+name: 🚀 New feature
```
Feediver1, line 1: Wondering if we should still use this...I have not seen anyone using it, and new feature tickets auto-generated by Aha do not seem to use it either.

Deflaimun: if we do remove it, it must be on the `main` branch

JakeSCahill (approved):

PR #399, opened 2024-04-08
Explain why we use Docker Compose in the quickstart

JakeSCahill:
Fixes https://github.com/redpanda-data/documentation-private/issues/2141

Also provides the command for setting replication factor to three for the three-broker deployment.


micheleRP (approved):
lgtm

PR #401, opened 2024-04-08
add page aliases to fully deprecated commands

Deflaimun:
Adds aliases to the deprecated page for fully deprecated rpk commands


JakeSCahill (approved):
Great, thanks 💯

PR #420, opened 2024-04-16
Updated throughput throttling documentation to reflect revised functionality

greg-szrama:
Fixes: https://github.com/redpanda-data/documentation-private/issues/2255

[Throughput documentation](https://deploy-preview-420--redpanda-docs-preview.netlify.app/24.1/manage/cluster-maintenance/manage-throughput/)
[New tunables](https://deploy-preview-420--redpanda-docs-preview.netlify.app/24.1/reference/tunable-properties/#kafka_throughput_throttling_v2)
[New internal metrics](https://deploy-preview-420--redpanda-docs-preview.netlify.app/24.1/reference/internal-metrics-reference/#vectorized_kafka_quotas_balancer_runs)


modules/reference/pages/tunable-properties.adoc
```diff
@@ -1372,6 +1372,38 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
 
 ---
 
+[[kafka_throughput_throttling_v2]]
+=== kafka_throughput_throttling_v2
+
+This disables version two of throughput throttling, introduced with Redpanda v23.3.8. When this is disabled, Redpanda uses the throttling mechanism from v23 and older. This defaults to true for all new or upgraded Redpanda clusters.
```
BenPope:
```suggestion
This enables version two of throughput throttling, introduced with Redpanda v23.3.8. Set to false to use the throttling mechanism from v23.3.7 and older. This defaults to true for all new or upgraded Redpanda clusters.
```
Or, from https://github.com/redpanda-data/redpanda/blob/v23.3.8/src/v/config/configuration.cc#L2792-L2793
```suggestion
Use throughput throttling based on a shared token bucket instead of balancing quota between shards.
```

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1372,6 +1372,38 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+
+*Type*: unsigned integer
+
+*Units*:
+
+*Default*: 1
+
+*Range*:
+
+*Restart required*:
```
BenPope: https://github.com/redpanda-data/redpanda/blob/v23.3.8/src/v/config/configuration.cc#L2801
```suggestion
*Restart required*: false
```

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1372,6 +1372,38 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+WARNING: Disabling this property causes your Redpanda cluster to use an outdated throughput throttling mechanism. Only set this to `false` when advised to do so by Redpanda support.
+
+---
+
+[[kafka_throughput_replenish_threshold]]
+=== kafka_throughput_replenish_threshold
+
+Threshold for refilling the token bucket. Will be clamped between 1 and rate.
+
+*Type*: unsigned integer
```
BenPope: It's a signed 64bit integer.
https://github.com/redpanda-data/redpanda/blob/v23.3.8/src/v/config/configuration.h#L523C36-L524

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1372,6 +1372,38 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+
+*Units*:
+
+*Default*: 1
+
+*Range*:
```
BenPope: For ingress: [1, kafka_throughput_limit_node_in_bps]
For egress: [1, kafka_throughput_limit_node_out_bps]

It's clamped automatically and silently.

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1372,6 +1372,38 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+---
+
+[[kafka_throughput_replenish_threshold]]
+=== kafka_throughput_replenish_threshold
+
+Threshold for refilling the token bucket. Will be clamped between 1 and rate.
+
+*Type*: unsigned integer
+
+*Units*:
```
BenPope: It doesn't have a unit.

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1372,6 +1372,38 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+*Restart required*: true
+
+WARNING: Disabling this property causes your Redpanda cluster to use an outdated throughput throttling mechanism. Only set this to `false` when advised to do so by Redpanda support.
+
+---
+
+[[kafka_throughput_replenish_threshold]]
+=== kafka_throughput_replenish_threshold
+
+Threshold for refilling the token bucket. Will be clamped between 1 and rate.
```
BenPope: https://github.com/redpanda-data/redpanda/blob/v23.3.8/src/v/config/configuration.cc#L2799C8-L2800C42
```suggestion
Threshold for refilling the token bucket. Will be clamped between 1 and kafka_throughput_limit_node_*_bps.

File: modules/manage/pages/cluster-maintenance/manage-throughput.adoc [hunk 7]
```diff
@@ -12,7 +12,7 @@ To manage the volume of traffic going through a broker, Redpanda implements thro
 
 With Redpanda's xref:get-started:architecture.adoc#thread-per-core-model[thread-per-core model], the Kafka API traffic to and from a client connection is processed by a single core (shard). In order to manage throughput quotas efficiently, broker quotas are distributed between shards, and each per-shard quota is in turn shared by all connections served by the shard. Splitting broker quota optimally between shards is done behind the scenes by the quota balancer component.
 
-To distribute the broker throughput quota, the balancer periodically monitors the throughput rate of a broker's shards, and it distributes more quota to the shards that can make better use of it than the others. Each shard has a minimum throughput quota value, which is configurable both as a percentage of the default quota and as an absolute rate limit.
+To distribute the broker throughput quota, the balancer periodically monitors the throughput rate of a broker's shards, and it distributes more quota to the shards that can make better use of it than the others.
```
BenPope: The quota balancer is not in use when v2 is enabled; all shards share a single token bucket (well, one for ingress and one for egress). I'd be tempted to drop both of these paragraphs.

Feediver1:
@greg-szrama Not sure why this is still in Draft state? Please update. Thx

modules/reference/pages/tunable-properties.adoc
greg-szrama: I used a little bit of both pieces of language.

greg-szrama:
I've implemented guidance from @BenPope and taken this out of draft mode. Please review.

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1372,6 +1372,36 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
 
 ---
 
+[[kafka_throughput_throttling_v2]]
+=== kafka_throughput_throttling_v2
+
+This enables an updated throttling algorithm based on a shared token bucket, introduced with Redpanda v23.3.8. Set this property to false if you need to use the quota balancing algorithm from Redpanda v23.3.7 and older. This defaults to true for all new or upgraded Redpanda clusters.
```
Feediver1:
```suggestion
Enables an updated throttling algorithm based on a shared token bucket, introduced with Redpanda v23.3.8. Set this property to false if you need to use the quota balancing algorithm from Redpanda v23.3.7 and older.  This property defaults to true for all new or upgraded Redpanda clusters.
```

Feediver1: @mattschumpert How is it that we added a new property in a patch release (only bug fixes--no new functionality)? I thought this was not allowed given our [release process](https://redpandadata.atlassian.net/wiki/spaces/product/pages/106397701/Redpanda+Releases+Versioning+Support)?

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -10,9 +10,9 @@ The network bandwidth and disk utilization of brokers may be overloaded by clien
...
 
-With Redpanda's xref:get-started:architecture.adoc#thread-per-core-model[thread-per-core model], the Kafka API traffic to and from a client connection is processed by a single core (shard). In order to manage throughput quotas efficiently, broker quotas are distributed between shards, and each per-shard quota is in turn shared by all connections served by the shard. Splitting broker quota optimally between shards is done behind the scenes by the quota balancer component.
+The throttling mechanism uses shared token buckets, one for ingress and one for egress. Tokens are added to the buckets at a constant rate and the number of tokens added is based on the amount of time elapsed since the last request for tokens. If the number of tokens in the bucket is below the `kafka_throughput_replenish_threshold`, then the bucket is topped up.
```
Feediver1: @greg-szrama Something wrong here--none of this content appears in the HTML preview for this topic. Here is what I see in the [Preview](https://deploy-preview-420--redpanda-docs-preview.netlify.app/24.1/manage/cluster-maintenance/manage-throughput/):

**Manage Throughput**
Manage the throughput of Kafka traffic at the cluster level, with configurable properties that limit and protect the use of disk and network resources for individual brokers and for an entire cluster. Set broker-wide throughput limits for Kafka API traffic.

modules/reference/pages/tunable-properties.adoc
Feediver1: @mattschumpert @BenPope Should this update discussing the versions be included?

mattschumpert: yes its good as is

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,17 @@
 :description: Manage the throughput of Kafka traffic with configurable properties.
 :page-categories: Management, Networking
 
-Manage the throughput of Kafka traffic at the cluster level, with configurable properties that limit and protect the use of disk and network resources for individual brokers and for an entire cluster. Set broker-wide throughput limits for Kafka API traffic.
+Redpanda allows cluster administrators to configure throughput limits at both the broker level and the client level. These configurable limits help protect the use of disk and network resources. Limits applied on the broker apply to all clients that produce or consume there. Alternatively, the client limits allow you to selectively throttle a single client or group of clients.
```
mattschumpert: "protect the use of disk and network resources, and also protect individual clients from 'noisy neighbors' they may consume extreme amounts of available throughput that crowd out other clients'

(referring to the per client limits @greg-szrama )

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,17 @@
...
 
-With Redpanda's xref:get-started:architecture.adoc#thread-per-core-model[thread-per-core model], the Kafka API traffic to and from a client connection is processed by a single core (shard). In order to manage throughput quotas efficiently, broker quotas are distributed between shards, and each per-shard quota is in turn shared by all connections served by the shard. Splitting broker quota optimally between shards is done behind the scenes by the quota balancer component.
+The throttling mechanism uses shared token buckets, one for ingress and one for egress. Tokens are added to the buckets at a constant rate and the number of tokens added is based on the amount of time elapsed since the last request for tokens. If the number of tokens in the bucket is below the `kafka_throughput_replenish_threshold`, then the bucket is topped up.
```
mattschumpert: @gene-redpanda  this is too technical of an explanation. The user doesn't think in terms of token buckets or toping up and they have no control over this implementation detail. They just set overall throughput quotas or per client quotas.

We can just say that the per broker quotas for ingress and egress are shared across all cpu cores (shards)

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,17 @@
...
 
-To distribute the broker throughput quota, the balancer periodically monitors the throughput rate of a broker's shards, and it distributes more quota to the shards that can make better use of it than the others. Each shard has a minimum throughput quota value, which is configurable both as a percentage of the default quota and as an absolute rate limit.
+Each request removes tokens from the corresponding ingress or egress bucket. If the bucket goes negative, Redpanda adds a throttling delay to the response which represents the amount of time needed for the bucket to return to a positive number of tokens. The delay will not exceed `max_kafka_throttle_delay_ms`. The shared bucket is topped up based on the amount of time since the last request for tokens.
```
mattschumpert: @greg-szrama same comment here.  too technical.

we can say that the throughput throttling self-adjusts frequently according to the frequency set in max_kafka_throttle_delay_ms

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -25,27 +25,17 @@ The properties for broker-wide throughput quota balancing are configured at the
...
-| xref:reference:cluster-properties.adoc#kafka_quota_balancer_min_shard_throughput_bps[kafka_quota_balancer_min_shard_throughput_bps]
-| The lowest value of the throughput quota a shard can get in the process of quota balancing, in bytes per second. If set as `0`, there is no minimum.
-
 | xref:reference:tunable-properties.adoc#max_kafka_throttle_delay_ms[max_kafka_throttle_delay_ms]
 | The maximum delay inserted in the data path of Kafka API requests to throttle them down. Configuring this to be less than the Kafka client timeout can ensure that the delay that's inserted won't be long enough to cause a client timeout by itself.
 
-| xref:reference:cluster-properties.adoc#kafka_quota_balancer_window_ms[kafka_quota_balancer_window_ms]
-| The time window the balancer uses to average the current throughput measurement.
+| xref:reference:tunable-properties.adoc#kafka_throughput_replenish_threshold[kafka_throughput_replenish_threshold]
+| The threshold for refilling a token bucket. This is automatically clamped to the corresponding throughput limit for ingress or egress.
```
mattschumpert: @greg-szrama lets explain this in a way the user can best understand it.  I myself am unclear what this means

modules/reference/pages/cluster-properties.adoc
```diff
@@ -1008,6 +911,7 @@ Size of the Kafka server TCP transmit buffer. If `null`, the property is disable
 
 ---
 
+[[kafka_throughput_limit_node_in_bps]]
```
mattschumpert, line 971: in what units (we should say below bytes per second)

modules/reference/pages/cluster-properties.adoc
```diff
@@ -1030,6 +934,7 @@ If `null`, the property is disabled, and traffic is not limited.
 
 ---
 
+[[kafka_throughput_limit_node_out_bps]]
```
mattschumpert, line 994: in what units (we should say below bytes per second for added clarity)

modules/reference/pages/internal-metrics-reference.adoc
```diff
@@ -65,6 +65,46 @@ Can indicate latency caused by disk operations.
...
+
+Number of times the throughput quota balancer has executed.
+
+*Type*: counter
+
+---
+
+=== vectorized_kafka_quotas_quota_effective
+
+Current effective quota, in bytes per second.
```
mattschumpert: what quota @greg-szrama  the node throughput quota?

modules/reference/pages/internal-metrics-reference.adoc
```diff
@@ -65,6 +65,46 @@ Can indicate latency caused by disk operations.
...
+
+=== vectorized_kafka_quotas_traffic_intake
+
+Total amount of Kafka traffic (in bytes) taken in from clients for processing.
+
+*Type*: counter
+
+---
+
+=== vectorized_kafka_quotas_traffic_egress
```
mattschumpert, line 106: @BenPope how are these different from the normal overall kafka throughput metrics?  If they are the doc description doesn't make it clear how.

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1372,6 +1372,36 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
 
 ---
 
+[[kafka_throughput_throttling_v2]]
+=== kafka_throughput_throttling_v2
+
+Enables an updated throttling algorithm based on a shared token bucket, introduced with Redpanda v23.3.8. Set this property to false if you need to use the quota balancing algorithm from Redpanda v23.3.7 and older.  This property defaults to true for all new or upgraded Redpanda clusters.
```
mattschumpert: "updated algorithm for enforcing node throughput limits..." 

(assuming this is for the node-wise throughput limits)

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1372,6 +1372,36 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+*Restart required*: true
+
+WARNING: Disabling this property causes your Redpanda cluster to use an outdated throughput throttling mechanism. Only set this to `false` when advised to do so by Redpanda support.
+
+---
+
+[[kafka_throughput_replenish_threshold]]
+=== kafka_throughput_replenish_threshold
+
+Threshold for refilling the token bucket. The range for this threshold is automatically clamped to the corresponding throughput limit for ingress and egress.
```
mattschumpert: @BenPope  does this mean: 'increments of throughput in which throughput throttling is re-evaluated' trying to translate to user-speak here.

mattschumpert (changes requested):

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
BenPope: Lets remove it from the main docs page; it's a tunable.

The bucket works on atomics. An atomic write is relatively expensive. By setting the threshold above 1, the number of writes may be reduced. This threshold is the minimum amount of refill that is required in order to actually top up the tokens.

The value provided is clamped between 1 and bucket capacity. The capacity comes from `kafka_throughput_limit_node_in_bps` or `kafka_throughput_limit_node_out_bps`.

modules/reference/pages/cluster-properties.adoc
BenPope: > in what units (we should say below bytes per second for added clarity)

Yes, bytes per second

modules/reference/pages/internal-metrics-reference.adoc
BenPope: > what quota @greg-szrama the node throughput quota?

This is primarily for the quota balancer, but it's left intact. Without the quota balancer, it's node throughput / shard count. It will be aggregated by shard if `aggregate_metrics` is true.

modules/reference/pages/internal-metrics-reference.adoc
BenPope: > @BenPope how are these different from the normal overall kafka throughput metrics? If they are the doc description doesn't make it clear how.

This is the amount of traffic that was considered by the throttler. So it depends on `kafka_throughput_controlled_api_keys`.

modules/reference/pages/tunable-properties.adoc
BenPope: > "updated algorithm for enforcing node throughput limits..."
> 
> (assuming this is for the node-wise throughput limits)

Yes, this is for controlling node-wide limit.

modules/reference/pages/tunable-properties.adoc
BenPope: > @BenPope does this mean: 'increments of throughput in which throughput throttling is re-evaluated' trying to translate to user-speak here.

It's reevaluated every request, but the number of tokens wouldn't be updated if the amount to replenish is less than the threshold. This allows avoiding the atomic write of the token count on every request.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
BenPope: > @greg-szrama same comment here. too technical.
> 
> we can say that the throughput throttling self-adjusts frequently according to the frequency set in max_kafka_throttle_delay_ms

I think it's ok to mention a token bucket, that throttling is performed via `throttle_time_ms` (which is a field in the response, and is clamped to `max_kafka_throttle_delay_ms`. If the client fails to backoff, reading its next request is delayed in order to apply backpressure.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
BenPope: > If the number of tokens in the bucket is below the `kafka_throughput_replenish_threshold`, then the bucket is topped up.

If the number of tokens to be replenished is above `kafka_throughput_replenish_threshold` then the tokens are replenished.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: I like that, I added a blurb to the end of the paragraph to help talk about client limits.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
Feediver1:
```suggestion
The throttling mechanism uses shared token buckets, one for ingress and one for egress. Tokens are added to the buckets at a constant rate and the number of tokens added is based on the amount of time elapsed since the last request for tokens. If the number of tokens to be replenished is above `kafka_throughput_replenish_threshold` then the tokens are replenished.
```

BenPope: There's some duplication in saying "rate" and then tokens added per time. This whole section is too detailed.

greg-szrama: I reworked the language in this section to try to share a more applicable level of detail.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: I reworked the language in this section to try to share a more applicable level of detail.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: Removed the property from here, but kept a link to it in the writeup before the table.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: I've reworked this section so this language is no longer present.

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1373,6 +1373,42 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
 
 ---
 
+[[kafka_throughput_throttling_v2]]
+=== kafka_throughput_throttling_v2
+
+Enables an updated algorithm for enforcing node throughput limits based on a shared token bucket, introduced with Redpanda v23.3.8. Set this property to false if you need to use the quota balancing algorithm from Redpanda v23.3.7 and older.  This property defaults to true for all new or upgraded Redpanda clusters.
```
Feediver1:
```suggestion
Enables an updated algorithm for enforcing node throughput limits based on a shared token bucket, introduced with Redpanda v23.3.8. Set this property to `false` if you need to use the quota balancing algorithm from Redpanda v23.3.7 and older.  This property defaults to `true` for all new or upgraded Redpanda clusters.
```

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1373,6 +1373,42 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+WARNING: Disabling this property causes your Redpanda cluster to use an outdated throughput throttling mechanism. Only set this to `false` when advised to do so by Redpanda support.
+
+---
+
+[[kafka_throughput_replenish_threshold]]
+=== kafka_throughput_replenish_threshold
+
+Threshold for refilling the token bucket as part of enforcing throughput limits. This only applies when xref:kafka_throughput_throttling_v2[] is `true`.
+
+This threshold is evaluated with each request for data. When the amount of tokens to replenish exceeds this threshold, then we add tokens to the token bucket. This ensures we're not updating the atomic for the token count with each request. The range for this threshold is automatically clamped to the corresponding throughput limit for ingress and egress.
```
Feediver1:
```suggestion
This threshold is evaluated with each request for data. When the number of tokens to replenish exceeds this threshold, then tokens are added to the token bucket. This ensures that the atomic is not being updated for the token count with each request. The range for this threshold is automatically clamped to the corresponding throughput limit for ingress and egress.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,17 @@
...
+== Advanced throughput limit tuning
 
-To distribute the broker throughput quota, the balancer periodically monitors the throughput rate of a broker's shards, and it distributes more quota to the shards that can make better use of it than the others. Each shard has a minimum throughput quota value, which is configurable both as a percentage of the default quota and as an absolute rate limit.
+The throttling mechanism separates ingress and egress quotas and applies them across all CPU cores. When one of the throttling quotas is reached, Redpanda adds a throttling delay, `throttle_time_ms`, to subsequent responses. If the client still fails to back off, then the next read operation is delayed. Redpanda attempts to set the throttling delay to an amount that will put the cluster back below the throttling threshold. This delay will not exceed the value set in `max_kafka_throttle_delay_ms`. Administrators may also use the tunable property  xref:reference:tunable-properties.adoc#kafka_throughput_replenish_threshold[`kafka_throughput_replenish_threshold`] to adjust the rate that Redpanda eases throttling once it is engaged.
```
Feediver1: If you can actually count something (like clusters, or seconds, etc), use _number_. If it is something hard to quantify, like the number of grains in a cup of sugar, use _amount_. Seems like you are using amount here when you should be using number...

greg-szrama: Switchied to 'number'.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,17 @@
...
 
-Manage the throughput of Kafka traffic at the cluster level, with configurable properties that limit and protect the use of disk and network resources for individual brokers and for an entire cluster. Set broker-wide throughput limits for Kafka API traffic.
+Redpanda allows cluster administrators to configure throughput limits at both the broker level and the client level. For brokers, these configurable limits help protect the use of disk and network resources. Limits applied on the broker apply to all clients that produce or consume there. Alternatively, the client limits allow you to selectively throttle a single client or group of clients. These limits help protect individual clients from 'noisy neighbors' that may consume extreme amounts of available throughput and crowd out other clients.
 
 == Broker-wide throughput limits
 
 The network bandwidth and disk utilization of brokers may be overloaded by clients that produce or consume throughput without limits. To prevent resource overloading caused by unconstrained throughput and to configure back pressure, Redpanda provides runtime-configurable properties that limit and balance throughput of Kafka API traffic.
```
mattschumpert: 'network & storage bandwidth of  brokers may be overloaded by. '

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,17 @@
...
 
 To manage the volume of traffic going through a broker, Redpanda implements throughput quotas on the ingress and egress sides of every broker. The throughput quota accounts for all Kafka API traffic going in or out of a broker, with the value of quota representing the allowed rate of data passing through in one direction. When a connection is in breach of the quota, the throttler advises the client about the delay (throttle time) that would bring the rate back to the allowed level, and it implements that delay before handling Kafka API requests. To control the quotas, Redpanda provides configurable rate limits for total ingress and egress traffic through a broker.
 
-With Redpanda's xref:get-started:architecture.adoc#thread-per-core-model[thread-per-core model], the Kafka API traffic to and from a client connection is processed by a single core (shard). In order to manage throughput quotas efficiently, broker quotas are distributed between shards, and each per-shard quota is in turn shared by all connections served by the shard. Splitting broker quota optimally between shards is done behind the scenes by the quota balancer component.
+== Advanced throughput limit tuning
```
mattschumpert: Much better, but I think the advanced section should go below the properties tables.  the properties are the basic stuff, this stuff should be at the bottom.  And I assume we should say 'under normal circumstances these should not need tuning'.  True @BenPope ?

mattschumpert (approved):

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: I think reordering this is perfectly fair, especially since we're not referencing a lot of these advanced properties in that table. I was trying to make sure it didn't get confused with the topic-level throughput features but I think it'll be clear.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: Modified.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -25,29 +23,20 @@ The properties for broker-wide throughput quota balancing are configured at the
...
-* `kafka_quota_balancer_min_shard_throughput_bps` doesn't override the `kafka_throughput_limit_node_in_bps` and `kafka_throughput_limit_node_out_bps` limit settings. Consequently, the value of
-`kafka_throughput_limit_node_in_bps` or `kafka_throughput_limit_node_out_bps` can result in lesser throughput than `kafka_quota_balancer_min_shard_throughput_bps`.
 ====
 
+=== Advanced broker-wide throughput limit tuning
+
+The throttling mechanism separates ingress and egress quotas and applies them across all CPU cores. When one of the throttling quotas is reached, Redpanda adds a throttling delay, `throttle_time_ms`, to subsequent responses. If the client still fails to back off, then the next read operation is delayed. Redpanda attempts to set the throttling delay to a number that will put the cluster back below the throttling threshold. This delay will not exceed the value set in `max_kafka_throttle_delay_ms`. Administrators may also use the tunable property  xref:reference:tunable-properties.adoc#kafka_throughput_replenish_threshold[`kafka_throughput_replenish_threshold`] to adjust the rate that Redpanda eases throttling once it is engaged.
```
Deflaimun: What are those values? cluster properties? if yes, please link to them 

ie
- throttle_time_ms
- max_kafka_throttle_delay_ms

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1373,6 +1373,42 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+
+Enables an updated algorithm for enforcing node throughput limits based on a shared token bucket, introduced with Redpanda v23.3.8. Set this property to `false` if you need to use the quota balancing algorithm from Redpanda v23.3.7 and older.  This property defaults to `true` for all new or upgraded Redpanda clusters.
+
+*Type*: boolean
+
+*Default*: true
+
+*Restart required*: true
+
+WARNING: Disabling this property causes your Redpanda cluster to use an outdated throughput throttling mechanism. Only set this to `false` when advised to do so by Redpanda support.
```
Deflaimun:
```suggestion
WARNING: Disabling this property is not recommended. It causes your Redpanda cluster to use an outdated throughput throttling mechanism. Only set this to `false` when advised to do so by Redpanda support.
```

Deflaimun (approved):
Check comments. LGTM

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: There's not one for `throttle_time_ms` since it's a property in the response object. I did add a link for `max_kafka_throttle_delay_ms`.

modules/reference/pages/tunable-properties.adoc
```diff
@@ -1373,6 +1373,42 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+[[kafka_throughput_throttling_v2]]
+=== kafka_throughput_throttling_v2
+
+Enables an updated algorithm for enforcing node throughput limits based on a shared token bucket, introduced with Redpanda v23.3.8. Set this property to `false` if you need to use the quota balancing algorithm from Redpanda v23.3.7 and older.  This property defaults to `true` for all new or upgraded Redpanda clusters.
+
+*Type*: boolean
+
+*Default*: true
+
+*Restart required*: true
```
BenPope: https://github.com/redpanda-data/redpanda/blob/v24.1.1-rc8/src/v/config/configuration.cc#L2899
```suggestion
*Restart required*: false
```

modules/reference/pages/internal-metrics-reference.adoc
```diff
@@ -71,6 +71,46 @@ Can indicate latency caused by disk operations.
...
+
+Current effective quota for the quota balancer, in bytes per second.
+
+*Type*: counter
+
+---
+
+=== vectorized_kafka_quotas_throttle_time
+
+Sequence of throttle times, in seconds.
```
BenPope: It's a histogram, not a sequence.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,15 @@
 :description: Manage the throughput of Kafka traffic with configurable properties.
 :page-categories: Management, Networking
 
-Manage the throughput of Kafka traffic at the cluster level, with configurable properties that limit and protect the use of disk and network resources for individual brokers and for an entire cluster. Set broker-wide throughput limits for Kafka API traffic.
+Redpanda allows cluster administrators to configure throughput limits at both the broker level and the client level. For brokers, these configurable limits help protect the use of disk and network resources. Limits applied on the broker apply to all clients that produce or consume there. Alternatively, the client limits allow you to selectively throttle a single client or group of clients. These limits help protect individual clients from 'noisy neighbors' that may consume extreme amounts of available throughput and crowd out other clients.
```
BenPope: At a high level the broker and client throttling work similarly, so I added high-level context at the top.

I tried to make the wording less awkward, too, but I may not have achieved that.
```suggestion
Throughput throttling can be configured for ingress and egress at the broker or client level to prevent unbounded network and disk usage of the broker by clients. Broker-wide limits apply to all clients connected to the broker to restrict total traffic on the broker. Client limits apply to the set of clients defined by their `client_id`, and are useful to prevent a set of clients from starving other clients that are using the broker.

Throughput limits are enforced by applying backpressure to clients, either via the `throttle_time_ms` field in the response, or, if that isn't honoured, directly inserting a delay on the next read of the connection. The maximum delay that shall be requested is governed by xref:reference:tunable-properties.adoc#max_kafka_throttle_delay_ms[max_kafka_throttle_delay_ms].
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,15 @@
...
+The network and storage bandwidth of brokers may be overloaded by clients that produce or consume throughput without limits. To prevent resource overloading caused by unconstrained throughput and to configure back pressure, Redpanda provides runtime-configurable properties that limit and balance throughput of Kafka API traffic.
 
 To manage the volume of traffic going through a broker, Redpanda implements throughput quotas on the ingress and egress sides of every broker. The throughput quota accounts for all Kafka API traffic going in or out of a broker, with the value of quota representing the allowed rate of data passing through in one direction. When a connection is in breach of the quota, the throttler advises the client about the delay (throttle time) that would bring the rate back to the allowed level, and it implements that delay before handling Kafka API requests. To control the quotas, Redpanda provides configurable rate limits for total ingress and egress traffic through a broker.
```
BenPope:
```suggestion
Redpanda provides configurable rate limits for total ingress and egress traffic through a broker.

Broker-wide throughput limits apply to all clients that haven't been excluded with `kafka_throughput_control`.

Broker-wide throughput limits apply to Kafka request types that exist in `kafka_throughput_controlled_api_keys`, which defaults to `produce` and `fetch`.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -25,29 +23,20 @@ The properties for broker-wide throughput quota balancing are configured at the
...
 | xref:reference:tunable-properties.adoc#max_kafka_throttle_delay_ms[max_kafka_throttle_delay_ms]
 | The maximum delay inserted in the data path of Kafka API requests to throttle them down. Configuring this to be less than the Kafka client timeout can ensure that the delay that's inserted won't be long enough to cause a client timeout by itself.
 
-| xref:reference:cluster-properties.adoc#kafka_quota_balancer_window_ms[kafka_quota_balancer_window_ms]
-| The time window the balancer uses to average the current throughput measurement.
 |===
 
 [NOTE]
 ====
 * By default, both `kafka_throughput_limit_node_in_bps` and `kafka_throughput_limit_node_out_bps` are disabled, no throughput limits are applied. You must manually set them to enable quota balancing with throughput limits.
```
BenPope:
```suggestion
* By default, both `kafka_throughput_limit_node_in_bps` and `kafka_throughput_limit_node_out_bps` are disabled, no throughput limits are applied. You must manually set them to enable throughput throttling.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -25,29 +23,20 @@ The properties for broker-wide throughput quota balancing are configured at the
 | xref:reference:cluster-properties.adoc#kafka_throughput_limit_node_out_bps[kafka_throughput_limit_node_out_bps]
 | A broker's total throughput limit for egress Kafka traffic.
```
BenPope, line 27: Should document / reference `kafka_throughput_controlled_api_keys` and `kafka_throughput_control` (both cluster properties)

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -25,29 +23,20 @@ The properties for broker-wide throughput quota balancing are configured at the
...
-`kafka_throughput_limit_node_in_bps` or `kafka_throughput_limit_node_out_bps` can result in lesser throughput than `kafka_quota_balancer_min_shard_throughput_bps`.
 ====
 
+=== Advanced broker-wide throughput limit tuning
+
+The throttling mechanism separates ingress and egress quotas and applies them across all CPU cores. When one of the throttling quotas is reached, Redpanda adds a throttling delay, `throttle_time_ms`, to subsequent responses. If the client still fails to back off, then the next read operation is delayed. Redpanda attempts to set the throttling delay to a number that will put the cluster back below the throttling threshold. This delay will not exceed the value set in xref:reference:tunable-properties.adoc#max_kafka_throttle_delay_ms[`max_kafka_throttle_delay_ms`]. Administrators may also use the tunable property  xref:reference:tunable-properties.adoc#kafka_throughput_replenish_threshold[`kafka_throughput_replenish_threshold`] to adjust the rate that Redpanda eases throttling once it is engaged.
```
BenPope:
```suggestion
```

modules/reference/pages/internal-metrics-reference.adoc
Deflaimun:
```suggestion
Histogram of throttle times, in seconds.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
Feediver1: @mattschumpert Are you in agreement with this update request? @BenPope is likely offline, and we would like to get this as far along today as possible. Please advise.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
Feediver1: Same here @mattschumpert Are you good with the being dleted? @BenPope If online, please confirm--did you intend for this description to be removed from the docs?

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
mattschumpert: For this one I actually like having both the description and the properties just below present

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
mattschumpert: This looks fine to me

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
mattschumpert: @Feediver1 I rather suggested this part get moved to the bottom of the page since its highly advanced stuff most people won't touch.  Not worthless, might be fine to keep it, but only for advanced users or when recommended by Redpanda Support

BenPope: > @BenPope If online, please confirm--did you intend for this description to be removed from the docs?

Yes. I suggested a diluted version of this to the top, where it also applies to client throttling.

Previously I suggested removing the tunable, since it's a tunable, and in benchmarking I couldn't find a measurable difference in performance.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
BenPope: I found this section to be fairly repetitive and wordy within itself, and also duplicates much of my suggested rewording at the top scope.

greg-szrama: On further review, I'm thinking this comment was meant to go along with the below comment:

@[BenPope](https://github.com/BenPope) BenPope [9 hours ago](https://github.com/redpanda-data/docs/pull/420#discussion_r1580732412)
Should document / reference kafka_throughput_controlled_api_keys and kafka_throughput_control (both cluster properties)

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: I'll go ahead and remvoe this section then.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: I'm going to work on this a bit, @BenPope , I want to make sure the top portion that discusses both broker-wide and client limits doesn't confuse the bit here on just broker concerns. I'll worsmith it to make sure it's not repetitive information and adds value specifically on the broker-wide configuration question.

BenPope: The backpressure mechanism is essentially the same for broker and client throttling; a token bucket, throttle_time_ms, forced delay by not reading the next request.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: @BenPope - Are these new properties? We don't have them in the cluster properties list. Can you help flesh out details for them?

Specifically, if appropriate:
Type
Units
Default
Range
Restart required

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: I've wordsmithed this a bit to remove redundancy and try to improve the flow for the user reading the documentation.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
BenPope: Not new.

### `kafka_throughput_controlled_api_keys`
Source: https://github.com/redpanda-data/redpanda/blob/v23.3.13/src/v/config/configuration.cc#L2852-L2858

Type: `list<string>`
Units: N/A
Default: `["produce", "fetch"]`
Range: N/A
Restart required: No
Visibility: User
Example values (possibly incomplete now): https://github.com/redpanda-data/redpanda/pull/10285/files#diff-a31fcb7d2066e40789fe6b678cfd993c94e3bebc66ddf852934f737d63379057R70-R83

### `kafka_throughput_control`
Source: https://github.com/redpanda-data/redpanda/blob/v23.3.13/src/v/config/configuration.cc#L2859-L2873
Units: N/A
Default: `[]`
Range: N/A
Restart required: No
Visibility: User
Example: `[{'name': 'first_group','client_id': 'client1'}, {'client_id': 'consumer-\d+'}, {'name': 'catch all'}]`

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
BenPope: I suggested we added some commentary on the exclusions defined by `kafka_throughput_control` and `kafka_throughput_controlled_api_keys`. Maybe it's ok to defer that to the config section that follows.

BenPope: Oh, I see it now! Had an out of date version.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -25,27 +25,20 @@ The properties for broker-wide throughput quota balancing are configured at the
 | xref:reference:cluster-properties.adoc#kafka_throughput_limit_node_out_bps[kafka_throughput_limit_node_out_bps]
 | A broker's total throughput limit for egress Kafka traffic.
 
-| xref:reference:cluster-properties.adoc#kafka_quota_balancer_node_period_ms[kafka_quota_balancer_node_period_ms]
-| The period at which the quota balancer runs to balance throughput quota between a broker's shards.
-
-| xref:reference:cluster-properties.adoc#kafka_quota_balancer_min_shard_throughput_ratio[kafka_quota_balancer_min_shard_throughput_ratio]
-| The lowest value of the throughput quota a shard can get in the process of quota balancing, expressed as a ratio of the default shard quota. If set as `0`, there is no minimum, and if set as `1`, no quota can be taken away by the balancer.
+| kafka_throughput_control
+| List of clients for whom broker-wide limits do not apply
```
BenPope, line 29: Is this specifically worded differently than in the code?

> List of throughput control groups that define exclusions from node-wide throughput limits.

I ask because the configuration allows regex, so it doesn't have to be an exact match, which is somewhat implied by the wording above, but not so much by the text in the code.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,17 @@
...
+== Broker-wide throughput limits
 
-With Redpanda's xref:get-started:architecture.adoc#thread-per-core-model[thread-per-core model], the Kafka API traffic to and from a client connection is processed by a single core (shard). In order to manage throughput quotas efficiently, broker quotas are distributed between shards, and each per-shard quota is in turn shared by all connections served by the shard. Splitting broker quota optimally between shards is done behind the scenes by the quota balancer component.
+Broker-wide throughput quotas account for all Kafka API traffic going in to or out of a broker, with the value of the quota representing the allowed rate of data passing through in one direction. To control the quotas, Redpanda provides configurable rate limits for total ingress and egress traffic. Redpanda also provides configurable properties which allow excluding clients from throughput throttling and give control over which Kafka request types are subject to throttling limits.
```
BenPope: It still feels a bit wordy.

```suggestion
Broker-wide throughput quotas account for ingress and egress of all Kafka API traffic for the broker. It's possible to exclude traffic from throttling based on the `client_id` of a client, or the Kafka request type.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: I'll have a commit in shortly on these two properties. Sorry for the confusion there. I'm mirroring what the code has and providing examples.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: I trimmed it down further, I've just been trying to avoid the terms ingress and egress here. I know it's wordier but I do think it reads cleaner for the audience:

Broker-wide throughput limits account for all Kafka API traffic going in to or out of the broker. The limit values represent the allowed rate of data in bytes per second passing through in each direction. Redpanda also provides administrators the power to exclude clients from throughput throttling and to fine-tune which Kafka request types are subject to throttling limits.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
greg-szrama: Based on Ben's feedback - I have one open question on the kafka_throughput_control property. The code seems ambiguous as to how it behaves if you define a name but no client_id for a throughput control group. The example in the code is:

[{'name': 'first_group','client_id': 'client1'}, {'client_id': 'consumer-\d+'}, {'name': 'catch all'}]

I need to document the behavior for when the list has an entry with a name but no client_id. As it stands, I can omit that example, but it seems to be an important capability/distinction for configuring throughput throttling.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -25,27 +25,20 @@ The properties for broker-wide throughput quota balancing are configured at the
...
-
-| xref:reference:cluster-properties.adoc#kafka_quota_balancer_min_shard_throughput_ratio[kafka_quota_balancer_min_shard_throughput_ratio]
-| The lowest value of the throughput quota a shard can get in the process of quota balancing, expressed as a ratio of the default shard quota. If set as `0`, there is no minimum, and if set as `1`, no quota can be taken away by the balancer.
+| xref:reference:cluster-properties.adoc#kafka_throughput_control[kafka_throughput_control]
+| List of clients for whom broker-wide limits do not apply
 
-| xref:reference:cluster-properties.adoc#kafka_quota_balancer_min_shard_throughput_bps[kafka_quota_balancer_min_shard_throughput_bps]
-| The lowest value of the throughput quota a shard can get in the process of quota balancing, in bytes per second. If set as `0`, there is no minimum.
+| xref:reference:cluster-properties.adoc#kafka_throughput_controlled_api_keys[kafka_throughput_controlled_api_keys]
+| List of Kafka request types subject to broker-wide throughput limits; defaults to `produce` and `fetch.
```
BenPope:
```suggestion
| List of Kafka request types subject to broker-wide throughput limits; defaults to `produce` and `fetch`.

File: modules/reference/pages/cluster-properties.adoc [hunk 37]
```diff
@@ -935,17 +935,21 @@ Each throughput control group consists of:
 
 Example values:
 
-* `[{'name': 'first_group','client_id': 'client1'}, {'name': 'catch all'}]`
-* `[{'client_id': 'consumer-\d+'}]`
+* `[{'name': 'first_group','client_id': 'client1'}, {'client_id': 'consumer-\d+'}]`
+* `[{'name': 'catch all'}]`
 
-A connection is assigned the first matching group and is then excluded from throughput control.
+A connection is assigned the first matching group and is then excluded from throughput control. You can optionally omit the `client_id` and specify only a `name`, as shown. In this situation, all clients will match the rule and Redpanda will exclude them from all from node-wide throughput control.
```
BenPope: Can you check your wording here, please.

You can optionally omit `client_id` independently of `name`

You may specify `"client_id": "+empty"` which matches when a client does not send a `client_id`.

greg-szrama: I revamped the wording to make the permutations more clear.

modules/reference/pages/cluster-properties.adoc
```diff
@@ -899,127 +899,76 @@ Maximum latency threshold for Kafka queue depth control depth tracking.
...
-*Supported versions*: Redpanda v23.1 or later
-
----
-
-=== kafka_quota_balancer_window_ms
-
-Time window used to average the current throughput measurement for the quota balancer.
-
-*Units*: milliseconds
+*Default*: null
```
BenPope:
```default
*Default*: `[]`
```

modules/reference/pages/cluster-properties.adoc
```diff
@@ -899,127 +899,76 @@ Maximum latency threshold for Kafka queue depth control depth tracking.
...
 
-*Default*: 5000
-
-*Range*: [1, ...]
-
-*Restart required*: no
+*Restart required*: false
```
BenPope:
```suggestion
*Restart required*: no

File: modules/reference/pages/cluster-properties.adoc [hunk 40]
```diff
@@ -899,127 +899,76 @@ Maximum latency threshold for Kafka queue depth control depth tracking.
...
 
-*Default*: null
+*Default*: `["produce", "fetch"]`
 
-*Range*: [32 KiB, ...], aligned to 4096 bytes
-
----
-
-=== kafka_rpc_server_tcp_send_buf
+*Restart required*: false
```
BenPope:
```suggestion
*Restart required*: no

File: modules/reference/pages/tunable-properties.adoc [hunk 41]
```diff
@@ -1373,6 +1373,42 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+[[kafka_throughput_throttling_v2]]
+=== kafka_throughput_throttling_v2
+
+Enables an updated algorithm for enforcing node throughput limits based on a shared token bucket, introduced with Redpanda v23.3.8. Set this property to `false` if you need to use the quota balancing algorithm from Redpanda v23.3.7 and older.  This property defaults to `true` for all new or upgraded Redpanda clusters.
+
+*Type*: boolean
+
+*Default*: true
+
+*Restart required*: false
```
BenPope:
```suggestion
*Restart required*: no

File: modules/reference/pages/tunable-properties.adoc [hunk 42]
```diff
@@ -1373,6 +1373,42 @@ Maximum size of the user-space receive buffer. If `null`, this limit is not appl
...
+
+This threshold is evaluated with each request for data. When the number of tokens to replenish exceeds this threshold, then tokens are added to the token bucket. This ensures that the atomic is not being updated for the token count with each request. The range for this threshold is automatically clamped to the corresponding throughput limit for ingress and egress.
+
+*Type*: signed 64-bit integer
+
+*Default*: 1
+
+*Range*: For ingress, [1, xref:reference:cluster-properties.adoc#kafka_throughput_limit_node_in_bps[`kafka_throughput_limit_node_in_bps`]]. For egress, [1, xref:reference:cluster-properties.adoc#kafka_throughput_limit_node_out_bps[`kafka_throughput_limit_node_out_bps`]]
+
+*Restart required*: false
```
BenPope:
```suggestion
*Restart required*: no

**Review by BenPope — COMMENTED** [review]
I think we're almost there!

I noticed that in most places restart is no instead of false. The only that that isn't a nit is throughput_control, which is an empty list, and not nullable.

File: modules/reference/pages/cluster-properties.adoc [hunk 43]
```diff
@@ -899,127 +899,76 @@ Maximum latency threshold for Kafka queue depth control depth tracking.
...
+* `[{'name': 'catch all'}]`
+* `[{'name': 'missing_id', 'client_id': '+empty'}]`
 
-*Default*: 256
+A connection is assigned the first matching group and is then excluded from throughput control. A `name` is not required, but can help you categorize the exclusions. Specifying `+empty` for the `client_id` will match on clients that opt not to send a `client_id`. You can also optionally omit the `client_id` and specify only a `name`, as shown. In this situation, all clients will match the rule and Redpanda will exclude them from all from node-wide throughput control.
 
-*Range*: [0, ...]
+*Type*: list of control groups of the format `{'name' : 'group name', 'client_id' : 'regex pattern'}`
+
+*Default*: null
```
Deflaimun:
```suggestion
*Default*: `[]`
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,17 @@
...
 
-== Broker-wide throughput limits
+== Throughput throttling enforcement
 
-The network bandwidth and disk utilization of brokers may be overloaded by clients that produce or consume throughput without limits. To prevent resource overloading caused by unconstrained throughput and to configure back pressure, Redpanda provides runtime-configurable properties that limit and balance throughput of Kafka API traffic.
+Throughput limits are enforced by applying backpressure to clients. When a connection is in breach of the throughput limit, the throttler advises the client about the delay (throttle time) that would bring the rate back to the allowed level. Redpanda starts by adding a `throttle_time_ms` field to responses. If that isn't honoured, delays are inserted on the connection's next read operation. The throttling delay may not exceed the limit set by xref:reference:tunable-properties.adoc#max_kafka_throttle_delay_ms[`max_kafka_throttle_delay_ms`].
```
Feediver1:
```suggestion
Throughput limits are enforced by applying backpressure to clients. When a connection is in breach of the throughput limit, the throttler advises the client about the delay (throttle time) that would bring the rate back to the allowed level. Redpanda starts by adding a `throttle_time_ms` field to responses. If that isn't honored, delays are inserted on the connection's next read operation. The throttling delay may not exceed the limit set by xref:reference:tunable-properties.adoc#max_kafka_throttle_delay_ms[`max_kafka_throttle_delay_ms`].
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -2,17 +2,17 @@
...
+== Broker-wide throughput limits
 
-With Redpanda's xref:get-started:architecture.adoc#thread-per-core-model[thread-per-core model], the Kafka API traffic to and from a client connection is processed by a single core (shard). In order to manage throughput quotas efficiently, broker quotas are distributed between shards, and each per-shard quota is in turn shared by all connections served by the shard. Splitting broker quota optimally between shards is done behind the scenes by the quota balancer component.
+Broker-wide throughput limits account for all Kafka API traffic going in to or out of the broker. The limit values represent the allowed rate of data in bytes per second passing through in each direction. Redpanda also provides administrators the power to exclude clients from throughput throttling and to fine-tune which Kafka request types are subject to throttling limits.
```
Feediver1:
```suggestion
Broker-wide throughput limits account for all Kafka API traffic going in to or out of the broker. The limit values represent the allowed rate of data in bytes per second passing through in each direction. Redpanda also provides administrators the ability to exclude clients from throughput throttling and to fine-tune which Kafka request types are subject to throttling limits.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -25,27 +25,20 @@ The properties for broker-wide throughput quota balancing are configured at the
...
-| xref:reference:cluster-properties.adoc#kafka_quota_balancer_window_ms[kafka_quota_balancer_window_ms]
-| The time window the balancer uses to average the current throughput measurement.
 |===
 
 [NOTE]
 ====
-* By default, both `kafka_throughput_limit_node_in_bps` and `kafka_throughput_limit_node_out_bps` are disabled, no throughput limits are applied. You must manually set them to enable quota balancing with throughput limits.
-* `kafka_quota_balancer_min_shard_throughput_bps` doesn't override the `kafka_throughput_limit_node_in_bps` and `kafka_throughput_limit_node_out_bps` limit settings. Consequently, the value of
-`kafka_throughput_limit_node_in_bps` or `kafka_throughput_limit_node_out_bps` can result in lesser throughput than `kafka_quota_balancer_min_shard_throughput_bps`.
+* By default, both `kafka_throughput_limit_node_in_bps` and `kafka_throughput_limit_node_out_bps` are disabled, no throughput limits are applied. You must manually set them to enable throughput throttling.
```
Feediver1:
```suggestion
* By default, both `kafka_throughput_limit_node_in_bps` and `kafka_throughput_limit_node_out_bps` are disabled, and no throughput limits are applied. You must manually set them to enable throughput throttling.
```

Feediver1 (approved):

PR #422, opened 2024-04-16
Bump supported Helm version

RafalKorepta:
In https://github.com/redpanda-data/helm-charts/pull/1199 the minimum supported Helm version was changed. This commit sync that version


JakeSCahill (approved):
Thanks @RafalKorepta. We should wait until this is merged: https://github.com/redpanda-data/helm-charts/pull/1199/files

Also, I'll need to run our automation to update the compatibility matrix.

PR #425, opened 2024-04-17
Update from write_caching on/off to write_caching_default true/false

nvartolomei:
* Cluster level `write_caching` was renamed to `write_caching_default`
* Cluster level write caching values {off, on} were renamed to {false, true}
* Topic level `write.caching` values {off, on} were renamed to {false, true}

### DIFF
--- a/modules/develop/pages/config-topics.adoc
+++ b/modules/develop/pages/config-topics.adoc
@@ -95,7 +95,7 @@ rpk topic alter-config [TOPICS…] —-set cleanup.policy=compact
 
 Write caching is a relaxed mode of xref:develop:produce-data/configure-producers.adoc#acksall[`acks=all`] that provides better performance at the expense of durability. It acknowledges a message as soon as it is received and acknowledged on a majority of brokers, without waiting for it to be written to disk. This provides lower latency while still ensuring that a majority of brokers acknowledge the write.
 
-Write caching applies to user topics. It does not apply to transactions or consumer offsets: data written in the context of a transaction and consumer offset commits are always fsynced.
+Write caching applies to user topics. It does not apply to transactions or consumer offsets: data written in the context of a transaction and consumer offset commits is always written to disk and fsynced before being acknowledged to the client.
 
 NOTE: For clusters in xref:reference:rpk/rpk-redpanda/rpk-redpanda-mode.adoc#development-mode[development mode], write caching is enabled by default. For clusters in production mode, it is disabled by default. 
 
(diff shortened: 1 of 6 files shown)


Feediver1 (approved):
lgtm

modules/develop/pages/config-topics.adoc
```diff
@@ -93,27 +93,27 @@ rpk topic alter-config [TOPICS…] —-set cleanup.policy=compact
 
 === Configure write caching
 
-Write caching is a relaxed mode of xref:develop:produce-data/configure-producers.adoc#acksall[`acks=all`] that provides better performance at the expense of durability. It acknowledges a message as soon as it is received and acknowledged on a majority of brokers, without waiting for it to fsync to disk. This provides lower latency while still ensuring that a majority of brokers acknowledge the write. 
+Write caching is a relaxed mode of xref:develop:produce-data/configure-producers.adoc#acksall[`acks=all`] that provides better performance at the expense of durability. It acknowledges a message as soon as it is received and acknowledged on a majority of brokers, without waiting for it to be written to disk. This provides lower latency while still ensuring that a majority of brokers acknowledge the write.
 
-Write caching applies to user topics. It does not apply to transactions or consumer offsets:  transactions and offset commits are always fsynced. 
+Write caching applies to user topics. It does not apply to transactions or consumer offsets: data written in the context of a transaction and consumer offset commits are always fsynced.
```
Deflaimun:
```suggestion
Write caching applies to user topics. It does not apply to transactions or consumer offsets: data written in the context of a transaction and consumer offset commits are always written to disk.
```

Since this PR updates fsync to written to disk, should we also update this section?

nvartolomei: You prompted me to think more about this, I think this does a better job:

> Write caching applies to user topics. It does not apply to transactions or consumer offsets: data written in the context of a transaction and consumer offset commits is always written to disk and fsynced before being acknowledged to the client.

Writing to disk and fsync are 2 different operations. I changed some `fsynced` to `written` because saying `not fsync` still leaves the user with the question "is it written to disk or not?" The updated wording explicitly says that we ack the data before it even reaches disk. In the case of transactions, for the same reason, we better explicitly mention that we do fsync the data and not just write it. It is a much stronger guarantee and a more precise definition of what is happening under the hood.

Deflaimun: that sounds much better. Thanks for this @nvartolomei

bharathv (approved):

PR #427, opened 2024-04-17
Adds back the required page alias

JakeSCahill:
Adds back the alias that was removed in https://github.com/redpanda-data/docs/pull/424

Now that we have the [`modify-redirect` extension](https://github.com/redpanda-data/docs-extensions-and-macros/pull/36), the production site will no longer include redirect loops.

Note: Local builds will still experience redirect loops. We are working on a fix for this too.


Deflaimun (approved):

Deflaimun:
Do we want to merge the changes in package-lock?

JakeSCahill:
> Do we want to merge the changes in package-lock?

Yeah it just makes sure you get the required version of the extension when doing local builds.

PR #429, opened 2024-04-18
Add Policy permissions to CloudStorage bucket documentation

frenchfrywpepper:
The `s3:PutBucketPolicy` and `s3:DeleteBucketPolicy` permissions were recently introduced in order to support remote read replicas in cloud.

Also modified the - - to a * for Get and List, as this is more correct.


Feediver1 (approved):
Thanks @frenchfrywpepper !

PR #434, opened 2024-04-22
How to enable CPU pinning in Kubernetes

JakeSCahill:
Fixes https://github.com/redpanda-data/documentation-private/issues/2356

Preview: https://deploy-preview-434--redpanda-docs-preview.netlify.app/current/manage/kubernetes/k-manage-resources/

Note: Planning to add some images to illustrate CPU pinning vs shared pool.

Related issue: https://github.com/redpanda-data/helm-charts/issues/1230


modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,63 @@
...
+
+- **Loss of CPU cache affinity**: As processes move across cores, they lose the cache locality, leading to increased cache misses and higher latency.
+- **Scheduling latency**: Dynamic CPU allocation can introduce delays as the scheduler determines which core a process should run on, based on current load and availability.
+
+Enabling CPU pinning for Redpanda in Kubernetes requires the following:
+
+- *Static CPU Manager policy*:
++
+To enable CPU pinning in Kubernetes, you must enable the https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` CPU Manager policy], which bypasses the default CPU allocation behavior managed by the CFS.
```
JakeSCahill:
```suggestion
- *Static CPU Manager policy*: To enable CPU pinning in Kubernetes, you must enable the https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` CPU Manager policy], which bypasses the default CPU allocation behavior managed by the CFS.
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,63 @@
...
++
+When the `static` policy is enabled, each Pod is assigned specific CPU cores, and these assignments remain constant throughout the Pod's lifecycle. This fixed allocation is crucial for Redpanda in production, because it maintains CPU cache affinity. Cache affinity refers to the performance benefit that arises when a process is able to repeatedly access data from the same CPU cache, rather than re-fetching the data from main memory or a distant cache associated with another CPU core.
+- *Overprovisioning disabled (default)*:
++
+When overprovisioning is disabled in Redpanda, each thread runs on a dedicated core for maximum performance.
```
JakeSCahill:
```suggestion
- *Overprovisioning disabled (default)*: When overprovisioning is disabled in Redpanda, each thread runs on a dedicated core for maximum performance.
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -122,17 +118,21 @@ helm upgrade --install redpanda redpanda/redpanda --namespace <namespace> --crea
...
 . <<qos, Enable the `Guaranteed` QoS class for your Pods that run Redpanda>>.
++
+Only containers that are both part of a `Guaranteed` Pod and have integer CPU requests are assigned exclusive CPUs.
 
 [[pinning]]
 === Disable pinning in shared environments
 
 On a dedicated worker node, CPU pinning allows maximum throughput, but in a development environment, where you may run other Pods on the same worker node as Redpanda, CPU pinning can cause your other processes to run slower than usual. In these shared environments, you can enable overprovisioning so that Redpanda reduces the aggressiveness of CPU usage to prevent resource hogging and ensuring fair CPU time distribution among all processes.
 
+CAUTION: When `resources.cpu.cores` is set to a fractional value, such as `1100m`, Redpanda may not use all the allocated CPU due to CFS quotas, leading to potential performance bottlenecks.
```
chrisseto: Maybe include both forms of fractional values here? 
```suggestion
CAUTION: When `resources.cpu.cores` is set to a fractional value, such as `1100m` or `1.1`, the Pod is not eligible for exclusive access to the cores which leads to performance bottlenecks.
```

chrisseto: I feel this doesn't convey the core issue just the side effect. Putting on my user hat I feel that "CFS quotas" is suddenly being lobbed at me and I likely won't understand why CFS is a problem in this content.

```suggestion
CAUTION: When `resources.cpu.cores` is set to a fractional value, such as `1100m`, the Redpanda Pod is not eligible for exclusive core access which will lead to potential performance bottlenecks.
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -38,27 +38,23 @@ Redpanda is designed to maximize throughput by xref:get-started:architecture.ado
...
+To enable CPU pinning in Kubernetes, you must enable the https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` policy] for the CPU manager. With this policy enabled, the CPU Manager controls a shared pool of CPUs. Initially, this shared pool contains all the CPUs in the worker node. However, when a container is created with an integral CPU request in a https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed[`Guaranteed` Pod], CPUs for that container are removed from the shared pool and assigned exclusively to that container for its lifetime. Other containers are migrated off these exclusively allocated CPUs.
 
 === Enable CPU pinning
 
+To enable CPU pinning for Redpanda in Kubernetes:
+
 . Ensure that your worker nodes have enough CPU cores to allocate to Redpanda. Sharing cores with other applications can negate the benefits of CPU pinning.
```
chrisseto: This line feels like a red herring to me. K8s won't schedule the Redpanda Pod if it can't provide it with the # of CPUs it's requested. The issue of sharing cores would really only pop up if _not_ using the static CPU policy and additional Pods are being scheduled on the same node a redpanda.

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -38,27 +38,23 @@ Redpanda is designed to maximize throughput by xref:get-started:architecture.ado
...
 
-. Ensure the https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` policy] is enabled on your worker nodes.
+. Ensure the https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` policy] is enabled for the CPU Manager on your Kubernetes cluster. With this policy enabled, a request for an integral CPU amount in a `Guaranteed` Pod will result in that container being allocated whole CPUs with a guarantee that no other container will run on them.
```
chrisseto: How about:
```suggestion
. Ensure your Kubernetes cluster is using the  https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` CPU Manager]. With this policy enabled, a request for an integral CPU amount in a `Guaranteed` Pod will result in that container being allocated exclusive access to that number of CPUs through [cgroup's cpuset](https://www.kernel.org/doc/Documentation/cgroup-v1/cpusets.txt).
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,55 @@
...
+- <<cpu, Enable CPU pinning>>. This configuration allows Redpanda to maintain exclusive access to designated CPU cores, improving throughput and reducing latency caused by CPU context switches.
+
+- <<qos, Enable the `Guaranteed` quality of service class for Pods that run Redpanda>>. This setup ensures that the CPU and memory allocated to Redpanda are not subject to throttling or other contention issues, providing a stable and predictable performance environment.
+
+- <<memory, Enable memory locking>>. This configuration prevents the operating system from paging out Redpanda's memory to disk, which can significantly impact performance.
+
+== Limitations
 
-See xref:deploy:deployment-option/self-hosted/kubernetes/k-requirements.adoc[] for memory, CPU, and storage.
+You cannot decrease the number of CPU cores in a running cluster.
```
chrisseto: Is there a piece of documentation explaining this a bit further? I understand the limitation but prior to fully understanding it I immediately think "That's some bullshit or your handwaving around a detail that's critically important for me to know".

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,55 @@
...
 
-Make sure that your physical or virtual machines have enough resources to give to Redpanda. You assign resources to each Redpanda broker using the `values.yaml` settings. If you need to replace a worker node with one that has more resources, xref:./decommission-brokers.adoc[decommission the brokers] before you replace the worker nodes.
+[[cpu]]
+== CPU pinning
 
-== Configure CPU resources
+Redpanda is designed to maximize throughput by xref:get-started:architecture.adoc#tpc[pinning its threads to specific CPU cores]. However, CPU pinning is disabled by default in Kubernetes because it uses the https://www.kernel.org/doc/html/latest/scheduler/sched-design-CFS.html[Completely Fair Scheduler (CFS)] to dynamically manage CPU allocation. The CFS distributes CPU time among all processes based on their priority and resource requests, which can lead to frequent context switches and migration of processes across different cores. This behavior can degrade the performance of Redpanda due to:
```
chrisseto: It's not explicitly disabled by anything in the Kubernetes layer. I am curious to know if seastar is doing any type of detection. How about saying "CPU pinning is, by default, unavailable in Kubernetes due to its usage of the CFS"?

It may also be useful to introduce the term "hardware locality here"? Another way to explain the performance impact could be something like: "Redpanda and Seastar rely on hardware locality for optimal performance. CFS is unaware of hardware locality and can result in a negative performance impact due to:"

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,55 @@
...
 
-By default, Redpanda pins its threads to all cores that you allocate to it. The more cores that are allocate to Redpanda, the more throughput it can support. You can configure the CPU resources allocated to each Redpanda broker by overriding the default values. For a description of CPU resource units, see the https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu[Kubernetes documentation^].
+- **Loss of CPU cache affinity**: As processes move across cores, they lose the cache locality, leading to increased cache misses and higher latency.
```
chrisseto, line 38: 👨‍🍳 💋

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,55 @@
...
+To enable CPU pinning for Redpanda in Kubernetes:
+
+. Ensure that your worker nodes have enough CPU cores to allocate to Redpanda. Sharing cores with other applications can negate the benefits of CPU pinning.
+
+. Ensure the https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` policy] is enabled for the CPU Manager on your Kubernetes cluster. With this policy enabled, a request for an integral CPU amount in a `Guaranteed` Pod will result in that container being allocated whole CPUs with a guarantee that no other container will run on them.
++
+[source,bash]
+----
+# Check if the static policy is enabled
+cat /var/lib/kubelet/config.yaml | grep cpuManagerPolicy
```
chrisseto: I think we'll want to change this heuristic to something more generally available. Most users won't have access to the kubelet config yaml or won't want to run a privileged pod to get access. It should be sufficient to check if the Pod is in fact running at a Guarenteed QoS and then check if the redpanda Pod can observe whether or not it can see that a cpuset has been assigned to it. It may even be helpful for us to add a check in the container startup so we can log out a _very_ obvious line that says "HEY NO CPUSET FOUND ITS GONNA BE SLOWER".

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -77,34 +107,31 @@ helm upgrade --install redpanda redpanda/redpanda --namespace <namespace> --crea
...
+. <<qos, Enable the `Guaranteed` QoS class for your Pods that run Redpanda>>.
++
+Only containers that are both part of a `Guaranteed` Pod and have integer CPU requests are assigned exclusive CPUs.
 
+[[pinning]]
 === Disable pinning in shared environments
 
-By default, Redpanda pins its threads to all cores that you allocate to it.
-On a dedicated worker node, CPU pinning allows maximum throughput, but in a development environment, where you may run other Pods on the same worker node as Redpanda, CPU pinning can cause your processes to run slower than usual. In these shared environments, you can tell Redpanda not to pin its threads or memory, and reduce the amount of polling it does to a minimum.
+On a dedicated worker node, CPU pinning allows maximum throughput, but in a development environment, where you may run other Pods on the same worker node as Redpanda, CPU pinning can cause your other processes to run slower than usual. In these shared environments, you can enable overprovisioning so that Redpanda reduces the aggressiveness of CPU usage to prevent resource hogging and ensuring fair CPU time distribution among all processes.
```
chrisseto: If I understand correctly, CPU Pinning + CPUSets would mitigate the need for dedicates kubernetes nodes _and_ it would make repanda steal less CPU from other Pods as well? I think it's good to tell users how to disable pinning, if they want, but the premise under which this concept is introduced feels incorrect to me.

chrisseto (commented):
This is a crazy complex topic to broach. You did a great job; I'm being _very_ nit picky as I worry that any degree of incorrectness could lead end users astray. It's also useful to push the limits on our own understanding. (So take my suggestions with a grain of salt).

I do wonder if it's worth taking a step back and thinking about this a bit more as a feature rather than a documentation gap. For example, we could add log lines to the redpanda container so users don't have to run commands manually or we could make the `--overprovisioned` flag dependent on whether or not the resource requests make the Pod eligible for having exclusive core access. That could also be a follow up item as this is a massive improvement as is 🔥

modules/manage/pages/kubernetes/k-manage-resources.adoc
JakeSCahill: The only thing I could find was [this issue](https://github.com/redpanda-data/redpanda/issues/3668) where @dotnwat says that we don't allow brokers to restart with a reduced number of cores.

modules/manage/pages/kubernetes/k-manage-resources.adoc
JakeSCahill:
```suggestion
Redpanda is designed to maximize throughput by xref:get-started:architecture.adoc#tpc[pinning its threads to specific CPU cores]. However, CPU pinning is unavailable by default in Kubernetes because it uses the https://www.kernel.org/doc/html/latest/scheduler/sched-design-CFS.html[Completely Fair Scheduler (CFS)] to dynamically manage CPU allocation. The CFS distributes CPU time among all processes based on their priority and resource requests, which can lead to frequent context switches and migration of processes across different cores. This behavior can degrade the performance of Redpanda due to:
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,55 @@
...
 
-For production clusters, you should never set `resources.cpu.cores` to anything less than full cores.
+To enable CPU pinning in Kubernetes, you must enable the https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` policy] for the CPU manager. With this policy enabled, the CPU Manager controls a shared pool of CPUs. Initially, this shared pool contains all the CPUs in the worker node. However, when a container is created with an integral CPU request in a https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed[`Guaranteed` Pod], CPUs for that container are removed from the shared pool and assigned exclusively to that container for its lifetime. Other containers are migrated off these exclusively allocated CPUs.
+
+=== Enable CPU pinning
+
+To enable CPU pinning for Redpanda in Kubernetes:
+
+. Ensure that your worker nodes have enough CPU cores to allocate to Redpanda. Sharing cores with other applications can negate the benefits of CPU pinning.
```
JakeSCahill:
```suggestion
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
JakeSCahill: Removed.

modules/manage/pages/kubernetes/k-manage-resources.adoc
JakeSCahill: I created an issue to log this info: https://github.com/redpanda-data/helm-charts/issues/1247

@chrisseto please could you provide the commands we should document instead?

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -165,15 +191,25 @@ helm upgrade --install redpanda redpanda/redpanda --namespace <namespace> --crea
 --
 ======
 
+If you're experimenting with Redpanda in Kubernetes, you can also set the number of CPU cores to millicores to automatically disable CPU pinning.
```
JakeSCahill:
```suggestion
If you're experimenting with Redpanda in Kubernetes, you can also set the number of CPU cores to millicores to automatically enable overprovisioning.
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -77,34 +106,31 @@ helm upgrade --install redpanda redpanda/redpanda --namespace <namespace> --crea
...
+. <<qos, Enable the `Guaranteed` QoS class for your Pods that run Redpanda>>.
++
+Only containers that are both part of a `Guaranteed` Pod and have integer CPU requests are assigned exclusive CPUs.
 
+[[pinning]]
 === Disable pinning in shared environments
 
-By default, Redpanda pins its threads to all cores that you allocate to it.
-On a dedicated worker node, CPU pinning allows maximum throughput, but in a development environment, where you may run other Pods on the same worker node as Redpanda, CPU pinning can cause your processes to run slower than usual. In these shared environments, you can tell Redpanda not to pin its threads or memory, and reduce the amount of polling it does to a minimum.
+On a dedicated worker node, CPU pinning allows maximum throughput, but in a development environment, where you may run other Pods on the same worker node as Redpanda, CPU pinning can cause your other processes to run slower than usual. In these shared environments, you can enable overprovisioning so that Redpanda reduces the aggressiveness of CPU usage to prevent resource hogging and ensuring fair CPU time distribution among all processes.
```
JakeSCahill:
```suggestion
If you don't enable CPU pinning and Redpanda runs in a shared environment, you can make Redpanda less aggressive in CPU usage by enabling overprovisioning. This adjustment ensures a fairer distribution of CPU time among all processes, improving overall system efficiency at the cost of Redpanda's performance.

You can enable overprovisioning by either setting the CPU request to a fractional value or setting `overprovisioned` to `true`. 
```

@chrisseto is this better?

modules/manage/pages/kubernetes/k-manage-resources.adoc
JakeSCahill: https://github.com/redpanda-data/docs/pull/434#discussion_r1577728613

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,54 @@
...
 
-See xref:deploy:deployment-option/self-hosted/kubernetes/k-requirements.adoc[] for memory, CPU, and storage.
+You cannot decrease the number of CPU cores in a running cluster.
 
-Make sure that your physical or virtual machines have enough resources to give to Redpanda. You assign resources to each Redpanda broker using the `values.yaml` settings. If you need to replace a worker node with one that has more resources, xref:./decommission-brokers.adoc[decommission the brokers] before you replace the worker nodes.
+[[cpu]]
+== CPU pinning
 
-== Configure CPU resources
+Redpanda is designed to maximize throughput by xref:get-started:architecture.adoc#tpc[pinning its threads to specific CPU cores]. However, CPU pinning is unavailable by default in Kubernetes because it uses the https://www.kernel.org/doc/html/latest/scheduler/sched-design-CFS.html[Completely Fair Scheduler (CFS)] to dynamically manage CPU allocation. The CFS distributes CPU time among all processes based on their priority and resource requests, which can lead to frequent context switches and migration of processes across different cores. This behavior can degrade the performance of Redpanda due to:
```
JakeSCahill:
```suggestion
Redpanda is designed to maximize throughput by xref:get-started:architecture.adoc#tpc[pinning its threads to specific CPU cores] (CPU pinning). However, CPU pinning is unavailable by default in Kubernetes because it uses the https://www.kernel.org/doc/html/latest/scheduler/sched-design-CFS.html[Completely Fair Scheduler (CFS)] to dynamically manage CPU allocation. The CFS distributes CPU time among all processes based on their priority and resource requests, which can lead to frequent context switches and migration of processes across different cores. This behavior can degrade the performance of Redpanda due to:
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
JakeSCahill: This idea is already introduced in the bulleted list below this paragraph.

modules/manage/pages/kubernetes/k-manage-resources.adoc
JakeSCahill: Users need to know that the policy is enabled before they deploy Redpanda. So checking if the Pod is running at a Guarenteed QoS won't work. Should we instead not provide a command and link to the Kubernetes docs for info?

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,58 @@
...
+
+image::shared:guaranteed-pod.png[High-level view of Guaranteed Pods with exclusive CPUs]
+
+=== Enable CPU pinning
+
+To enable CPU pinning for Redpanda in Kubernetes:
+
+
+. Ensure your Kubernetes cluster is using the  https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` CPU Manager]. With this policy enabled, a request for an integral CPU amount in a `Guaranteed` Pod will result in that container being allocated exclusive access to that number of CPUs through [cgroup's cpuset](https://www.kernel.org/doc/Documentation/cgroup-v1/cpusets.txt).
```
JakeSCahill:
```suggestion
. Ensure your Kubernetes cluster is using the  https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` CPU Manager]. With this policy enabled, a request for an integral CPU amount in a `Guaranteed` Pod will result in that container being allocated exclusive access to that number of CPUs through https://www.kernel.org/doc/Documentation/cgroup-v1/cpusets.txt[cgroup's cpuset].
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
chrisseto: Dang, yeah that's probably the best thing we've got for now. It's not a great reference but I think it would be useful for users to understand this is a limitation of redpanda/seastar not anything k8s adjacent.

modules/manage/pages/kubernetes/k-manage-resources.adoc
chrisseto: That issue is perfect. I wonder if there's a way to check for QoS within a Pod 🤔 

Yeah let's point to k8s docs for now and we can update this once we tackle the linked issue.

modules/manage/pages/kubernetes/k-manage-resources.adoc
chrisseto: LGTM but I have one tiny suggestion: Instead of `, improving overall system efficiency.` can we call out that over provisioned will decrease RP's perf? `, improving overall system efficiency at the cost of redpanda's performance`

chrisseto (approved):
LGTM!

modules/manage/pages/kubernetes/k-manage-resources.adoc
JakeSCahill: I will clarify that this is a Redpanda limitation 👍

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,58 @@
...
+- <<cpu, Enable CPU pinning>>. This configuration allows Redpanda to maintain exclusive access to designated CPU cores, improving throughput and reducing latency caused by CPU context switches.
+
+- <<qos, Enable the `Guaranteed` quality of service class for Pods that run Redpanda>>. This setup ensures that the CPU and memory allocated to Redpanda are not subject to throttling or other contention issues, providing a stable and predictable performance environment.
+
+- <<memory, Enable memory locking>>. This configuration prevents the operating system from paging out Redpanda's memory to disk, which can significantly impact performance.
+
+== Limitations
 
-See xref:deploy:deployment-option/self-hosted/kubernetes/k-requirements.adoc[] for memory, CPU, and storage.
+You cannot decrease the number of CPU cores in a running cluster.
```
JakeSCahill:
```suggestion
Redpanda does not support decreasing the CPU cores for brokers in an existing cluster.
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,58 @@
...
+
+To enable CPU pinning for Redpanda in Kubernetes:
+
+
+. Ensure your Kubernetes cluster is using the  https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy[`static` CPU Manager]. With this policy enabled, a request for an integral CPU amount in a `Guaranteed` Pod will result in that container being allocated exclusive access to that number of CPUs through https://www.kernel.org/doc/Documentation/cgroup-v1/cpusets.txt[cgroup's cpuset].
++
+[source,bash]
+----
+# Check if the static policy is enabled
+cat /var/lib/kubelet/config.yaml | grep cpuManagerPolicy
```
JakeSCahill:
```suggestion
```

modules/manage/pages/kubernetes/k-manage-resources.adoc
```diff
@@ -7,27 +7,58 @@
...
+----
```
JakeSCahill:
```suggestion
```

Feediver1 (approved):
lgtm

PR #452, opened 2024-04-25
Update disk-utilization.adoc

jason-da-redpanda:
It reads currently like this param is needed if you are on 22.2.27 for example

that is not the case...

this feature is only in 23.1 and higher 

this param is needed on any cluster that has been upgraded to 23.1* or higher from an earlier version than 23.1

### DIFF
--- a/modules/manage/pages/cluster-maintenance/disk-utilization.adoc
+++ b/modules/manage/pages/cluster-maintenance/disk-utilization.adoc
@@ -158,7 +158,7 @@ For periodic offset expiration, set the retention duration of consumer group off
 | Retention duration of consumer group offsets.
 
 | xref:reference:tunable-properties.adoc#legacy_group_offset_retention_enabled[`legacy_group_offset_retention_enabled`]
-| Enable group offset retention for Redpanda versions earlier than v23.1.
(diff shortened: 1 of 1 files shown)


modules/manage/pages/cluster-maintenance/disk-utilization.adoc
```diff
@@ -158,7 +158,7 @@ For periodic offset expiration, set the retention duration of consumer group off
 | Retention duration of consumer group offsets.
 
 | xref:reference:tunable-properties.adoc#legacy_group_offset_retention_enabled[`legacy_group_offset_retention_enabled`]
-| Enable group offset retention for Redpanda versions earlier than v23.1.
+| Enable group offset retention for Clusters that have been upgraded from Redpanda versions earlier than v23.1.
```
micheleRP:
```suggestion
| Enable group offset retention for Redpanda clusters upgraded from versions prior to v23.1.
```

micheleRP: Thanks @jason-da-redpanda! Are you OK with this edit?

micheleRP (approved):
Thanks @jason-da-redpanda! If my edit looks good to you, we can merge this.

jason-da-redpanda (commented):
fine by me

PR #462, opened 2024-04-29
Add SASL config to the Topic resource

JakeSCahill:
A community member reported that our Managed Cloud tutorials don't work when using the Redpanda Operator. This issue was due to the Topic resource not having the required SASL configuration settings. This PR fixes that issue and also adds a step to the uninstallation section to remove the alias.


RafalKorepta (approved):

Feediver1 (approved):
lgtm

PR #468, opened 2024-05-01
Update partner-integration.adoc

Feediver1:
Fixes https://github.com/redpanda-data/documentation-private/issues/2401

[Preview](https://deploy-preview-468--redpanda-docs-preview.netlify.app/current/reference/partner-integration/)


micheleRP (approved):
Addition looks good, but should we also move this out of Reference?

PR #475, opened 2024-05-02
#2413 consolidate tier info for APIs

micheleRP:
fixes https://github.com/redpanda-data/documentation-private/issues/2413

Review deadline: Thursday, May 16

Right now, we have Redpanda Cloud supported [tiers](https://docs.redpanda.com/current/deploy/deployment-option/cloud/cloud-overview/#cluster-tiers) and [regions](https://docs.redpanda.com/current/deploy/deployment-option/cloud/regions/) in separate places in the doc. This PR consolidates the tiers and regions and it separates out the content for BYOC and Dedicated for better readability.

In the UI, you only see the options available based on the provider, tier, and region you select. But API users need to input their selection, so the [API docs](https://docs.api.redpanda.com/) list additional details like zones/machine types. After the Cloud API reference doc is published, we'll link to that for API users. 

It also simplifies the Cloud side-nav:
- Redpanda Cloud Overview
- What’s New in Redpanda Cloud
- Serverless
- BYOC
- Dedicated
- Create a Topic
- Manage Billing

preview:
- [BYOC Throughput Tiers](https://deploy-preview-475--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/cloud/byoc-tiers/)
- [Dedicated Throughput Tiers](https://deploy-preview-475--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/cloud/dedicated/dedicated-tiers/)
- [Cloud Overview](https://deploy-preview-475--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/cloud/cloud-overview/#dedicated-cloud)


modules/deploy/partials/cloud/tiers.adoc
```diff
@@ -0,0 +1,610 @@
...
+Amazon Web Services (AWS)::
++
+--
+|=== 
+| Region | Zones | Available throughput tiers
+| af-south-1 
+a| * afs1-az1
+* afs1-az2 
+* afs1-az3
+a|
```
kbatuigas: If there are no available throughput tiers for a region / zone, does that mean the region and zone itself is not available for this cluster type?

modules/deploy/pages/deployment-option/cloud/tiers.adoc
```diff
@@ -0,0 +1,10 @@
+= Throughput Tiers
+:description: Select a throughput tier for your Redpanda Cloud cluster.
+:page-context-links: [{"name": "BYOC", "to": "deploy:deployment-option/cloud/tiers.adoc" },{"name": "Dedicated", "to": "deploy:deployment-option/cloud/dedicated/dedicated-tiers.adoc" } ]
```
kbatuigas: I don't know enough about context links so maybe this isn't possible: I wonder if we should also add the "byoc" indication the BYOC page/URL. I don't think it's a big deal but maybe there is some benefit

modules/ROOT/nav.adoc
```diff
@@ -58,17 +58,19 @@
 ** xref:deploy:deployment-option/cloud/index.adoc[Cloud]
 *** xref:deploy:deployment-option/cloud/cloud-overview.adoc[Redpanda Cloud Overview]
 *** xref:deploy:deployment-option/cloud/whats-new-cloud.adoc[]
+*** xref:deploy:deployment-option/cloud/serverless.adoc[Serverless]
+*** xref:deploy:deployment-option/cloud/provision-a-byoc-cluster/index.adoc[BYOC]
```
kbatuigas, line 62: This is just to do a high level rearrangement of Cloud nav, right? So we're not yet looking into, say, also moving the VPC peering docs under BYOC and Dedicated (not sure if we were going to do that with the cloud migration either way)

modules/deploy/partials/cloud/tiers.adoc
micheleRP: Right, I've removed them!

modules/deploy/pages/deployment-option/cloud/tiers.adoc
micheleRP: good idea! I've renamed it to byoc-tiers.adoc, so it appears in the URL.

modules/ROOT/nav.adoc
micheleRP: Yeah, not yet! Let's do that as part of the Cloud doc reorg.

kbatuigas (approved):

modules/deploy/pages/deployment-option/cloud/create-byoc-cluster-aws.adoc
```diff
@@ -8,7 +8,7 @@ To create a Redpanda cluster in your virtual private cloud (VPC), log in to http
 == Create a BYOC cluster
 
 . On the Clusters page, click *Create cluster*, then click *Create BYOC cluster*.
-. Enter a cluster name, then select the namespace, provider (AWS), xref:deploy:deployment-option/cloud/regions.adoc[region], availability zone type (single AZ or multi AZ) and zones, xref:deploy:deployment-option/cloud/cloud-overview.adoc#cluster-tiers[tier], and Redpanda version.
+. Enter a cluster name, then select the namespace, provider (AWS), xref:deploy:deployment-option/cloud/byoc-tiers.adoc[region, availability zone type (single AZ or multi AZ), zones], and Redpanda version.
```
Feediver1: Just checking...so you specify an AZ type, a region, and a zone, separately?

modules/deploy/partials/cloud/tiers.adoc
```diff
@@ -0,0 +1,568 @@
...
++
+--
+|=== 
+| Region | Zones | Available throughput tiers
+| asia-south1 
+a| * asia-south1-a 
+* asia-south1-b 
+* asia-south1-c 
+a| * tier-1-gcp-v2-x86
+* tier-2-gcp-v2-x86
```
Feediver1: So this region does not get Tiers 8 & 9?

modules/deploy/pages/deployment-option/cloud/create-byoc-cluster-aws.adoc
micheleRP: Yes here's the UI:
image

modules/deploy/partials/cloud/tiers.adoc
micheleRP: correct, not today

modules/deploy/pages/deployment-option/cloud/byoc-tiers.adoc
```diff
@@ -0,0 +1,11 @@
+= Throughput Tiers
+:description: Select a throughput tier for your Redpanda Cloud cluster.
+:page-context-links: [{"name": "BYOC", "to": "deploy:deployment-option/cloud/byoc-tiers.adoc" },{"name": "Dedicated", "to": "deploy:deployment-option/cloud/dedicated/dedicated-tiers.adoc" } ]
+:page-aliases: deploy:deployment-option/cloud/regions.adoc
+:page-cloud: true
+:env-byoc: true
+
+NOTE: Serverless clusters are not on tiers. See xref:deploy:deployment-option/cloud/serverless.adoc#limits[Serverless limits]. To change your cluster type, contact https://support.redpanda.com/hc/en-us/requests/new[support^].
```
towfiqa: We should improve the wording here. "Tiers don't apply to Serverless Clusters" or "Serverless Clusters don't need a throughput tier" @vladoschreiner maybe has a better wording here.

modules/deploy/pages/deployment-option/cloud/byoc-tiers.adoc
```diff
@@ -0,0 +1,11 @@
+= Throughput Tiers
```
towfiqa: No explanation is given about what a Throughput Tier is. We should link to some sort of definition. Customers should understand WHY they need a tier.

modules/deploy/pages/deployment-option/cloud/cloud-overview.adoc
```diff
@@ -30,8 +30,7 @@ NOTE: Serverless is currently in a limited availability (LA) release with xref:d
 === Dedicated Cloud
 
 With Dedicated clusters, you host your data in Redpanda's VPC, and Redpanda handles provisioning, operations, and maintenance. Dedicated clusters are single-tenant deployments that support private networking (that is, VPC peering to talk over private IPs) for better data isolation.
```
towfiqa: The statement "host your data in Redpanda's VPC" is misleading. Re-word.

modules/deploy/pages/deployment-option/cloud/cloud-overview.adoc
```diff
@@ -30,8 +30,7 @@ NOTE: Serverless is currently in a limited availability (LA) release with xref:d
...
-
-Available on AWS or GCP. 
+When you create a Dedicated cluster, you select a xref:deploy:deployment-option/cloud/dedicated/dedicated-tiers.adoc[throughput tier and region] on either AWS or GCP.
```
towfiqa: Why do I need to select a throughput tier. This should be covered somewhere and linked in this paragraph.

modules/deploy/pages/deployment-option/cloud/cloud-overview.adoc
```diff
@@ -40,10 +39,7 @@ To start using Dedicated, contact https://redpanda.com/try-redpanda?section=ente
 === Bring Your Own Cloud (BYOC)
 
 With BYOC clusters, you deploy Redpanda in your own VPC, so all data is
```
towfiqa: This is only for BYO VPC. Re-word for standard BYOC.

modules/deploy/pages/deployment-option/cloud/create-byoc-cluster-aws.adoc
```diff
@@ -8,7 +8,7 @@ To create a Redpanda cluster in your virtual private cloud (VPC), log in to http
 == Create a BYOC cluster
 
 . On the Clusters page, click *Create cluster*, then click *Create BYOC cluster*.
-. Enter a cluster name, then select the namespace, provider (AWS), xref:deploy:deployment-option/cloud/regions.adoc[region], availability zone type (single AZ or multi AZ) and zones, xref:deploy:deployment-option/cloud/cloud-overview.adoc#cluster-tiers[tier], and Redpanda version.
+. Enter a cluster name, then select the namespace, provider (AWS), xref:deploy:deployment-option/cloud/byoc-tiers.adoc[region, availability zones, tier], and Redpanda version.
```
towfiqa: Namespace should be changed to "resource group"

modules/deploy/pages/deployment-option/cloud/create-byoc-cluster-gcp.adoc
```diff
@@ -11,7 +11,7 @@ NOTE: With standard BYOC clusters, Redpanda manages security policies and resour
 
 . On the Clusters page, click *Create cluster*, then click *Create BYOC cluster*.
 +
-Enter a cluster name, then select the namespace, provider (GCP), xref:deploy:deployment-option/cloud/regions.adoc[region], availability zone type (single AZ or multi AZ) and zones, xref:deploy:deployment-option/cloud/cloud-overview.adoc#cluster-tiers[tier], and Redpanda version.
+Enter a cluster name, then select the namespace, provider (GCP), xref:deploy:deployment-option/cloud/byoc-tiers.adoc[region, availability zones, tier], and Redpanda version.
```
towfiqa: Namespace should be renamed to resource groups

modules/deploy/pages/deployment-option/cloud/create-dedicated-cloud-cluster-aws.adoc
```diff
@@ -9,7 +9,7 @@ After you log in to https://cloud.redpanda.com[Redpanda Cloud^], you land on the
 
 . On the Clusters page, click *Create cluster*, then click *Create Dedicated cluster*.
 +
-Enter a cluster name, then select the namespace, cloud provider (AWS or GCP), xref:deploy:deployment-option/cloud/regions.adoc[region], availability zone type (single AZ or multi AZ) and zones, xref:deploy:deployment-option/cloud/cloud-overview.adoc#cluster-tiers[tier], and Redpanda version.
+Enter a cluster name, then select the namespace, cloud provider (AWS or GCP), xref:deploy:deployment-option/cloud/dedicated/dedicated-tiers.adoc[region, availability zones, tier], and Redpanda version.
```
towfiqa: Namespace should be renamed resource group

modules/deploy/pages/deployment-option/cloud/dedicated/dedicated-tiers.adoc
```diff
@@ -0,0 +1,9 @@
+= Throughput Tiers
+:description: Select a throughput tier for your Redpanda Cloud cluster.
+:page-context-links: [{"name": "BYOC", "to": "deploy:deployment-option/cloud/byoc-tiers.adoc" },{"name": "Dedicated", "to": "deploy:deployment-option/cloud/dedicated/dedicated-tiers.adoc" } ]
+:page-cloud: true
+:env-dedicated: true
+
+NOTE: Serverless clusters are not on tiers. See xref:deploy:deployment-option/cloud/serverless.adoc#limits[Serverless limits]. To change your cluster type, contact https://support.redpanda.com/hc/en-us/requests/new[support^].
```
towfiqa: RE-word "not in tiers"

modules/deploy/partials/cloud/tiers.adoc
towfiqa, line 1: This page is called "Throughput Tiers". No explanation is given for what a throughput tier. But the first thing on the page tells us Serverless has no tiers. Then we jump immediately into Tier limits. And to finish things off we tell them Legacy tiers have different limits and they should contact Support. 

This page needs to be re-worked entirely. 
First start by removing that note about Serverless Clusters. Having this be the first thing on the page is very confusing. And also we repeat it later on the page. Which is also confusing because the page is for BYOC. Why are we talking about Serverless? 

Then start with what a throughput tier is and why a customer cares about this. This is THE most important decision a customer will make when creating a cluster. They should understand why and what that entails. Then talk about the characteristics of those tiers. I know we internally call these "limits" but for customers we should find better terminology. Limits has a negative connotation and implies something we enforce, which we do not. I will defer to @bpraseed on what that terminology should be or if we should continue using limits.

We should then remove the reference to legacy tiers at the top of the page. Customers don't have an easy way to know if they have legacy or new tiers. We can put this as a "Note" at the bottom of the page or remove the reference entirely. Having this at the top of the page feels alarming to me.

Then for the supported regions, is there a reason we are specifying "tier-1-aws-v2-arm"? If this is intended for the API folks, this should be in the API reference. Not here. The tiers should appear closer to what customers see in the UI today. Can we make this table a drop down type of information? The long scrolling page is not easy to use. And add AWS and GCP as a select. I missed the "tabs" the first time I tried this.

modules/deploy/pages/deployment-option/cloud/cloud-overview.adoc
```diff
@@ -40,10 +39,7 @@ To start using Dedicated, contact https://redpanda.com/try-redpanda?section=ente
...
-contained in your own environment. This provides an additional layer of security
-and isolation. Redpanda handles provisioning, operations, and maintenance. See <<BYOC architecture>>.
-
-Available on AWS or GCP. 
+contained in your own environment. This provides an additional layer of security and isolation. When you create a BYOC cluster, you select a xref:deploy:deployment-option/cloud/byoc-tiers.adoc[throughput tier and region] on either AWS or GCP. Redpanda handles provisioning, operations, and maintenance. See also: <<BYOC architecture>>.
 
 ==== Sign up for BYOC
```
towfiqa, line 44: Is this info needed for an overview page? It seems a little crammed in there.

towfiqa (dismissed):
Praseed should still review the accuracy of the region-AZs-Tier information

modules/deploy/pages/deployment-option/cloud/cloud-overview.adoc
micheleRP: Changed to "...you host your data on Redpanda cloud resources (AWS or GCP)..."

bpraseed:
Let's drop the zones from the region table as its a function of availability by the cloud provider. 
"Available throughput tiers" - Could we explain this better ? eg: This is needed for API/Terraform etc. 
@towfiqa - is there way to automatically map say Tier-1 to tier-1-aws-v2-arm ? Our goal has been to not leaked infrastructure details for Dedicated.

modules/deploy/pages/deployment-option/cloud/byoc-tiers.adoc
micheleRP: removed this, since the pages are specific to BYOC or Dedicated

modules/deploy/pages/deployment-option/cloud/cloud-overview.adoc
micheleRP: reworded: "With BYOC clusters, you deploy Redpanda in your own cloud (AWS or GCP),..."

modules/deploy/partials/cloud/tiers.adoc
micheleRP: @towfiqa and @bpraseed: I've removed the zone/machine type details from the main docs. After the API docs are published, we can add links to them here. I'm not sure what else to say about our tiers. Can you please review this and share any guidance about how they should decide which tier to pick, or what happens if they want/need to switch tiers?

modules/deploy/pages/deployment-option/cloud/byoc-tiers.adoc
hcoyote: tiers are more than just throughput. I'd almost drop the the word from the framing and describe that Tiers are shaped according to workload characteristics, including throughput requirements, partition sizing, etc.

modules/deploy/pages/deployment-option/cloud/cloud-overview.adoc
hcoyote: We need a comparison somewhere between the three different cloud managed deployment models. 

I think it's reasonable here.

modules/deploy/partials/cloud/tiers.adoc
hcoyote: noting here as well: don't call these throughput tiers because that causes confusion with customers who we have to size up along a different workload dimension. Maybe call it "Workload Tiers" or something like that.

modules/deploy/pages/deployment-option/cloud/byoc-tiers.adoc
micheleRP: Changed to: "When you create a BYOC cluster, you select your usage tier. Each tier provides tested and guaranteed workload configurations for throughput, logical partitions, and connections. The following table lists the supported tiers for BYOC clusters." (similar to Dedicated)

micheleRP:
Merging so link is available in API docs. For any additional updates, we'll open a new PR.

PR #476, opened 2024-05-03
docs: update public api docs to v1beta2

sago2k8:
Update references to the public api v1beta1 in the gcp private service connect and aws private link documentation.

ref https://github.com/redpanda-data/documentation-private/issues/2415


micheleRP (approved):
Thank you, @sago2k8!

PR #479, opened 2024-05-06
add internal property warning

Deflaimun:
Resolves https://github.com/redpanda-data/documentation-private/issues/2373
Fixes: DOC-198

### DIFF
--- a/modules/manage/partials/internal-use-property.adoc
+++ b/modules/manage/partials/internal-use-property.adoc
@@ -1,1 +1,1 @@
-WARNING: This property is for Redpanda internal use only. Do not modify it unless specifically instructed by Redpanda Support.
+WARNING: This property is for Redpanda internal use only. Do not modify it unless specifically instructed by https://support.redpanda.com/hc/en-us[Redpanda Support^].
--- a/modules/manage/partials/internal-use-property.adoc
+++ b/modules/manage/partials/internal-use-property.adoc
@@ -1,1 +1,1 @@
-WARNING: This property is for Redpanda internal use only. Do not modify it unless specifically instructed by https://support.redpanda.com/hc/en-us[Redpanda Support^].
+WARNING: This property is for Redpanda internal use only. Do not use or modify this property unless specifically instructed to do so by [https://support.redpanda.com/hc/en-us[Redpanda](https://support.redpanda.com/hc/en-us%5BRedpanda) Support^]. Using this property without explicit guidance from Redpanda Support could result in data loss.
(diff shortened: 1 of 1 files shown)


modules/manage/partials/internal-use-property.adoc
```diff
@@ -0,0 +1 @@
+WARNING: This property is exclusive for internal use. Please do not modify it unless specifically instructed by Redpanda Support.
```
kbatuigas: I'd phrase it along the lines of:
`WARNING: This property is for Redpanda internal use only. Do not modify it unless specifically instructed by Redpanda Support.`

Deflaimun:
```suggestion
WARNING: This property is for Redpanda internal use only. Do not modify it unless specifically instructed by Redpanda Support.
```

modules/manage/partials/internal-use-property.adoc
```diff
@@ -0,0 +1 @@
+WARNING: This property is for Redpanda internal use only. Do not modify it unless specifically instructed by Redpanda Support.
```
wzzzrd86: Can we link to [https://support.redpanda.com/hc/en-us](https://support.redpanda.com/hc/en-us)

Deflaimun: Yes! Sounds good to me

Deflaimun:
```suggestion
WARNING: This property is for Redpanda internal use only. Do not modify it unless specifically instructed by https://support.redpanda.com/hc/en-us[Redpanda Support^].
```

modules/manage/partials/internal-use-property.adoc
```diff
@@ -0,0 +1 @@
+WARNING: This property is for Redpanda internal use only. Do not modify it unless specifically instructed by https://support.redpanda.com/hc/en-us[Redpanda Support^].
```
Feediver1: So does this appear once at the top of the topic? If so, then:

WARNING: _This property is for Redpanda internal use only_. Do not use or modify any property here unless specifically instructed to do so by https://support.redpanda.com/hc/en-us[Redpanda Support^]. Using any property here without explicit guidance from Redpanda support could result in data loss.

Deflaimun: No. The idea here is to inject this admonition into every property that is meant for internal-only usage. 

Do you want to create a separate section for those properties?

Deflaimun: Going with individual admonitions for properties after a decision in a Zoom Call.

Deflaimun:
```suggestion
WARNING: This property is for Redpanda internal use only. Do not use or modify this property unless specifically instructed to do so by [https://support.redpanda.com/hc/en-us[Redpanda](https://support.redpanda.com/hc/en-us%5BRedpanda) Support^]. Using this property without explicit guidance from Redpanda Support could result in data loss.
```

Feediver1 (approved):

PR #486, opened 2024-05-08
Clarify when controller snapshots are enabled

Feediver1:
Fixes https://github.com/redpanda-data/documentation-private/issues/2189

Preview:
[Controller partition and snapshots](https://deploy-preview-486--redpanda-docs-preview.netlify.app/current/get-started/architecture/#controller-partition-and-snapshots)

[Upgrades - Limitations](https://deploy-preview-486--redpanda-docs-preview.netlify.app/current/upgrade/rolling-upgrade/#limitations)

### DIFF
--- a/modules/get-started/pages/architecture.adoc
+++ b/modules/get-started/pages/architecture.adoc
@@ -47,7 +47,7 @@ If a follower triggers an election, but the incumbent leader subsequently spring
 
 Redpanda stores metadata update commands (such as creating and deleting topics or users) in a system partition called the controller partition. A new snapshot is created after each controller command is added, or, with rapid updates, after a set period of time (default is 60 seconds). Controller snapshots save the current cluster metadata state to disk, so startup is fast. For example, with a partition that has moved several times, a snapshot can restore the latest state without replaying every move command.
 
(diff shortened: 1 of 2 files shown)


Feediver1:
@ztlpn Thinking that now, the entire **Controller snapshots** entry in Upgrade Limitations should be removed, no?

ztlpn:
> Thinking that now, the entire Controller snapshots entry in Upgrade Limitations should be removed, no?

@Feediver1 sounds good, it is only relevant for the 23.2 branch. Given that the docs are now versioned, it should just be removed in 24.1 and 23.3 branches.

Feediver1:
@ztlpn I removed the limitation. We do not typically go back into older versions and add updates. So I think this is good to go.

micheleRP:
@ztlpn: Is the correct term "controller partition" or "controller broker"? (See the current glossary entry for [controller broker](https://deploy-preview-486--redpanda-docs-preview.netlify.app/current/reference/glossary/#controller-broker).)

ztlpn (approved):

ztlpn:
> Is the correct term "controller partition" or "controller broker"?

Both are valid terms. Controller partition is the partition present on all brokers that stores cluster metadata. Controller broker is the leader of this partition (one broker in the cluster).

Btw, I wonder where this came from:
> Controller responsibilities: Elect new partition leaders when a broker fails or is offline

This is true for kafka I think, but definitely not true for redpanda.

micheleRP (approved):
lgtm

micheleRP:
> > Is the correct term "controller partition" or "controller broker"?
> 
> Both are valid terms. Controller partition is the partition present on all brokers that stores cluster metadata. Controller broker is the leader of this partition (one broker in the cluster).
> 
> Btw, I wonder where this came from:
> 
> > Controller responsibilities: Elect new partition leaders when a broker fails or is offline
> 
> This is true for kafka I think, but definitely not true for redpanda.

Thanks @ztlpn! We'll update the glossterm in a separate PR!

PR #488, opened 2024-05-13
Replace deprecated `license_key` and `license_secret_ref` with `enterprise`

JakeSCahill:
See https://redpandadata.slack.com/archives/C01H6JRQX1S/p1715631770730369

These fields were deprecated in favor of `enterprise` in https://github.com/redpanda-data/helm-charts/pull/723

### DIFF
--- a/modules/get-started/pages/licenses.adoc
+++ b/modules/get-started/pages/licenses.adoc
@@ -55,7 +55,7 @@ To apply the license key to your cluster, run:
 
 `rpk cluster license set`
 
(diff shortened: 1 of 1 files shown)


chrisseto (approved):
Thanks Jake!

This change was pretty old IIRC 🤔 We should probably come up with an automatic way of detecting and warning users about usages of deprecated fields and then we can run that against the docs just to be safe.

The JSON schema we generate has deprecation markers within it too. I wonder if there's a JSON schema linter we could just reuse.

Deflaimun (approved):

PR #496, opened 2024-05-15
Add server-side apply flag to kubectl apply

RafalKorepta:
## Description

Due to a repetition for Redpanda in version v1alpha1 and v1alpha2 the kubectl apply fails with

```
The CustomResourceDefinition "redpandas.cluster.redpanda.com" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```

With server-side flag the annotation is not added to each resource.

## Reference

https://github.com/kubernetes/kubectl/blob/514f46729f82412dd9cc41f206058bc4ae9b62b0/pkg/cmd/apply/apply.go#L762-L784 https://github.com/redpanda-data/redpanda-operator/commit/62a11afb2727a3ea551dab3127c1c05de77661f8

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


JakeSCahill (approved):
Thanks @RafalKorepta!

PR #502, opened 2024-05-17
Add `health_monitor_tick_interval` to deprecations

JakeSCahill:
## Description

This property was deprecated in 22.3 and backported to 22.2.

I also updated the order of deprecations to display the latest version at the top. @Deflaimun are the `rpk` deprecations automated? If so, we should update the script to display them in order and use Asciidoc syntax for the headings.

@BenPope do you remember the reason for the deprecation/if we have an alternative property? i couldn't see anything in the codebase.

Review deadline: 20 May

## Page previews

https://deploy-preview-502--redpanda-docs-preview.netlify.app/current/upgrade/deprecated/

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/upgrade/pages/deprecated/index.adoc
+++ b/modules/upgrade/pages/deprecated/index.adoc
@@ -24,6 +24,10 @@ This index helps you to identify deprecated features in Redpanda releases and pl
 | 23.2.1
 | Use the xref:23.2@upgrade:deprecated/cluster-resource.adoc[Redpanda resource] instead.
 
+| xref:reference:tunable-properties.adoc#health_monitor_tick_interval[`health_monitor_tick_interval`]
+| 22.2.1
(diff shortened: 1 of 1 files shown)


BenPope:
The PR cover describes the change: https://github.com/redpanda-data/redpanda/pull/5922

It just doesn't poll any more.

BenPope (approved):
I approve `health_monitor_tick_interval`, I don't know about the others.

micheleRP (approved):
lgtm

JakeSCahill:
@Deflaimun what should we do with this PR? We no longer have the tunable properties doc and don't seem to provide the docs for deprecated properties anymore. But the context here is useful for users.

Deflaimun:
Updated with the latest changes from the automation

PR #541, opened 2024-06-07
Add troubleshooting for failedScheduling errors in K8s

JakeSCahill:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2496
Review deadline: 7 June

## Page previews

[Preview](https://deploy-preview-541--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/self-hosted/kubernetes/local-guide/?tab=tabs-7-helm-operator#failedscheduling)

## Checks

- [ ] New feature
- [x] Content gap
- [x] Support Follow-up


modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
```diff
@@ -406,6 +406,82 @@ kubectl logs -f <pod-name> --namespace <namespace>
...
 //end::deployment-pod-pending[]
 
+//tag::deployment-failed-scheduling[]
+=== FailedScheduling
+
+If you see this error, your cluster does not have enough nodes to satisfy the anti-affinity rules:
+
+[.no-copy]
+----
+Warning  FailedScheduling  18m  default-scheduler  0/1 nodes are available: 1 node(s) didn't match pod anti-affinity rules. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod..
```
Deflaimun:
```suggestion
Warning  FailedScheduling  18m  default-scheduler  0/1 nodes are available: 1 node(s) didn't match pod anti-affinity rules. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod.
```

If this error message is extracted from the code, ignore the suggeston

Deflaimun:
This page is huge. Should we have a separate one dedicated for troubleshoot?

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
JakeSCahill: it is 😄

JakeSCahill:
> This page is huge. Should we have a separate one dedicated for troubleshoot?

We do have a separate one, which is even bigger. We use filters here to include only the troubleshooting for deployments. The idea is that users can do a Ctrl + F on that page to see the solution to their errors rather than clicking a link to the troubleshooting guide.

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
```diff
@@ -406,6 +406,82 @@ kubectl logs -f <pod-name> --namespace <namespace>
...
+apiVersion: cluster.redpanda.com/v1alpha1
+kind: Redpanda
+metadata:
+  name: redpanda
+spec:
+  chartRef: {}
+  clusterSpec:
+    statefulset:
+      podAntiAffinity:
+        type: soft
```
hcoyote, line 446: I set this to type=custom which worked for me (based off of a kapa suggestion). Is soft the preferable way?

hcoyote (approved):

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
JakeSCahill: Soft still allows scheduling the Pods onto separate workers but only if they’re available. The rule becomes [`preferredDuringSchedulingIgnoredDuringExecution`](https://github.com/redpanda-data/helm-charts/blob/5f287d45a3bda2763896840e505fb3de82b968b6/charts/redpanda/templates/_statefulset.tpl#L88)

JakeSCahill: @chrisseto or @RafalKorepta can confirm

RafalKorepta: Type `custom` is a way to provide your own pod anti affinity. 

https://github.com/redpanda-data/helm-charts/blob/5f287d45a3bda2763896840e505fb3de82b968b6/charts/redpanda/templates/_statefulset.tpl#L81-L100

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
```diff
@@ -406,6 +406,82 @@ kubectl logs -f <pod-name> --namespace <namespace>
...
+
+The Helm chart configures default `podAntiAffinity` rules to make sure that only one Pod running a Redpanda broker is scheduled on each worker node. To learn why, see xref:deploy:deployment-option/self-hosted/kubernetes/k-requirements.adoc#number-of-workers[Number of workers].
+
+To resolve this issue, do one of the following:
+
+- Create additional worker nodes.
+
+- Modify the anti-affinity rules (for development purposes only).
++
+If adding nodes is not an option, you can modify the `podAntiAffinity` rules in your StatefulSet to be less strict.
```
RafalKorepta, line 427: I feel like it might be a good place to reference Kuberentes docs with an anchor https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#types-of-inter-pod-affinity-and-anti-affinity or even the design https://github.com/kubernetes/design-proposals-archive/blob/main/scheduling/podaffinity.md of pod affinity and anti affinity.

RafalKorepta (approved):
No blockers. Only suggestion. LGTM

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
JakeSCahill, line 427:
```suggestion
If adding nodes is not an option, you can modify the `podAntiAffinity` rules in your StatefulSet to be less strict. For information about Pod affinity and anti-affinity types, see the https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#types-of-inter-pod-affinity-and-anti-affinity[Kubernetes documentation^].
```
```suggestion
If adding nodes is not an option, you can modify the `podAntiAffinity` rules in your StatefulSet to be less strict.
```

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
chrisseto: This error can also happen when using volumes that are bound to nodes (NVME for example) as we suggest doing or if your nodes are too small for redpanda's resource requests. (The exact error message may be a touch different. Though it should still be `FailedScheduling`.)

Nodes too small is simple enough to fix but the volume mount issue is much more complicated and is probably better handled by support for the immediate future. Would we want to call out that there are other ways this issue could be encountered so users don't end up at a "dead end" here?

chrisseto (approved):
LGTM but there are multiple paths that can lead to scheduling failures.

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
Feediver1: Does the warning really include two periods at the end? Regardless of how it appears in the actual warning, let's go with a single period at the end. :-)

Feediver1 (approved):
lgtm

modules/manage/pages/kubernetes/troubleshooting/k-troubleshoot.adoc
Deflaimun: @Feediver1 see https://github.com/redpanda-data/docs/pull/541#discussion_r1631386836

PR #548, opened 2024-06-11
remove auto-gen rpk pages

Deflaimun:
## Description

Remove the autogen pages from the `shared` branch in favor of `autogen` branch


Deflaimun:
since we have to support multiple concurrent versions, we still have to maintain separate versions of rpk. changing this PR to draft mode

PR #567, opened 2024-06-25
Add user guide for data transforms

JakeSCahill:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2242
Resolves https://github.com/redpanda-data/documentation-private/issues/2341
Resolves https://github.com/redpanda-data/documentation-private/issues/2572
Review deadline: 1 July

## Page previews

[Data transforms docs preview](https://deploy-preview-567--redpanda-docs-preview.netlify.app/current/develop/data-transforms/)

Related PRs: https://github.com/redpanda-data/docs-ui/pull/205 and https://github.com/redpanda-data/docs-extensions-and-macros/pull/56

## Checks

- [ ] New feature
- [x] Content gap
- [ ] Support Follow-up


modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+----
+
+The `transform.go` file contains an empty transform function.
+The `transform.yaml` file specifies the configuration settings for the transform function.
+
+See also: xref:develop:data-transforms/configure.adoc[].
+
+== Build transform functions
+
+You can develop your transform logic with one of the available xref:reference:data-transforms/sdks.adoc[software development kits (SDKs)] that allow you to run data transforms on a Redpanda cluster.
```
oleiman: nitpick(pedantic)

```suggestion
You can develop your transform logic with one of the available xref:reference:data-transforms/sdks.adoc[software development kits (SDKs)] that allow your transform code to interact with a Redpanda cluster.
```

Which is a little more strictly correct. I'm not sure it's clearer though.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+----
+.
+├── go.mod
+├── go.sum
+├── README.md
+├── transform.go
+└── transform.yaml
+----
+
+The `transform.go` file contains an empty transform function.
```
oleiman: Not quite true - I believe it's the identity transform. That is, a transform that takes records from the input topic and forwards them, unmodified, to each of the configured output topics.

I think it's a meaningful distinction because, after calling `transform init`, the resulting transform can be built, deployed, and observed without modification.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+The `transform.yaml` file specifies the configuration settings for the transform function.
+
+See also: xref:develop:data-transforms/configure.adoc[].
+
+== Build transform functions
+
+You can develop your transform logic with one of the available xref:reference:data-transforms/sdks.adoc[software development kits (SDKs)] that allow you to run data transforms on a Redpanda cluster.
+
+All transform functions must register a callback with the `OnRecordWritten()` method.
```
oleiman: > `OnRecordWritten()`

nitpick: This is specific to the golang SDK. The rust version will be similar but have slightly different name (`on_record_written` iirc), then the (unreleased) JavaScript version won't require registration at all, though we'll expect the callback to have a specific name.

I don't have a better suggestion offhand, just observing.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+)
+
+func main() {
+	transform.OnRecordWritten(handleRecord)
+}
+
+// Required callback, even if it does nothing
+func handleRecord(event transform.WriteEvent, writer transform.RecordWriter) error {
+	return nil
+}
```
oleiman: suggestion: consider populating these with the exact code emitted by `rpk`, as a hedge against confusion like "the transform function has a different name in my gen'ed code" or whatever.

```suggestion
package main

import (
        "github.com/redpanda-data/redpanda/src/transform-sdk/go/transform"
)

func main() {
        // Register your transform function.
        // This is a good place to perform other setup too.
        transform.OnRecordWritten(doTransform)
}

// doTransform is where you read the record that was written, and then you can
// output new records that will be written to the destination topic
func doTransform(e transform.WriteEvent, w transform.RecordWriter) error {
        return w.Write(e.Record())
}
```

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+use redpanda_transform_sdk::*;
+
+fn main() {
+  on_record_written(handle_record);
+}
+
+// Required callback, even if it does nothing
+fn handle_record(_event: WriteEvent, _writer: &mut RecordWriter) -> anyhow::Result<()> {
+  Ok(())
+}
```
oleiman:
```suggestion
use redpanda_transform_sdk::*;
use std::error::Error;

fn main() {
    // Register your transform function.
    // This is a good place to perform other setup too.
    on_record_written(my_transform);
}

// my_transform is where you read the record that was written, and then you can
// return new records that will be written to the output topic
fn my_transform(event: WriteEvent, writer: &mut RecordWriter) -> Result<(), Box<dyn Error>> {
    writer.write(event.record)?;
    Ok(())
}
```

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+----
+--
+======
+
+When creating a transform function, you should run any initialization steps in the `main()` function because it's only run once when the transform function is first deployed. If you're using Go, you can also use the standard predefined `init()` function.
```
oleiman: Oh, here it is. Feel free to disregard previous comment that suggests something similar

oleiman: >use the standard predefined `init()` function

Is there an example of this somewhere? I'm not a golang expert, so, offhand, I have no clue how this interacts with the whole `redpanda -> wasm engine -> wasm module -> main` flow.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+    // Log the error and continue processing
+    error!("Error: Record key is nil");
+    // Do not return the error
+    return Ok(());
+  }
+
+  if let Err(err) =  writer.write(record) {
+    // Log the error and continue processing
+    error!("Error writing record: {}", err);
+  }
```
oleiman: So this is kind of interesting. In this case, the nominal error is coming from _redpanda_ rather than client code, so I'd argue that it's probably better to let the error escape. 

Since there's no network hop between transform execution and the cluster, and the client code here is hosted entirely inside the broker, this is probably an issue with redpanda itself (or the VM, or whatever), so logging the error and committing progress is effectively data loss with respect to the output topic(s). 

Part of the issue here is that the broker hosting the transform doesn't have a way to distinguish between different types of errors, so we classify everything as fatal and restart the transform from the last committed offset.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+}
+
+fn handle_record(event: WriteEvent, writer: &mut RecordWriter) -> anyhow::Result<()> {
+  let record = event.record;
+  if record.key().is_none() {
+    // Log the error and continue processing
+    error!("Error: Record key is nil");
+    // Do not return the error
+    return Ok(());
+  }
```
oleiman: By contrast, this is an issue with the data, so we effectively discard the record and move on. Or maybe write to another output topic. 

In the future, emitting certain errors here might tell the transforms engine to send the record to a dead letter queue.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+[[errors]]
+=== Error handling
+
+To ensure the reliability and continuous operation of your data transforms, handle errors gracefully within all transform functions. Allowing errors to escape from a transform function can cause significant disruptions. Unhandled errors can cause your transform functions to fail, leading to data processing interruptions.
+
+Redpanda tracks the offsets of the records processed by transform functions. If an error escapes the Wasm virtual machine (VM), the VM will fail. When the Wasm engine detects this failure and starts a new VM, the transform function will try to process the input topics from the last processed offset again, leading to the same error and causing the function to be stuck in an error loop.
+
+To prevent this behavior, log errors and take appropriate action without stopping the processing of records. This approach ensures that your transform function can continue to operate even when encountering problematic records.
+
+In this example, an error will be logged if a record does not contain a key and that record will not be processed. Subsequent records will continue to be processed.
```
oleiman: I think this might be a bit heavy handed. You could get into a situation where there's some issue with a deployed transform that just prevents writing to the output topic(s) completely, in which case you'd keep discarding records in perpetuity with no indication aside from the transform logs.

Not sure of exact wording, but I think we want to stress the effect of allowing errors to escape (i.e. VM stops, restarts at last committed) vs eating errors (input topic advances, processing continues) without being overly prescriptive about usage.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+To ensure that you are notified of any errors or issues in your data transforms, Redpanda provides metrics that you can use to monitor the state of your data transforms.
+
+See also:
+
+- xref:develop:data-transforms/configure.adoc#log[Configure transform logging]
+- xref:develop:data-transforms/monitor.adoc[Monitor data transforms]
+
+=== Avoid state management
+
+Relying on state within transform functions can lead to inconsistencies and unpredictable behavior. Data transforms operate with at-least-once semantics, meaning transform functions might be executed more than once. Redpanda may also restart a transform function at any point, which causes its state to be lost.
```
oleiman:
```suggestion
Relying on in-memory state across transform invocations can lead to inconsistencies and unpredictable behavior. Data transforms operate with at-least-once semantics, meaning a transform function might be executed more than once for a given record. Redpanda may also restart a transform function at any point, which causes its state to be lost.
```

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,654 @@
...
+package main
+
+import (
+  "fmt"
+  "os"
+	"github.com/redpanda-data/redpanda/src/transform-sdk/go/transform"
+)
+
+func main() {
+	inputTopic, ok := os.LookupEnv("REDPANDA_INPUT_TOPIC")
```
oleiman: nitpick: whitespace

oleiman:
cc @rockwotj!

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,655 @@
...
+fn do_nothing(_event: WriteEvent, _writer: &mut RecordWriter) -> anyhow::Result<()> {
+  Ok(())
+}
+----
+--
+======
+
+=== Connect to the Schema Registry
+
+You can use the Schema Registry client library to read and write schemas as well as serialize and deserialize records. This client library is useful when working with schema-based topics in your data transforms.
```
oleiman, line 313: >as well as serialize and deserialize records

wording nitpick: this is true for the golang side, but the rust version requires a third party library for ser/deser.

modules/develop/pages/data-transforms/configure.adoc
```diff
@@ -0,0 +1,116 @@
...
+As well as custom environment variables set in either the <<cl, command-line>> or the <<config-file, configuration file>>, Redpanda makes some built-in environment variables available to your transform functions. These variables include:
+
+- `REDPANDA_INPUT_TOPIC`: The input topic specified.
+- `REDPANDA_OUTPUT_TOPIC_0..REDPANDA_OUTPUT_TOPIC_N`: The output topics in the order specified on the command line or in the configuration file. For example, `REDPANDA_OUTPUT_TOPIC_0` is the first variable, `REDPANDA_OUTPUT_TOPIC_1` is the second variable, and so on.
+
+Transform functions are isolated from the broker's internal environment variables to maintain security and encapsulation. Each transform function only uses the environment variables explicitly provided to it.
+
+NOTE: Built-in environment variables cannot be overwritten. Attempting to overwrite them in your custom environment variables results in an `Invalid transform environment` error during deployment.
+
+== Configure the Wasm engine
```
oleiman, line 56: general question/comment: should we include default values for configs?

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,655 @@
...
+    },
+    {
+      "key": "node",
+      "value": {
+        "intValue": 0
+      }
+    }
+  ]
+}
+----
```
rockwotj, line 202: This is a good opportunity to showcase `rpk transform logs`, because we don't tell people how to see this. Using the rpk command we can show people the command that generates this.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,655 @@
...
+--
+======
+
+=== Connect to the Schema Registry
+
+You can use the Schema Registry client library to read and write schemas as well as serialize and deserialize records. This client library is useful when working with schema-based topics in your data transforms.
+
+See also:
+
+- xref:manage:schema-reg/schema-reg-overview.adoc[].
```
rockwotj: Should we be consistent with `.`?

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,655 @@
...
+  "fmt"
+	"github.com/redpanda-data/redpanda/src/transform-sdk/go/transform"
+	"github.com/redpanda-data/redpanda/src/transform-sdk/go/transform/sr"
+)
+
+func main() {
+  client := sr.NewClient()
+	transform.OnRecordWritten(doNothing)
+
+  // Use the client...
```
rockwotj: You can't use the client here, `transform.OnRecordWritten` runs forever and never returns.

This should be passed into the callback, or registered as a global variable (same for Rust).

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -0,0 +1,95 @@
...
+[source,bash]
+----
+rpk transform deploy --file=https://my-site/my-transform.wasm
+----
+
+== Edit existing transform functions
+
+To make changes to an existing transform function:
+
+. xref:develop:data-transforms/build.adoc[Make you changes to the code].
```
oleiman:
```suggestion
. xref:develop:data-transforms/build.adoc[Make your changes to the code].
```

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -0,0 +1,95 @@
...
+
+== Troubleshoot
+
+This section provides guidance on how to diagnose and troubleshoot issues with building or deploying data transforms.
+
+=== Invalid transform environment
+
+This error means that one or more custom environment variables have the same name as the xref:develop:data-transforms/configure.adoc#built-in[built-in ones].
+
+Change the name of those custom environment variables to something unique.
```
oleiman: from the source code, I think the restriction is simply "environment variables must not begin with 'REDPANDA'"

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,655 @@
...
+
+// Required callback, even if it does nothing
+fn do_nothing(_event: WriteEvent, _writer: &mut RecordWriter) -> anyhow::Result<()> {
+  Ok(())
+}
+----
+--
+======
+
+==== Read schemas from the Schema Registry
```
rockwotj: This is all awesome, but we should show people how to interact with schemas when transforms are being executed as that's generally what people would want. Here's the pattern I expect in psuedo code:

```
client: sr_client

my_schema = "..."
my_schema_id = -1

main()
  client = create_sr_client()
  my_schema_id = write_schema(my_schema)
  on_record_written(my_transform)

my_transform(record, writer)
  schema_id, value = get_schema_id(record.value())
  record_schema = sr_client.get_schema(schema_id)
  struct = deserialize(value, my_schema, record_schema) # Assuming compatibility
  new_struct = do_transform(struct)
  new_value = serialize(new_struct, my_schema)
  new_value = add_schema_id(new_value, my_schema_id)
  writer.write(Record(record.key(), new_value))
```

we can certainly make this experience better over time.

modules/develop/pages/data-transforms/configure.adoc
```diff
@@ -0,0 +1,116 @@
...
+=== Built-In environment variables
+
+As well as custom environment variables set in either the <<cl, command-line>> or the <<config-file, configuration file>>, Redpanda makes some built-in environment variables available to your transform functions. These variables include:
+
+- `REDPANDA_INPUT_TOPIC`: The input topic specified.
+- `REDPANDA_OUTPUT_TOPIC_0..REDPANDA_OUTPUT_TOPIC_N`: The output topics in the order specified on the command line or in the configuration file. For example, `REDPANDA_OUTPUT_TOPIC_0` is the first variable, `REDPANDA_OUTPUT_TOPIC_1` is the second variable, and so on.
+
+Transform functions are isolated from the broker's internal environment variables to maintain security and encapsulation. Each transform function only uses the environment variables explicitly provided to it.
+
+NOTE: Built-in environment variables cannot be overwritten. Attempting to overwrite them in your custom environment variables results in an `Invalid transform environment` error during deployment.
```
rockwotj: We actually reserve anything that starts with `REDPANDA_` - Also there are size restrictions: https://github.com/redpanda-data/redpanda/blob/98e48acfb4dc7140328cc0c109a88f333f08a92f/src/v/cluster/plugin_frontend.cc#L282-L374

Key must be under 128 bytes, value under 2K total of 128 key value pairs, must be valid UTF8, no control characters (null bytes etc).

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -0,0 +1,95 @@
...
+- [ ] Do you have a Wasm binary located in the project directory?
+- [ ] Is the `transform.yaml` file correctly configured?
+- [ ] Does your transform function require any environment variables? If so, make sure to set them in the `transform.yaml` file or in the command-line when you deploy the binary.
+- [ ] Do your input and output topics already exist? Input and output topics must exist in your Redpanda cluster before you deploy the Wasm binary.
+
+. Deploy the Wasm binary:
++
+[source,bash]
+----
+rpk transform deploy --file=<path-to-wasm-binary> --var <environment-variable-name>=<value>
```
rockwotj: You don't need the `file` flag or the `var` flag if you're using transform.yaml

I recommend everyone use a `transform.yaml` file, but if you want to override a value on the CLI, you can do that (they get merged), or you can do all flags and drop the config file.

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -0,0 +1,95 @@
...
+rpk transform deploy --file=https://my-site/my-transform.wasm
+----
+
+== Edit existing transform functions
+
+To make changes to an existing transform function:
+
+. xref:develop:data-transforms/build.adoc[Make you changes to the code].
+. <<build, Rebuild>> the Wasm binary.
+. <<deploy, Redeploy>> the Wasm binary to the same Redpanda cluster.
```
rockwotj, line 106: Maybe useful to note that if you supplied values on the command line those can be overwritten the next deploy but they aren't "remembered" per se

modules/develop/pages/data-transforms/how-transforms-work.adoc
```diff
@@ -42,10 +44,18 @@ A new transform function reads the input topic from the latest offset. That is,
...
+If the previous instance of the transform function failed to commit its latest offsets before moving with the partition leader (for example, if the broker crashed), then it's likely that the new instance will reprocess some events. For broker failures, transform functions have at-least-once semantics, because records are retried from the committed last offset, and offsets are committed periodically. For more information, see xref:develop:data-transforms/run-transforms.adoc[].
+
+== Limitations
+
+- Transform functions have no external access to disk or network resources.
+- Only single record transforms is supported, but multiple output records from a single input record is supported. For aggregations, joins, or complex transformations, use Apache Flink.
```
oleiman:
```suggestion
- Only single record transforms is supported, but multiple output records from a single input record is supported. For aggregations, joins, or complex transformations, consider Redpanda Connect or Apache Flink.
```

😇

modules/develop/pages/data-transforms/test.adoc
```diff
@@ -0,0 +1,83 @@
+= Write Integration Tests for Transform Functions
```
rockwotj, line 1: This is awesome ❤️

modules/develop/pages/data-transforms/test.adoc
```diff
@@ -0,0 +1,83 @@
...
+go test
+```
+
+This will execute all tests in the current directory.
+
+== Integration tests
+
+Integration tests verify that your transform functions work correctly in a real Redpanda environment. You can use https://github.com/testcontainers/testcontainers-go/tree/main[testcontainers] to set up and manage a Redpanda instance for testing.
+
+For more detailed examples and helper code for setting up integration tests, refer to the SDK integration tests on https://github.com/redpanda-data/redpanda/tree/dev/src/transform-sdk/tests[GitHub].
```
rockwotj, line 83: We should work to moving this to a better example in our labs repo (no need to do anything right now). Mark shared some good stuff in community slack on this.

modules/develop/pages/data-transforms/monitor.adoc
```diff
@@ -0,0 +1,70 @@
+= Monitor Data Transforms
```
rockwotj, line 1: 👏

modules/develop/pages/data-transforms/test.adoc
oleiman, line 1: fyi: https://redpandacommunity.slack.com/archives/C05EAMK60QK/p1719567347109569

modules/develop/pages/data-transforms/versioning-compatibility.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+| 1.0.x       | 24.1.x, 23.3.x
+| 0.2.x       | 23.3.x
+| 0.1.x       | 23.3.x
+|===
+
+
+== Backward compatibility
+
+SDK versions will continue to work in the supported versions of Redpanda. However, they may not work in newer versions of Redpanda. Always check the compatibility matrix when upgrading either the SDK or Redpanda.
```
rockwotj: > However, they may not work in newer versions of Redpanda.

These two sentences to me counter each other.

Maybe something like:

SDK versions will have backwards compatibilty and continuted to be supported in newer versions of the broker. However, newer SDK versions may not work with older versions of Redpanda. Always check this compatibility matrix when upgrading the SDK or Redpanda.

modules/develop/pages/data-transforms/versioning-compatibility.adoc
```diff
@@ -0,0 +1,67 @@
...
+This compatibility matrix provides detailed information on which versions of the SDK are supported with which versions of Redpanda. Always consult the compatibility matrix to ensure that you are using compatible versions.
+
+=== Golang SDK
+
+|===
+| Golang SDK Version | Redpanda Version
+
+| 1.0.x       | 24.1.x, 23.3.x
+| 0.2.x       | 23.3.x
+| 0.1.x       | 23.3.x
```
oleiman: question (for product maybe): should we include pre-release versions here? it pads out the table, but strictly speaking we don't support anything pre-1.0.x.

cc @mattschumpert

modules/develop/pages/data-transforms/versioning-compatibility.adoc
oleiman: In general, directing users to the compatibility matrix is probably best, but I think the common case will be for new versions of redpanda to maintain compatibility with older SDK versions. Otherwise a transform might die across a broker upgrade?

Whereas new SDK versions may require a newer redpanda version to function.

oleiman (commented):
So awesome. Thanks for all this ❤️

modules/develop/pages/data-transforms/build.adoc
oleiman: worth noting that rust is not particularly great at "global instance of complicated (possibly) mutable thing". 🤷

modules/develop/pages/data-transforms/build.adoc
rockwotj: PS unless you snuck in a PR I missed JS will have registration too with `onRecordWritten`

oleiman: no, i am just javascript blind. I did not notice that the generated identity transform is passing a lambda into it.

modules/develop/pages/data-transforms/configure.adoc
JakeSCahill: > total of 128 key value pairs

Is this including the built-in ones?

modules/develop/pages/data-transforms/deploy.adoc
JakeSCahill: Given this, I'm going to add the limitations to a central place and link to it from here

https://github.com/redpanda-data/docs/pull/567#discussion_r1660391784

modules/develop/pages/data-transforms/build.adoc
JakeSCahill: I'll add docs for JS support after this PR is merged.

modules/develop/pages/data-transforms/configure.adoc
rockwotj: No just user defined

modules/develop/pages/data-transforms/build.adoc
JakeSCahill: It's strange that the boilerplate function names differ between Rust and Go. We should update that in `rpk`:

```go
package main

import (
	"github.com/redpanda-data/redpanda/src/transform-sdk/go/transform"
)

func main() {
        // Register your transform function. 
        // This is a good place to perform other setup too.
	transform.OnRecordWritten(doTransform)
}

// doTransform is where you read the record that was written, and then you can
// output new records that will be written to the destination topic
func doTransform(e transform.WriteEvent, w transform.RecordWriter) error {
	return w.Write(e.Record())
}
```

```rust
use redpanda_transform_sdk::*;

fn main() {
  // Register your transform function.
  // This is a good place to perform other setup too.
  on_record_written(my_transform);
}

// my_transform is where you read the record that was written, and then you can
// return new records that will be written to the output topic
fn my_transform(event: WriteEvent, writer: &mut RecordWriter) -> Result<(), Box<dyn Error>> {
  writer.write(event.record)?;
  Ok(())
}
```

JakeSCahill: I'd rather the arguments be written in full like in the Rust example too.

modules/develop/pages/data-transforms/build.adoc
JakeSCahill: Adding a link to https://golangdocs.com/init-function-in-golang

modules/develop/pages/data-transforms/configure.adoc
JakeSCahill: Those are documented in the linked reference docs.

modules/develop/pages/data-transforms/versioning-compatibility.adoc
JakeSCahill: Good point. I'll remove those prerelease versions.

modules/develop/pages/data-transforms/test.adoc
JakeSCahill: https://github.com/redpanda-data/documentation-private/issues/2568

modules/develop/pages/data-transforms/build.adoc
rockwotj: We should name them both my transform.

As for the argument naming, that is sort of a go-ism, but we can consult more gophers on the team and get their thoughts.

https://go.dev/wiki/CodeReviewComments#variable-names

modules/develop/pages/data-transforms/build.adoc
rockwotj: Let's link to something official: https://go.dev/doc/effective_go#init

modules/develop/pages/data-transforms/build.adoc
JakeSCahill: thanks! @rockwotj or @oleiman do you have a simple working example of this in Go and Rust?

For Rust, I don't know how to get a reference to the global SR client.

modules/develop/pages/data-transforms/build.adoc
JakeSCahill: TIL 😄 thanks!

JakeSCahill:
Thanks, both 😄 I've addressed all your feedback in https://github.com/redpanda-data/docs/pull/567/commits/fdc71a0f84b4d5db945becd3063a0fd38a2303cb

I'm unsure about the Schema Registry code examples. If you have suggestions for those, please could you provide the code?

modules/develop/pages/data-transforms/test.adoc
JakeSCahill: Added to https://github.com/redpanda-data/documentation-private/issues/2568

modules/develop/pages/data-transforms/build.adoc
oleiman: >working examples...?

Qualified yes:
- [golang](https://github.com/redpanda-data/redpanda/blob/dev/src/transform-sdk/go/transform/internal/testdata/schema-registry/transform.go) 
- [rust](https://github.com/redpanda-data/redpanda/blob/dev/src/transform-sdk/rust/examples/schema_registry/main.rs)

Qualified in the sense that these were written for integration testing and might not be suitable as docs-facing examples w/o some modification. Each exhibits a possible pattern for initializing and consuming an SR client though.

rockwotj:
I can write something up when I am back from OOO, but would not hold up all this great work before then.

modules/develop/pages/data-transforms/build.adoc
JakeSCahill: Let's leave the examples for now and before we release 24.2 we can update them to show how to use JSON schemas.

JakeSCahill: Since that's a new feature, we might want to showcase it here

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -65,46 +67,122 @@ import (
...
+// output new records that will be written to the destination topic
+func myTransform(event transform.WriteEvent, writer transform.RecordWriter) error {
+  return writer.Write(event.Record())
 }
 ----
 --
 Rust::
 +
 --
+All transform functions must register a callback with the `on_record_written()` method.
```
oleiman, line 84: 💯

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -65,46 +67,122 @@ import (
...
+By distinguishing between recoverable and critical errors, you can ensure that your transform functions are both resilient and robust. Handling recoverable errors internally helps maintain continuous operation, while allowing critical errors to escape ensures that the system can address severe issues effectively.
+
+Redpanda tracks the offsets of records that have been processed by transform functions. If an error escapes the Wasm virtual machine (VM), the VM will fail. When the Wasm engine detects this failure and starts a new VM, the transform function will retry processing the input topics from the last processed offset, potentially leading to repeated failures if the underlying issue is not resolved.
+
+Handling errors internally by logging them and continuing to process subsequent records can help maintain continuous operation. However, this approach can result in silently discarding problematic records, which may lead to unnoticed data loss if the logs are not monitored closely.
```
oleiman, line 115: 🔥

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -65,46 +67,122 @@ import (
...
+
+==== Allow errors to escape
+
+Allow errors to escape the Wasm VM only when they indicate a problem with the Redpanda infrastructure or the Wasm engine environment. Allowing the error to propagate ensures that the system can attempt to recover by restarting the VM and processing from the last committed offset.
```
oleiman: nitpick: I think you could _probably_ omit this in favor of the above paragraphs. But I'm not overly opinionated.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -65,46 +67,122 @@ import (
...
+
+func myTransform(event transform.WriteEvent, writer transform.RecordWriter) error {
+  err := writer.Write(event.Record())
+  if err != nil {
+    // Log the error and continue processing
+    log.Println("Error writing record:", err)
+    return err
+  }
+  // Return nil if no error occurred
+  return nil
```
oleiman: code/comment nitpick:

```suggestion
  if err != nil {
    log.Println("Error writing record:", err)
  }
  // If err is non-nil, it will fail the VM
  return err
```

modules/develop/pages/data-transforms/build.adoc
oleiman: ah, i see, you're introducing the example code. Feel free to disregard

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -121,25 +199,19 @@ import (
...
 
-  err := writer.Write(record)
-  if err != nil {
-    // Log the error and continue processing
-    log.Println("Error writing record:", err)
-  }
-  // Ensure no error is returned
-  return nil
+  return writer.Write(record)
 }
```
oleiman: idle thought: broadly this seems feels like a complete example, exhibiting both error-handling behaviors. If we detect a "bad" record, it can be discarded outright and processing continues. If writing a "good" record _fails_, we allow that error to propagate.

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -338,14 +407,14 @@ import (
 
 func main() {
   client := sr.NewClient()
```
oleiman: Think we want the client to be global here, like this:

https://github.com/redpanda-data/redpanda/blob/c7e27f720856b1787e4b26164863c3f7597acc1f/src/transform-sdk/go/transform/internal/testdata/schema-registry/transform.go#L26-L33

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -358,15 +427,15 @@ use redpanda_transform_sdk::*;
...
 
-    // Use the client
+  // Register your transform function.
+  on_record_written(my_transform);
 }
 
-// Required callback, even if it does nothing
-fn do_nothing(_event: WriteEvent, _writer: &mut RecordWriter) -> anyhow::Result<()> {
+fn my_transform(_event: WriteEvent, _writer: &mut RecordWriter) -> anyhow::Result<()> {
+  // Use the client
```
oleiman: >// Use the client

Can't do a straight global like go. I'm sure there's a subjectively cleaner way to accomplish this, but here's what we do for integration testing:

https://github.com/redpanda-data/redpanda/blob/c7e27f720856b1787e4b26164863c3f7597acc1f/src/transform-sdk/rust/examples/schema_registry/main.rs#L31-L47

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -391,27 +460,27 @@ package main
...
   "fmt"
-	"github.com/redpanda-data/redpanda/src/transform-sdk/go/transform"
-	"github.com/redpanda-data/redpanda/src/transform-sdk/go/transform/sr"
+  "github.com/redpanda-data/redpanda/src/transform-sdk/go/transform"
+  "github.com/redpanda-data/redpanda/src/transform-sdk/go/transform/sr"
 )
 
 func main() {
   // Initialize the Schema Registry client
   client := sr.NewClient()
```
oleiman: same comment. this scope doesn't extend to the transform function body.

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -65,19 +63,21 @@ rpk transform deploy --file=https://my-site/my-transform.wasm
 
 To make changes to an existing transform function:
 
-. xref:develop:data-transforms/build.adoc[Make you changes to the code].
+. xref:develop:data-transforms/build.adoc[Make your changes to the code].
 . <<build, Rebuild>> the Wasm binary.
 . <<deploy, Redeploy>> the Wasm binary to the same Redpanda cluster.
++
+If you provided configuration flags on the command line, those can be overwritten when you redeploy the Wasm binary.
```
oleiman: nitpick: more strict

```suggestion
Deploy-time configuration overrides must be provided each time you deploy a Wasm binary. Otherwise, they will be overwritten by default values or the configuration file's contents.
```

or some such. that's sort of clumsy.

oleiman (commented):
Looking really good. Most of my comments are around sample code.

I'm wondering whether we should hold off on the SR examples for now (put "examples coming soon" or something) until we can dial something in w/ Tyler. There's a bit of lang-specific mess to sort through, and it might be easier for everyone if we can standardize on something robust.

modules/develop/pages/data-transforms/build.adoc
JakeSCahill: Thanks! I've cleaned this part up to include this single complete example.

JakeSCahill:
Thanks @oleiman - in the interest of getting this merged today, I've removed the Schema Registry examples for now.

oleiman (approved):
lgtm 

:shipit:

modules/develop/pages/data-transforms/build.adoc
```diff
@@ -0,0 +1,330 @@
...
+You can view logs for transform functions using the `rpk transform logs <transform-function-name>` command.
+
+To ensure that you are notified of any errors or issues in your data transforms, Redpanda provides metrics that you can use to monitor the state of your data transforms.
+
+See also:
+
+- xref:develop:data-transforms/monitor#logs[View logs for transform functions]
+- xref:develop:data-transforms/monitor.adoc[Monitor data transforms]
+- xref:develop:data-transforms/configure.adoc#log[Configure transform logging]
+- xref:reference:rpk/rpk-transform/rp-transform-logs[]
```
kbatuigas:
```suggestion
- xref:reference:rpk/rpk-transform/rpk-transform-logs[]
```

modules/develop/pages/data-transforms/how-transforms-work.adoc
```diff
@@ -42,10 +44,31 @@ A new transform function reads the input topic from the latest offset. That is,
...
+
+This section outlines the limitations of data transforms. These constraints are categorized into general limitations affecting the overall functionality and specific limitations related to giving data transforms access to custom environment variables.
+
+=== General limitations
+
+- *No external access*: Transform functions have no external access to disk or network resources.
+- *Single record transforms*: Only single record transforms are supported, but multiple output records from a single input record are supported. For aggregations, joins, or complex transformations, consider xref:redpanda-connect:ROOT:about.adoc[Redpanda Connect] or https://flink.apache.org/[Apache Flink].
+- *Output topic limit*: Up to eight output topics are supported.
+- *Delivery semantics*: Transform functions have at-least-once delivery.
+- *Transaction API*: When clients use the Kafka Transactions API on partitions of an input topic, transform functions process only committed records.
```
kbatuigas:
```suggestion
- *Transactions API*: When clients use the Kafka Transactions API on partitions of an input topic, transform functions process only committed records.
```
nit

modules/develop/pages/data-transforms/how-transforms-work.adoc
```diff
@@ -42,10 +44,31 @@ A new transform function reads the input topic from the latest offset. That is,
...
+
+=== Environment variable limitations
+
+- *Maximum number of variables*: You can set up to 128 custom environment variables.
+- *Reserved prefix*: Variable keys must not start with `REDPANDA_`. This prefix is reserved for <<built-in,built-in environment variables>>.
```
kbatuigas:
```suggestion
- *Reserved prefix*: Variable keys must not start with `REDPANDA_`. This prefix is reserved for xref:develop:data-transforms/configure.adoc#built-in[built-in environment variables].
```
Moved to a new page?

modules/develop/pages/data-transforms/versioning-compatibility.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+=== Rust SDK
+
+|===
+| Rust SDK Version | Redpanda Version
+
+| 1.0.x       | 24.1.x, 23.3.x
+|===
+
+== Best Practices
```
kbatuigas:
```suggestion
== Best practices
```

kbatuigas (changes requested):
This all looks really good, bravo @JakeSCahill ! Just requesting a couple of minor changes.

The only other question I had was whether we have indeed decided to move away from publishing the reference docs on docs.redpanda.com. I know we've started to help review the reference content in the source PRs anyway, but curious to know if it was a product decision to publish the reference in just the package/crate docs (which to me implies that the docs team won't be maintaining it).

JakeSCahill:
Thanks for catching those @kbatuigas! I've committed all your feedback.

For the SDK reference, we agreed that since the SDK docs are autogenerated for Go and Rust, we should use those as the source of truth. It prevents APIs from becoming outdated in our docs. For example, the Rust SDK reference in our docs didn't include any Schema Registry info, but the autogenerated crate docs did. https://redpandadata.slack.com/archives/C01819S3ARH/p1719249225254979

JakeSCahill:
@oleiman @kbatuigas please could you take another quick look when you're back Friday? I changed the base to the beta branch since we need to get this one in: https://github.com/redpanda-data/documentation-private/issues/2572

https://github.com/redpanda-data/docs/pull/567/commits/1e1c197aeb2200e503d04da633a1d0bf640b53e1

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -0,0 +1,133 @@
...
+[source,bash]
+----
+rpk transform deploy --from-timestamp 1617181723
+----
+
+=== Share Wasm binaries
+
+You can also deploy data transforms on a Redpanda cluster by providing an addressable path to the Wasm binary. This is useful for sharing transform functions across multiple clusters or teams within your organization.
+
+For example, if the Wasm binary is hosted at `https://my-site/my-transform.wasm`, use the following command to deploy it:
```
kbatuigas: The URL renders as a clickable link, we could consider changing the formatting so that it's just a formatted string.

kbatuigas (approved):
🚀

modules/develop/pages/data-transforms/deploy.adoc
JakeSCahill:
```suggestion
For example, if the Wasm binary is hosted at `\https://my-site/my-transform.wasm`, use the following command to deploy it:
```

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -4,6 +4,7 @@
 
 {description}
 
+[[prerequisites]]
```
oleiman, line 7: This functionality is available v24.2+. Do we need to note that somewhere?

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -48,6 +47,43 @@ You can deploy your transform function using the xref:reference:rpk/rpk-transfor
...
+----
+rpk transform deploy --from-timestamp <unix-timestamp>
+----
+
+Example:
+
+[source,bash]
+----
+rpk transform deploy --from-timestamp 1617181723
+----
```
oleiman: This is not quite right. We have a single flag (`--from-offset`) and timestamp vs offset is determined by the format of the argument. This is somewhat similar to how `rpk topic consume` works, absent some bells and whistles.

Here's the help text:

```
The --from-offset flag can be used to specify where on the input topic the transform
should begin processing. Expressed as:

  - @T - Begin reading records with committed timestamp >= T (UNIX time, ms from epoch)
  - +N - Begin reading N records from the start of each input partition
  - -N - Begin reading N records prior to the end of each input partition

Note that the broker will only respect from-offset on the first deploy for a given
transform. Re-deploying the transform will cause processing to pick up at the last
committed offset. Recall that this state is maintained until the transform is deleted.
```

Wording could probably be improved somewhat, but that's an accurate description of how it works.

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -66,8 +102,10 @@ To make changes to an existing transform function:
 . xref:develop:data-transforms/build.adoc[Make your changes to the code].
 . <<build, Rebuild>> the Wasm binary.
 . <<deploy, Redeploy>> the Wasm binary to the same Redpanda cluster.
-+
-Deploy-time configuration overrides must be provided each time you deploy a Wasm binary. Otherwise, they will be overwritten by default values or the configuration file's contents.
+
+When you redeploy a Wasm binary with the same name, it will resume processing from the last offset it had previously processed. You can also configure the transform function to <<reprocess, reprocess some or all records>> before the last offset.
```
oleiman: `--from-offset` is only respected on initial deployment. if a user needs to redeploy a running transform and specify the start offset, they would need to delete the existing one and do a "clean" deploy.

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -48,6 +47,43 @@ You can deploy your transform function using the xref:reference:rpk/rpk-transfor
 rpk transform deploy
 ----
 
+When the transform function reaches Redpanda, it starts processing new records that are written to the input topic.
+
+[[reprocess]]
+=== Reprocess records
+
+In some cases, you may need to reprocess records from an input topic that already contains data. Processing existing records can be useful if, for example, the output topic was accidentally deleted, and you need to recreate the lost data by reprocessing those existing records.
```
oleiman, line 55: 🎯

modules/develop/pages/data-transforms/deploy.adoc
```diff
@@ -48,6 +47,43 @@ You can deploy your transform function using the xref:reference:rpk/rpk-transfor
...
+
+You can specify the exact point from which the transform function should start processing records in the input topic, starting from a specific offset or a specific timestamp. These options ensure that you can handle various scenarios where reprocessing is necessary to maintain data integrity and resilience.
```
oleiman: >specific offset

There is no option to specify an absolute offset. Records within a given _partition_ are totally ordered, but there's no absolute index for a record within a topic with multiple partitions. So `--from-offset` always specifies an offset delta. The most common usage is probably something like `--from-offset +0`, which means "start from the beginning of each partition".

Lmk if the help text in my other comment is unclear; happy to flesh it out a bit more.

oleiman: >specific timestamp

nitpick (pedantic): first record (per partition of the input topic) committed after the given timestamp. this one is definitely a nitpick.

oleiman:
@JakeSCahill looks good, but I think some wires got crossed with respect to the new rpk usage. lmk if I've misdocumented it somewhere, and I'll try to sync everything up.

oleiman (changes requested):
.

modules/develop/pages/data-transforms/deploy.adoc
JakeSCahill: This is going in beta docs only for now. So the version placeholder will automatically be replaced with the 24.2 RC version

JakeSCahill:
Thanks @oleiman ! I was going off the example in the [Jira](https://redpandadata.atlassian.net/browse/CORE-2841) 😅 It should be correct now.

PR #582, opened 2024-06-27
[Tiered Storage] Add rpk topic describe-storage to doc

kbatuigas:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/1775
Review deadline: **3 July 2024**

## Checks

- [ ] New feature
- [x] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/manage/partials/tiered-storage.adoc
+++ b/modules/manage/partials/tiered-storage.adoc
@@ -1068,7 +1068,7 @@ You can configure Redpanda to start a remote write periodically. This is useful 
 
 Setting idle timeout to a very short interval can create a lot of small files, which can affect throughput. If you decide to set a value for idle timeout, start with 600 seconds, which prevents the creation of so many small files that throughput is affected when you recover the files.
 
-Use the xref:reference:properties/object-storage-properties.adoc#cloud_storage_segment_max_upload_interval_sec[`cloud_storage_segment_max_upload_interval_sec`] property to set the number of seconds for idle timeout. If this property is empty, Redpanda uploads metadata to object storage, but the segment is not uploaded until it reaches the `segment.bytes` size. By default, `cloud_storage_segment_max_upload_interval_sec` is set to one hour (3600 seconds).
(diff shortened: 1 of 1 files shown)


modules/manage/partials/tiered-storage.adoc
```diff
@@ -1059,7 +1068,7 @@ You can configure Redpanda to start a remote write periodically. This is useful
 
 Setting idle timeout to a very short interval can create a lot of small files, which can affect throughput. If you decide to set a value for idle timeout, start with 600 seconds, which prevents the creation of so many small files that throughput is affected when you recover the files.
 
-Use the config_ref:cloud_storage_segment_max_upload_interval_sec,true,tunable-properties[] property to set the number of seconds for idle timeout. If this property is empty, Redpanda uploads metadata to object storage, but the segment is not uploaded until it reaches the `segment.bytes` size. By default, the property is null.
+Use the xref:reference:properties/object-storage-properties.adoc#cloud_storage_segment_max_upload_interval_sec[`cloud_storage_segment_max_upload_interval_sec`] property to set the number of seconds for idle timeout. If this property is empty, Redpanda uploads metadata to object storage, but the segment is not uploaded until it reaches the `segment.bytes` size. By default, `cloud_storage_segment_max_upload_interval_sec` is set to one hour (3600 seconds).
```
Deflaimun: Not sure if we should mention the default values here. They can change and we may forgot to update it back here. I would keep just a link to the property

kbatuigas: @Deflaimun Good call!

kbatuigas: @Deflaimun should we also change the phrasing around "If this property is empty..." (preceding sentence)? There is a default value, and I'm not sure if this property can ever be "empty" (I'm assuming that refers to nulled out)?

Deflaimun: @kbatuigas that's a good question. I don't know what happens to these if they're not present. I'm not 100% sure if not present = default value.

BenPope: In this case, the default is an hour, but it can be unset with:
```
rpk cluster config set cloud_storage_segment_max_upload_interval_sec null
```
Which means "Redpanda uploads metadata to object storage, but the segment is not uploaded until it reaches the `segment.bytes` size."

Lazin (approved):

Deflaimun (approved):

PR #583, opened 2024-06-27
[Tiered Storage] Clarify behavior for topics that already contain data

kbatuigas:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2282
Review deadline: **3 July 2024**

## Page previews

[Tiered Storage > Enable Tiered Storage](https://deploy-preview-583--redpanda-docs-preview.netlify.app/current/manage/tiered-storage/#enable-tiered-storage)

## Checks

- [ ] New feature
- [x] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/manage/partials/tiered-storage.adoc
+++ b/modules/manage/partials/tiered-storage.adoc
@@ -843,6 +843,8 @@ For additional properties, see <<Tiered Storage configuration properties>>.
 * <<Enable Tiered Storage for a cluster>>
 * <<Enable Tiered Storage for specific topics>>
 
+When Tiered Storage is enabled on a topic already containing data, it will proceed to upload any existing data in that topic on local storage to the object store bucket. It will start from the earliest offset available on local disk.
+
 ==== Enable Tiered Storage for a cluster
 
 To enable Tiered Storage for a cluster (in addition to setting `cloud_storage_enabled` to `true`), set the following cluster-level properties to `true`:
--- a/modules/manage/partials/tiered-storage.adoc
+++ b/modules/manage/partials/tiered-storage.adoc
@@ -843,14 +843,14 @@ For additional properties, see <<Tiered Storage configuration properties>>.
 * <<Enable Tiered Storage for a cluster>>
 * <<Enable Tiered Storage for specific topics>>
 
-When Tiered Storage is enabled on a topic already containing data, it will proceed to upload any existing data in that topic on local storage to the object store bucket. It will start from the earliest offset available on local disk.
+When you enable Tiered Storage on a topic already containing data, Redpanda uploads any existing data in that topic on local storage to the object store bucket. It will start from the earliest offset available on local disk. It is strongly recommended to avoid repeatedly toggling remote write on and off, because this can result in inconsistent data and data gaps in Tiered Storage for a topic.
 
 ==== Enable Tiered Storage for a cluster
 
 To enable Tiered Storage for a cluster (in addition to setting `cloud_storage_enabled` to `true`), set the following cluster-level properties to `true`:
 
-* config_ref:cloud_storage_enable_remote_write,true,tunable-properties[]
-* config_ref:cloud_storage_enable_remote_read,true,tunable-properties[]
+* config_ref:cloud_storage_enable_remote_write,true,properties/object-storage-properties[]
+* config_ref:cloud_storage_enable_remote_read,true,properties/object-storage-properties[]
 
 When you enable Tiered Storage for a cluster, you enable it for all existing topics in the cluster. When cluster-level properties are changed, the changes apply only to new topics, not existing topics.
 
@@ -1029,9 +1029,9 @@ Third-party tools that query space utilization from the Redpanda cluster might n
 
 Remote write is the process that constantly uploads log segments to object storage. The process is created for each partition and runs on the leader broker of the partition. It only uploads the segments that contain offsets that are smaller than the last stable offset. This is the latest offset that the client can read.
 
-To ensure all data is uploaded, you must enable remote write before any data is produced to the topic. If remote write is not enabled, data may be deleted due to retention settings.
+To ensure all data is uploaded, you must enable remote write before any data is produced to the topic. If you enable remote write after data has been written to the topic, only the data that currently exists on disk based on local retention settings will be scheduled for uploading. It is also strongly recommended to avoid repeatedly toggling remote write on and off, because this can result in inconsistent data and data gaps in Tiered Storage for a topic.
 
-To enable Tiered Storage, use remote write and remote read.
+To enable Tiered Storage, use both remote write and remote read.
 
 NOTE: If you only enable remote write on a topic, you have a simple backup that you can use for recovery. See xref:{data-archiving-link}[Data Archiving].
 
--- a/modules/manage/partials/tiered-storage.adoc
+++ b/modules/manage/partials/tiered-storage.adoc
@@ -843,7 +843,7 @@ For additional properties, see <<Tiered Storage configuration properties>>.
 * <<Enable Tiered Storage for a cluster>>
 * <<Enable Tiered Storage for specific topics>>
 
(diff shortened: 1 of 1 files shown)


modules/manage/partials/tiered-storage.adoc
```diff
@@ -843,6 +843,8 @@ For additional properties, see <<Tiered Storage configuration properties>>.
 * <<Enable Tiered Storage for a cluster>>
 * <<Enable Tiered Storage for specific topics>>
 
+When Tiered Storage is enabled on a topic already containing data, it will proceed to upload any existing data in that topic on local storage to the object store bucket. It will start from the earliest offset available on local disk.
```
Deflaimun: can this behavior be disabled? If yes, how? 
@hcoyote

kbatuigas:
Q: There is a [section](https://deploy-preview-583--redpanda-docs-preview.netlify.app/current/manage/tiered-storage/#remote-write) further down in the doc that says:

| To ensure all data is uploaded, you must enable remote write before any data is produced to the topic. If remote write is not enabled, data may be deleted due to retention settings.

Just wanted to confirm that this doesn't contradict with what we're adding to this doc in this PR -- I think what we're saying is, if you enable TS on a topic with data, any existing data is uploaded to object storage, but to _ensure_ that you don't lose any data, it's better to have enabled TS before there is any data in the topic.

modules/manage/partials/tiered-storage.adoc
hcoyote: Good question. didn't sound like the behavior could be disabled, but I didn't specifically ask about that in the original slack thread. @andrwng can you weigh in here?

hcoyote:
> Q: There is a [section](https://deploy-preview-583--redpanda-docs-preview.netlify.app/current/manage/tiered-storage/#remote-write) further down in the doc that says:
> 
> | To ensure all data is uploaded, you must enable remote write before any data is produced to the topic. If remote write is not enabled, data may be deleted due to retention settings.
> 
> Just wanted to confirm that this doesn't contradict with what we're adding to this doc in this PR -- I think what we're saying is, if you enable TS on a topic with data, any existing data is uploaded to object storage, but to _ensure_ that you don't lose any data, it's better to have enabled TS before there is any data in the topic.

I think the first sentence does contradict. I'm not sure the intent of the second sentence though. I read through the section context a few times and it's not clear to me.

Maybe something like this:

> To ensure all data is uploaded to Tiered Storage, you should enable remote write before any data is written to the topic. If you enable this after data has been written to the topic, only the data that currently exists on disk will be scheduled for uploading.

modules/manage/partials/tiered-storage.adoc
andrwng: This can't be disabled. Tiered storage today doesn't consider when in its lifespan TS has been enabled or disabled -- it only knows if it is enabled, it must upload everything it has that hasn't already been uploaded (based on understanding the highest uploaded so far, which in this case would be starting from 0).

FWIW toggling tiered storage on already active topics is not something we're generally comfortable advising. It leaves us open to weird behaviors, like gaps in our remote log. If we want to document this, can we caveat this by advising users to avoid toggling?

kbatuigas: @andrwng Does the phrasing you used work for this:
| When Tiered Storage is enabled on a topic already containing data, it will proceed to upload any existing data in that topic on local storage to the object store bucket. It will start from the earliest offset available on local disk. Enabling Tiered Storage on already active topics is not recommended as you may encounter gaps in the remote logs.
 
And would that plus @hcoyote 's suggestion in this [comment](https://github.com/redpanda-data/docs/pull/583#issuecomment-2195726797) be sufficient?
| To ensure all data is uploaded to Tiered Storage, you should enable remote write before any data is written to the topic. If you enable this after data has been written to the topic, only the data that currently exists on disk will be scheduled for uploading.

^ This would be added to the [Remote write](https://deploy-preview-583--redpanda-docs-preview.netlify.app/current/manage/tiered-storage/#remote-write) section.

hcoyote: Let's add an explicit note:

> Repeatedly toggling remote write on and off could lead to inconsistent data and data gaps in Tiered Storage for a topic. Redpanda discourages doing this.

andrwng: >Does the phrasing you used work for this:

SGTM!

>And would that plus @hcoyote 's suggestion in this https://github.com/redpanda-data/docs/pull/583#issuecomment-2195726797 be sufficient?

This is accurate. I'm not sure how detailed to get -- a natural follow-up question is "why wouldn't the data be on disk?" and to answer that admins would need to consider their local storage retention settings. Might be worth pointing to those too? OTOH given it's something we don't recommend anyway, not sure how best to convey that, if anything at all

modules/manage/partials/tiered-storage.adoc
```diff
@@ -843,14 +843,14 @@ For additional properties, see <<Tiered Storage configuration properties>>.
 * <<Enable Tiered Storage for a cluster>>
 * <<Enable Tiered Storage for specific topics>>
 
-When Tiered Storage is enabled on a topic already containing data, it will proceed to upload any existing data in that topic on local storage to the object store bucket. It will start from the earliest offset available on local disk.
+When you enable Tiered Storage on a topic already containing data, Redpanda uploads any existing data in that topic on local storage to the object store bucket. It will start from the earliest offset available on local disk. It is strongly recommended to avoid repeatedly toggling remote write on and off, because this can result in inconsistent data and data gaps in Tiered Storage for a topic.
```
andrwng: nit: maybe refer to it explicitly as `redpanda.remote.write`? Same below?

andrwng: Actually looks like "remote write" is pervasive throughout.

andrwng (approved):

modules/manage/partials/tiered-storage.adoc
```diff
@@ -843,6 +843,8 @@ For additional properties, see <<Tiered Storage configuration properties>>.
 * <<Enable Tiered Storage for a cluster>>
 * <<Enable Tiered Storage for specific topics>>
 
+When you enable Tiered Storage on a topic already containing data, Redpanda uploads any existing data in that topic on local storage to the object store bucket. It will start from the earliest offset available on local disk. It is strongly recommended to avoid repeatedly toggling remote write on and off, because this can result in inconsistent data and data gaps in Tiered Storage for a topic.
```
Feediver1:
```suggestion
When you enable Tiered Storage on a topic already containing data, Redpanda uploads any existing data in that topic on local storage to the object store bucket. It will start from the earliest offset available on local disk. Redpanda strongly recommends that you avoid repeatedly toggling remote write on and off, because this can result in inconsistent data and data gaps in Tiered Storage for a topic.
```

modules/manage/partials/tiered-storage.adoc
```diff
@@ -1027,9 +1029,9 @@ Third-party tools that query space utilization from the Redpanda cluster might n
 
 Remote write is the process that constantly uploads log segments to object storage. The process is created for each partition and runs on the leader broker of the partition. It only uploads the segments that contain offsets that are smaller than the last stable offset. This is the latest offset that the client can read.
 
-To ensure all data is uploaded, you must enable remote write before any data is produced to the topic. If remote write is not enabled, data may be deleted due to retention settings.
+To ensure all data is uploaded, you must enable remote write before any data is produced to the topic. If you enable remote write after data has been written to the topic, only the data that currently exists on disk based on local retention settings will be scheduled for uploading. It is also strongly recommended to avoid repeatedly toggling remote write on and off, because this can result in inconsistent data and data gaps in Tiered Storage for a topic.
```
Feediver1:
```suggestion
To ensure all data is uploaded, you must enable remote write before any data is produced to the topic. If you enable remote write after data has been written to the topic, only the data that currently exists on disk based on local retention settings will be scheduled for uploading. Redpanda strongly recommends that you avoid repeatedly toggling remote write on and off, because this can result in inconsistent data and data gaps in Tiered Storage for a topic.
```

Feediver1 (approved):

PR #587, opened 2024-07-01
Add description fetch strategy

Deflaimun:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2569
Review deadline: 4th July

## Page previews

https://deploy-preview-587--redpanda-docs-preview.netlify.app/24.2/reference/properties/cluster-properties/#fetch_read_strategy

## Checks

- [x] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/reference/pages/properties/cluster-properties.adoc
+++ b/modules/reference/pages/properties/cluster-properties.adoc
@@ -1202,7 +1202,7 @@ The strategy used to fulfill fetch requests.
 
 * `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. This should improve fetch latency and CPU utilization for most workloads.
 
-* `non_polling_with_debounce`: This option behaves like `non_polling` but includes a debounce mechanism with a fixed delay specified by <<fetch_reads_debounce_timeout,fetch_reads_debounce_timeout>> at the start of each fetch. This delay results in fewer, larger fetch responses, reducing overall reactor utilization and allowing other requests, such as produces, to increase. 
+* `non_polling_with_debounce`: This option behaves like non_polling but includes a debounce mechanism with a fixed delay specified by <<fetch_reads_debounce_timeout,fetch_reads_debounce_timeout>> at the start of each fetch. By introducing this delay, Redpanda can accumulate more data before processing, leading to fewer fetch operations and returning larger amounts of data. Enabling this option reduces reactor utilization but may also increase end-to-end latency.
 
 *Requires restart:* No
 
--- a/modules/reference/pages/properties/cluster-properties.adoc
+++ b/modules/reference/pages/properties/cluster-properties.adoc
@@ -1200,7 +1200,7 @@ The strategy used to fulfill fetch requests.
 
 * `polling`: Repeatedly polls every partition in the request for new data. The polling interval is set by <<fetch_reads_debounce_timeout,`fetch_reads_debounce_timeout`>> (deprecated).
 
-* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. Redpanda recommends setting this value for most workloads, as it can improve fetch latency and CPU utilization.
+* `non_polling`: The backend is signaled when a partition has new data, so Redpanda doesn’t need to repeatedly read from every partition in the fetch. Redpanda Data recommends using this value for most workloads, because it can improve fetch latency and CPU utilization.
 
 * `non_polling_with_debounce`: This option behaves like non_polling but includes a debounce mechanism with a fixed delay specified by <<fetch_reads_debounce_timeout,fetch_reads_debounce_timeout>> at the start of each fetch. By introducing this delay, Redpanda can accumulate more data before processing, leading to fewer fetch operations and returning larger amounts of data. Enabling this option reduces reactor utilization but may also increase end-to-end latency.
 
--- a/modules/reference/pages/properties/cluster-properties.adoc
+++ b/modules/reference/pages/properties/cluster-properties.adoc
@@ -1200,7 +1200,7 @@ The strategy used to fulfill fetch requests.
 
 * `polling`: Repeatedly polls every partition in the request for new data. The polling interval is set by <<fetch_reads_debounce_timeout,`fetch_reads_debounce_timeout`>>.
 
-* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. This should improve fetch latency and CPU utilization for most workloads.
+* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. Redpanda recommends setting this value for most workloads, as it can improve fetch latency and CPU utilization.
 
 * `non_polling_with_debounce`: This option behaves like non_polling but includes a debounce mechanism with a fixed delay specified by <<fetch_reads_debounce_timeout,fetch_reads_debounce_timeout>> at the start of each fetch. By introducing this delay, Redpanda can accumulate more data before processing, leading to fewer fetch operations and returning larger amounts of data. Enabling this option reduces reactor utilization but may also increase end-to-end latency.
 
--- a/modules/reference/pages/properties/cluster-properties.adoc
+++ b/modules/reference/pages/properties/cluster-properties.adoc
@@ -1198,13 +1198,19 @@ Maximum number of bytes returned in a fetch request.
 
 The strategy used to fulfill fetch requests.
 
+* `polling`: 
+
+* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. This should improve fetch latency and CPU utilization for most workloads.
+
+* `non_polling_with_debounce`: Allows for fetches to be delayed before they begin to work. This results in fewer/larger fetch responses, reducing overall reactor utilization, and allowing for other requests such as produces to increase. This option will add a fixed delay as specified by <<`fetch_reads_debounce_timeout`,fetch_reads_debounce_timeout>> to the start of every fetch.
+
 *Requires restart:* No
 
 *Optional:* Yes
 
 *Visibility:* `tunable`
 
-*Accepted Values:* `polling`, `non_polling`
+*Accepted Values:* `polling`, `non_polling`, `non_polling_with_debounce`
 
 *Default:* `non_polling`
 
--- a/modules/reference/pages/properties/cluster-properties.adoc
+++ b/modules/reference/pages/properties/cluster-properties.adoc
@@ -1202,7 +1202,7 @@ The strategy used to fulfill fetch requests.
 
 * `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. This should improve fetch latency and CPU utilization for most workloads.
 
-* `non_polling_with_debounce`: Allows for fetches to be delayed before they begin to work. This results in fewer/larger fetch responses, reducing overall reactor utilization, and allowing for other requests such as produces to increase. This option will add a fixed delay as specified by <<`fetch_reads_debounce_timeout`,fetch_reads_debounce_timeout>> to the start of every fetch.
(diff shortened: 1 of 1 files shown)


Deflaimun:
Need some help with the description for `polling`. Is there a defined poll interval? If yes, how to modify it? @ballard26

modules/reference/pages/properties/cluster-properties.adoc
```diff
@@ -1198,13 +1198,19 @@ Maximum number of bytes returned in a fetch request.
 
 The strategy used to fulfill fetch requests.
 
+* `polling`:
```
Feediver1: Why no description for this option?

Feediver1, line 1201: Why no description for this option?

modules/reference/pages/properties/cluster-properties.adoc
```diff
@@ -1198,13 +1198,19 @@ Maximum number of bytes returned in a fetch request.
 
 The strategy used to fulfill fetch requests.
 
+* `polling`: 
+
+* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. This should improve fetch latency and CPU utilization for most workloads.
+
+* `non_polling_with_debounce`: Allows for fetches to be delayed before they begin to work. This results in fewer/larger fetch responses, reducing overall reactor utilization, and allowing for other requests such as produces to increase. This option will add a fixed delay as specified by <<`fetch_reads_debounce_timeout`,fetch_reads_debounce_timeout>> to the start of every fetch.
```
Feediver1: looks to be a tagging issue here in the rendered output
image

Deflaimun: Fixed. Ty!

Feediver1:
@Deflaimun Eng has agreed to provide missing polling description/SME review by Friday.

modules/reference/pages/properties/cluster-properties.adoc
ballard26: I'd mark it as deprecated, then give it a description like;

```
Repeatedly polls every partition in the request for new data. The polling interval is set by <<fetch_reads_debounce_timeout,`fetch_reads_debounce_timeout`>
```

modules/reference/pages/properties/cluster-properties.adoc
```diff
@@ -1198,13 +1198,19 @@ Maximum number of bytes returned in a fetch request.
 
 The strategy used to fulfill fetch requests.
 
+* `polling`: 
+
+* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. This should improve fetch latency and CPU utilization for most workloads.
+
+* `non_polling_with_debounce`: Allows for fetches to be delayed before they begin to work. This results in fewer/larger fetch responses, reducing overall reactor utilization, and allowing for other requests such as produces to increase. This option will add a fixed delay as specified by <<fetch_reads_debounce_timeout,`fetch_reads_debounce_timeout`>> to the start of every fetch.
```
ballard26: Should we add a quick comment that it behaves the same as the `non_polling` option in addition to the extra functionality  allowing for delays?

modules/reference/pages/properties/cluster-properties.adoc
Feediver1: Added you as a reviewer @mattschumpert because this suggests we add this (`polling` option) to our list of deprecations. I know you mentioned a while back that we should alert you on any that we add.

modules/reference/pages/properties/broker-properties.adoc
```diff
@@ -137,6 +137,26 @@ TIP: For backward compatibility, `true` is the default. Redpanda recommends usin
 
 ---
 
+=== fips_mode
+
+Controls whether Redpanda starts in FIPS mode.  This property allows for three values: 
+
+* Disabled - Redpanda does not start in FIPS mode.
+
+* Permissive - Redpanda performs the same check as enabled, but a warning is logged, and Redpanda continues to run. Redpanda loads the OpenSSL FIPS provider into the OpenSSL library. After this completes, Redpanda is operating in FIPS mode, which means that the TLS cipher suites available to users are limited to the TLSv1.2 and TLSv1.3 NIST-approved cryptographic methods.
```
Feediver1: These options are also listed in https://docs.redpanda.com/beta/manage/security/fips-compliance/#configure-fips
One important piece of guidance is missing here:
permissive: This setting is a safety value option only, and _should not be used in a production environment_. If specified, Redpanda logs a WARNING, but continues operations even if the underlying operating system is not configured for FIPS. If set, your Redpanda instance is not running in FIPS compliance mode.

Feediver1: @michael-redpanda Can you please weigh in on these descriptions? I'm thinking they should be identical to what we have in https://docs.redpanda.com/beta/manage/security/fips-compliance/#configure-fips. AFAIK, the note in `permissive` is kind of critical, no?

Deflaimun: @Feediver1 this section is already out and should not be reviewed with this PR. Rebasing to fix the issue..

Deflaimun: I'll file a new PR to include your warning

modules/reference/pages/properties/cluster-properties.adoc
Deflaimun: Included your note and improved readability.

mattschumpert:
@ballard26 do we  really plan to actually remove the code for polling fetch and the config option completely in a future release?  If so, maybe it makes sense to comment its deprecated in the config listing/reference

 Otherwise, we're not removing the entire config property here (not deprecating the actual property) , so I would just think we'd highlight which is the 'default' and which is the 'recommended' value (which we should do regardless of deprecation). It's more important to indicate default/recommended choices in the reference.

I definitely don't think it's worth a top-level entry in our deprecated items page

cc 
@Deflaimun

modules/reference/pages/properties/cluster-properties.adoc
```diff
@@ -1198,13 +1198,19 @@ Maximum number of bytes returned in a fetch request.
 
 The strategy used to fulfill fetch requests.
 
+* `polling`: Repeatedly polls every partition in the request for new data. The polling interval is set by <<fetch_reads_debounce_timeout,`fetch_reads_debounce_timeout`>>.
+
+* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. This should improve fetch latency and CPU utilization for most workloads.
+
+* `non_polling_with_debounce`: This option behaves like `non_polling` but includes a debounce mechanism with a fixed delay specified by <<fetch_reads_debounce_timeout,fetch_reads_debounce_timeout>> at the start of each fetch. This delay results in fewer, larger fetch responses, reducing overall reactor utilization and allowing other requests, such as produces, to increase.
```
mattschumpert: What does 'allowing produce requests to increase' mean?     I think we mean it frees up resources for other requests. Consider rewording

mattschumpert (approved):

ballard26:
> @ballard26 do we really plan to actually remove the code for polling fetch and the config option completely in a future release? If so, maybe it makes sense to comment its deprecated in the config listing/reference

@mattschumpert Yeah, we will be removing the polling fetch code and config option entirely in a future release.

modules/reference/pages/properties/cluster-properties.adoc
Deflaimun:
```suggestion
* `non_polling_with_debounce`: This option behaves like non_polling but includes a debounce mechanism with a fixed delay specified by <<fetch_reads_debounce_timeout,fetch_reads_debounce_timeout>> at the start of each fetch. By introducing this delay, Redpanda can accumulate more data before processing, leading to fewer fetch operations and returning larger amounts of data. Enabling this option reduces reactor utilization but may also increase end-to-end latency.
```

modules/reference/pages/properties/cluster-properties.adoc
```diff
@@ -1198,13 +1198,19 @@ Maximum number of bytes returned in a fetch request.
 
 The strategy used to fulfill fetch requests.
 
+* `polling`: Repeatedly polls every partition in the request for new data. The polling interval is set by <<fetch_reads_debounce_timeout,`fetch_reads_debounce_timeout`>>.
+
+* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. This should improve fetch latency and CPU utilization for most workloads.
```
Deflaimun:
```suggestion
* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. Redpanda recommends setting this value for most workloads, as it can improve fetch latency and CPU utilization.
```

modules/reference/pages/properties/cluster-properties.adoc
Deflaimun: done

mattschumpert:
So we should make a comment it's deprecated here: https://deploy-preview-587--redpanda-docs-preview.netlify.app/24.2/reference/properties/cluster-properties/#fetch_read_strategy and in the cluster config reference then, and i guess in https://docs.redpanda.com/current/upgrade/deprecated/ if the entire config is indeed going away

Deflaimun:
The current policy is to add to those pages only when the feature is deprecated. Currently we don't add deprecation _notices_ . Since the feature is not deprecated right now, we can't add it there. 

What we can do is add (deprecated) at the end of the current config value.

modules/reference/pages/properties/cluster-properties.adoc
```diff
@@ -1198,13 +1198,19 @@ Maximum number of bytes returned in a fetch request.
 
 The strategy used to fulfill fetch requests.
 
+* `polling`: Repeatedly polls every partition in the request for new data. The polling interval is set by <<fetch_reads_debounce_timeout,`fetch_reads_debounce_timeout`>>.
```
Deflaimun:
```suggestion
* `polling`: Repeatedly polls every partition in the request for new data. The polling interval is set by <<fetch_reads_debounce_timeout,`fetch_reads_debounce_timeout`>> (deprecated).
```

modules/reference/pages/properties/cluster-properties.adoc
```diff
@@ -1198,13 +1198,19 @@ Maximum number of bytes returned in a fetch request.
 
 The strategy used to fulfill fetch requests.
 
+* `polling`: Repeatedly polls every partition in the request for new data. The polling interval is set by <<fetch_reads_debounce_timeout,`fetch_reads_debounce_timeout`>> (deprecated).
+
+* `non_polling`: The backend is signaled when a partition has new data, so it doesn’t need to repeatedly read from every partition in the fetch. Redpanda recommends setting this value for most workloads, as it can improve fetch latency and CPU utilization.
```
micheleRP:
```suggestion
* `non_polling`: The backend is signaled when a partition has new data, so Redpanda doesn’t need to repeatedly read from every partition in the fetch. Redpanda Data recommends using this value for most workloads, because it can improve fetch latency and CPU utilization.
```

modules/reference/pages/properties/cluster-properties.adoc
```diff
@@ -1198,13 +1198,19 @@ Maximum number of bytes returned in a fetch request.
...
+
+* `non_polling_with_debounce`: This option behaves like non_polling but includes a debounce mechanism with a fixed delay specified by <<fetch_reads_debounce_timeout,fetch_reads_debounce_timeout>> at the start of each fetch. By introducing this delay, Redpanda can accumulate more data before processing, leading to fewer fetch operations and returning larger amounts of data. Enabling this option reduces reactor utilization but may also increase end-to-end latency.
```
micheleRP:
```suggestion
* `non_polling_with_debounce`: This option behaves like `non_polling`, but it includes a debounce mechanism with a fixed delay specified by <<fetch_reads_debounce_timeout,fetch_reads_debounce_timeout>> at the start of each fetch. By introducing this delay, Redpanda can accumulate more data before processing, leading to fewer fetch operations and returning larger amounts of data. Enabling this option reduces reactor utilization, but it may also increase end-to-end latency.
```

micheleRP (approved):
please check the style suggestions, then approved!

PR #591, opened 2024-07-05
rpk 24.2

Deflaimun:
## Description

Updates rpk for 24.2
Review deadline: 9 July

### Changelog
New commands
[`rpk cluster quotas alter`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-cluster/rpk-cluster-quotas-alter/)
[`rpk cluster quotas describe`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-cluster/rpk-cluster-quotas-describe/)
[`rpk cluster quotas import`](https://deploy-preview-611--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-cluster/rpk-cluster-quotas-import/)
[`rpk cluster quotas`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-cluster/rpk-cluster-quotas/)
[`rpk transform pause`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-transform/rpk-transform-pause/)
[`rpk transform resume`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-transform/rpk-transform-resume/)

New flags or improved descriptions
[`rpk cluster partitions list`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-cluster/rpk-cluster-partitions-list/)
[`rpk cluster self-test start`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-cluster/rpk-cluster-self-test-start/)
[`rpk registry schema check combatibility`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-registry/rpk-registry-schema-check-compatibility/)
[`rpk registry schema create`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-registry/rpk-registry-schema-create/)
[`rpk registry schema get`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-registry/rpk-registry-schema-get/)
[`rpk transform deploy`](https://deploy-preview-591--redpanda-docs-preview.netlify.app/24.2/reference/rpk/rpk-transform/rpk-transform-deploy/)

## Checks

- [x] New feature
[...]

PR #589, opened 2024-07-04
Replace deprecated installCRDs option for cert-manager

JakeSCahill:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2571
Review deadline: 5 July

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up
- [x] Small fix (typos, links, copyedits, etc)

PR #592, opened 2024-07-08
Client throughput quota utilization

kbatuigas:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2422
Review deadline: 19 July 2024

* Describe added support for Kafka API based client throughput quota
* Include new metrics for monitoring client throughput

Tasks:

- [ ] new `rpk cluster quotas` commands (in PR https://github.com/redpanda-data/docs/pull/591)
- [ ] new metrics reference
- [ ] Update cluster config reference 
- [ ] Update https://docs.redpanda.com/current/develop/kafka-clients/#unsupported-kafka-features
- [ ] Upgrade guide
- [ ] What's New

## Page previews

[Manage Throughput](https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/manage/cluster-maintenance/manage-throughput/)
[Public Metrics Reference > Broker metrics](https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/reference/public-metrics-reference/#broker-metrics)
[Internal Metrics Reference](https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/reference/internal-metrics-reference/#vectorized_kafka_quotas_client_quota_throttle_time)
[Kafka Compatibility > Unsupported Kafka features](https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/develop/kafka-clients/#unsupported-kafka-features)
[Upgrade Redpanda > Prerequisites > Review incompatible changes](https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/upgrade/rolling-upgrade/#review-incompatible-changes)
[What's New > Client throughput management](https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/get-started/whats-new/#client-throughput-management)

## Checks

- [x] New feature
- [ ] Content gap
- [ ] Support Follow-up


modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,127 @@ The properties for broker-wide throughput quota balancing are configured at the
...
 == Client throughput limits
 
-Similar to <<broker-wide-throughput-limits>> but for clients, Redpanda provides configurable throughput quotas that apply to an individual client or a group of clients.
+Redpanda provides configurable throughput quotas that apply to an individual client or a group of clients. You can apply a quota for an individual client based on an exact match with its `client_id`, or a group of clients based on IDs that start with a given prefix. 
+
+As of v24.2, client throughput quotas are compatible with the https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client[Kafka API^] (AlterClientQuotas and DescribeClientQuotas), and are separate from quotas configured through cluster configuration in earlier Redpanda versions. The client throughput quotas no longer apply on a per-shard basis, and now limit the rates across a Redpanda broker's node. The quotas are neither shared nor balanced between brokers.
+
+Redpanda supports the following Kafka API-based quota types on clients:
+
+|===
```
kbatuigas, line 54: @pgellert Do the Kafka quota types correspond exactly with existing cluster config quota types and do we need to map those out here as well? @mattschumpert

modules/reference/pages/internal-metrics-reference.adoc
```diff
@@ -87,6 +87,22 @@ Current effective quota for the quota balancer, in bytes per second.
 
 ---
 
+=== vectorized_kafka_quotas_client_quota_throttle_time
```
kbatuigas, line 90: @pgellert Is this distinct from `vectorized_kafka_quotas_throttle_time` and can we provide a tip or further description to help distinguish between the two?

modules/reference/pages/public-metrics-reference.adoc
```diff
@@ -497,6 +497,22 @@ Latency of produce/consume requests per broker. This duration measures from when
...
+
+Client quota throttling delay, in seconds, per rule and quota type.
+
+*Type*: histogram
+
+---
+
+=== redpanda_kafka_quotas_client_quota_throughput
+
+Client quota throughput per rule and quota type.
```
kbatuigas, line 515: @pgellert This reference page also provides `*Labels*` info, can we provide it for the two new metrics?

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,127 @@ The properties for broker-wide throughput quota balancing are configured at the
...
-* `clients_prefix`: a client belonging to the group has this prefix in its `client_id` property.
-* `quota`: the maximum throughput rate of *each client in the group* in bytes per second.
+[,bash]
+----
+rpk cluster quotas alter --default client-id --add producer_byte_rate=1000000000
+----
 
-NOTE: For `kafka_client_group_fetch_byte_rate_quota`, a group of consumer clients is not a Kafka consumer group, and the `group_name` field is not a Kafka consumer group ID. Instead, it is the set of clients fetching from a Redpanda broker that are configured by the property to be throughput limited.
+[[migrate]]
+=== Migrate cluster configuration quotas to Kafka API-based quotas
```
kbatuigas, line 153: @pgellert Will ask Rogger to review rpk command syntax -- can you confirm if these are the correct migration steps? cc @mattschumpert

mattschumpert (commented):
Also need to represent the old cluster configs here: https://deploy-preview-592--redpanda-docs-preview.netlify.app/current/upgrade/deprecated/

And the new support (including RPK support and Console support) here: https://docs.redpanda.com/beta/get-started/whats-new/

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,134 @@ The properties for broker-wide throughput quota balancing are configured at the
...
 
-* The `target_quota_byte_rate` property applies to a producer client that isn't a member of a client group configured by xref:reference:cluster-properties.adoc#kafka_client_group_byte_rate_quota[`kafka_client_group_byte_rate_quota`]. It sets the maximum throughput quota of a client sending to a Redpanda broker.
-* The `target_fetch_quota_byte_rate` property applies to a consumer client that isn't a member of a client group configured by xref:reference:cluster-properties.adoc#kafka_client_group_fetch_byte_rate_quota[`kafka_client_group_fetch_byte_rate_quota`]. It sets the maximum throughput quota of a client fetching from a Redpanda broker.
+To set a throughput quota for a single client, use the `rpk cluster quotas alter` command. 
 
-The values of both `target_quota_byte_rate` and `target_fetch_quota_byte_rate` are throughput rate limits within a shard, in bytes per second.
+[,bash]
+----
+rpk cluster quotas alter --add consumer_byte_rate=200000 --name client-id-prefix=consumer-
+----
```
pgellert: This section is for quotas for an individual client, not a client prefix:
```suggestion
----
rpk cluster quotas alter --add consumer_byte_rate=200000 --name client-id=consumer-1
----
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,134 @@ The properties for broker-wide throughput quota balancing are configured at the
...
 
-To set throughput quota for a group of clients, configure the cluster properties xref:reference:cluster-properties.adoc#kafka_client_group_byte_rate_quota[`kafka_client_group_byte_rate_quota`] and xref:reference:cluster-properties.adoc#kafka_client_group_fetch_byte_rate_quota[`kafka_client_group_fetch_byte_rate_quota`].
+Alternatively, you can view or configure throughput quotas for a group of clients based on a match on client ID prefix. The following example sets the `consumer_byte_rate` quota to a group of clients that have IDs starting with `consumer-`:
+
+[,bash]
+----
+rpk cluster quotas alter --add consumer_byte_rate=200000 --name client-id-prefix=consumer-
+----
+
+NOTE: The client group that matches on `client-id-prefix` does not denote a Kafka consumer group. Instead, it is the set of clients fetching from a Redpanda broker that are configured by the property to be throughput limited.
```
pgellert: I think this is a very useful note. I am not following the last bit though so I suggested some edits to that.

For context, the client id is typically passed as an optional arugment to the method creating the kafka client. For reference, this is how franz-go does it: https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#ClientID.
```suggestion
NOTE: The client group that matches on `client-id-prefix` does not denote a Kafka consumer group. Instead, it matches all requests with a `client_id` field that has the configured prefix. The `client_id` field is typically a configurable parameter in kafka libraries when the client is constructed.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,134 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+=== Default client throughput limit
 
-* The `kafka_client_group_byte_rate_quota` property applies to producer clients. It sets a maximum throughput quota of traffic sent to Redpanda from each producer in the group named by the property.
-* The `kafka_client_group_fetch_byte_rate_quota` property applies to consumer clients. It sets a maximum throughput quota of traffic fetched from Redpanda by each consumer in the group named by the property.
+You can apply default throughput limits to clients. Redpanda applies the default limits if no quotas have been configured specific to a client ID or ID prefix. Defaults are only configurable based on the exact client ID.
```
pgellert: I think I know what you mean here, which is that to configure default quotas, you need to use the `--default client-id` argument and that `--default client-id-prefix` is not an option. But I think "Defaults are only configurable based on the exact client ID." is a bit misleading because defaults are not associated with an exact client ID so I think removing that sentence makes the whole section clearer.

```suggestion
You can apply default throughput limits to clients. Redpanda applies the default limits if no quotas have been configured specific to a client ID or ID prefix.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
pgellert: Yes, these are the steps necessary. A couple of things that I think would be useful to add here:
* I think we should call out the per-shard vs per-broker distinction here again. If let's say they had `target_fetch_quota_byte_rate` set to 100MBps/shard and they are running on 16 core brokers, they might want to set the new quotas to 1600MBps.
* I think a table would be useful here to show what cluster configs correspond to what rpk arguments:
```
target_quota_byte_rate                      | --default client-id --add producer_byte_rate=<value>
target_fetch_quota_byte_rate                | --default client-id --add consumer_byte_rate=<value>
kafka_admin_topic_api_rate                  | --default client-id --add controller_mutation_rate=<value>
kafka_client_group_byte_rate_quota          | --name client-id-prefix=<prefix> --add producer_byte_rate=<value>
kafka_client_group_fetch_byte_rate_quota    | --name client-id-prefix=<prefix> --add consumer_byte_rate=<value>
```
* `rpk cluster config get kafka_client_group_byte_rate_quota` would not return a number (`1000000000` like in the example). It has a more complex structure. So I would either replace the example with `rpk cluster config get target_fetch_quota_byte_rate`. Or, even better, change it such that `kafka_client_group_byte_rate_quota` returns something that's expected, eg:
```
% rpk cluster config get kafka_client_group_byte_rate_quota
"kafka_client_group_byte_rate_quota": [
  {
    "group_name": "group_1",
    "clients_prefix": "producer_group_alone_producer",
    "quota": 10240
  },
  {
    "group_name": "group_2",
    "clients_prefix": "producer_group_multiple",
    "quota": 20480
  },
]
% rpk cluster quotas alter --add producer_byte_rate=10240 --name client-id-prefix=producer_group_alone_producer
% rpk cluster quotas alter --add producer_byte_rate=20480 --name client-id-prefix=producer_group_multiple
```

modules/reference/pages/internal-metrics-reference.adoc
pgellert: Yeah, the naming is not amazing here. They are distinct.
* `vectorized_kafka_quotas_throttle_time` tracks throttle time based on [Broker-wide throughput limits](https://docs.redpanda.com/current/manage/cluster-maintenance/manage-throughput/#broker-wide-throughput-limits)
* `vectorized_kafka_quotas_client_quota_throttle_time` tracks throttle time based on [Client throughput limits](https://docs.redpanda.com/current/manage/cluster-maintenance/manage-throughput/#client-throughput-limits)

modules/reference/pages/public-metrics-reference.adoc
pgellert: Nice! Yes, these are the labels on both:
```
quota_rule=("not_applicable" | "kafka_client_default" | "cluster_client_default" | "kafka_client_prefix" | "cluster_client_prefix" | "kafka_client_id")
quota_type=("produce_quota" | "fetch_quota" | "partition_mutation_quota")
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
pgellert: Yes, I've added the mapping in this comment: https://github.com/redpanda-data/docs/pull/592#discussion_r1670312568. Feel free to move it wherever you think it makes the most sense.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,169 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+|===
++
+The client throughput quotas set through the Kafka API apply per broker, so you need to convert the cluster configuration values that were applied on a per-shard basis. For example, if you set `target_fetch_quota_byte_rate` to 100MBps/shard, and you run Redpanda on 16-core brokers, you can set the new consumer_byte_rate quota to 100 * 16 = 1600MBps.
+
+. Use `rpk cluster quotas alter` to set the corresponding client throughput quotas based on the Kafka API:
++
+[,bash]
+----
+rpk cluster quotas alter --name client-id-prefix=producer_group_alone_producer --add producer_byte_rate=10240
```
kbatuigas: @pgellert Is this also an example where the value could be converted from the per-shard setting to per-broker? Just checking because these are the same example values from the cluster config.

pgellert: All of the 5 cluster configs are converted (multiplied by the core count), but anything configured with `rpk cluster quotas alter` is not converted. So it's probably best for the example to also be scaled up to maintain the existing behaviour. Something like this:
```
rpk cluster quotas alter --name client-id-prefix=producer_group_alone_producer --add producer_byte_rate=$((10240*$BROKER_CORE_COUNT))
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,172 @@ The properties for broker-wide throughput quota balancing are configured at the
...
++
+Replace the placeholder values with the new quota values, accounting for the conversion to per-broker limits, for example 10240 * broker core count = new quota.
+
+. Use `rpk cluster config` to remove the configuration-based quotas:
++
+[,bash]
+----
+rpk cluster config
+----
+// TODO: rpk command to delete/unset quotas
```
pgellert: There might be an even nicer way to do it that I don't know about, but this is how I would reset the config:
```suggestion
----
rpk cluster config kafka_client_group_byte_rate_quota=
----
```

modules/deploy/pages/deployment-option/self-hosted/kubernetes/k-production-deployment.adoc
kbatuigas, line 1: From https://github.com/redpanda-data/docs/pull/589 cc @Deflaimun

modules/deploy/pages/deployment-option/self-hosted/kubernetes/local-guide.adoc
kbatuigas, line 1: https://github.com/redpanda-data/docs/pull/589

modules/deploy/partials/kubernetes/guides/deploy-redpanda.adoc
kbatuigas, line 1: https://github.com/redpanda-data/docs/pull/589

modules/reference/pages/properties/cluster-properties.adoc
kbatuigas, line 1: @mattschumpert We are removing the deprecated properties from this page and adding them to the [deprecated features](https://github.com/redpanda-data/docs/pull/592/files#diff-5abfc82b7bcc575f56c134942033175c6300617b1f473da68b52cbcc759d7cd1) page. That said, the current behavior for these properties seems to be that they are still visible, but you can't change them -- the rpk cluster config command returns an error. (With the exception of target_quota_byte_rate, which you _can_ still modify with rpk.)

Is that the intended experience and are you fine with this reference doc change? cc @Deflaimun

pgellert: > That said, the current behavior for these properties seems to be that they are still visible, but you can't change them -- the rpk cluster config command returns an error. (With the exception of target_quota_byte_rate, which you can still modify with rpk.)

I don't think this is the case. Why do you think so? I've now double-checked that all of the existing quota configs are still editable with a local build of `v24.2.1-rc2`:
```
((v24.2.1-rc2)) % rpk cluster config set target_quota_byte_rate=999999
Successfully updated configuration. New configuration version is 3.
((v24.2.1-rc2)) % rpk cluster config set target_fetch_quota_byte_rate=999999
Successfully updated configuration. New configuration version is 4.
((v24.2.1-rc2)) % rpk cluster config set kafka_admin_topic_api_rate=1000
Successfully updated configuration. New configuration version is 5.
((v24.2.1-rc2)) % rpk cluster config set kafka_client_group_byte_rate_quota="[{'group_name': 'first_group','clients_prefix': 'group_1','quota': 10240}]"
Successfully updated configuration. New configuration version is 6.
((v24.2.1-rc2)) % rpk cluster config set kafka_client_group_fetch_byte_rate_quota="[{'group_name': 'first_group','clients_prefix': 'group_1','quota': 10240}]"
Successfully updated configuration. New configuration version is 7.
```

kbatuigas: Thank you @pgellert , I see where I went wrong -- it was a syntax error on my part! All good now.

Still need @mattschumpert to confirm that he is good with removing these deprecate configs from the reference page, even though they can still be updated.

modules/upgrade/pages/deprecated/index.adoc
```diff
@@ -9,6 +9,16 @@ This index helps you to identify deprecated features in Redpanda releases and pl
...
 
+| 24.2.1
+a| Client throughput quota cluster configuration
+
+- `kafka_client_group_byte_rate_quota`
+- `kafka_client_group_fetch_byte_rate_quota`
+- `target_quota_byte_rate`
+- `target_fetch_quota_byte_rate`
+- `kafka_admin_topic_api_rate`
+| xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[Client throughput limits] are now Kafka API-based and should be managed using `rpk cluster quotas`.
```
Feediver1:
```suggestion
Use `rpk cluster quotas` to manage | xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[Client throughput limits] , which are now based on the Kafka API.
```

pgellert (approved):

modules/develop/pages/kafka-clients.adoc
```diff
@@ -47,6 +47,6 @@ Redpanda does not currently support the following Apache Kafka features:
 
 * Managing SASL users with Kafka APIs: DescribeUserSCRAMCredentialsRequest and AlterUserSCRAMCredentialsRequest APIs.
 * HTTP Proxy (pandaproxy): Unlike other REST proxy implementations in the Kafka ecosystem, Redpanda HTTP Proxy does not support topic and ACLs CRUD through the HTTP Proxy. HTTP Proxy is designed for clients producing and consuming data that do not perform administrative functions.
-* Quotas per user for bandwidth and API request rates (though xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[quotas per client and per client group (per CPU core)] are supported).
+* Quotas per user for bandwidth and API request rates (though xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[quotas per client and per client group] using AlterClientQuotas and DescribeClientQuotas APIs are supported).
```
Feediver1:
```suggestion
* Quotas per user for bandwidth and API request rates. However, xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[quotas per client and per client group] using AlterClientQuotas and DescribeClientQuotas APIs are supported.
```

modules/get-started/pages/whats-new.adoc
```diff
@@ -21,6 +21,13 @@ You can now deploy data transform functions that xref:develop:data-transforms/de
 
 The docs now also include an xref:develop:data-transforms/index.adoc[expanded guide] designed to help you master the creation, deployment, and management of data transforms in Redpanda.
 
+
+== Client throughput management
+
+Redpanda now includes `rpk` and Redpanda Console support for managing xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput limits] with the AlterClientQuotas and DescribeClientQuotas Kafka APIs. You can set client throughput quotas using `rpk` for individual client based on client ID, as well as a group of clients based on ID prefix matching. New xref:manage:cluster-maintenance/manage-throughput.adoc#monitor-client-throughput[metrics] are available for insight on client throughput utilization.
```
Feediver1:
```suggestion
Redpanda `rpk` and Redpanda Console now support managing xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput limits] using the AlterClientQuotas and DescribeClientQuotas Kafka APIs. You can set client throughput quotas using `rpk` for individual client based on client ID, as well as a group of clients based on ID prefix matching. New xref:manage:cluster-maintenance/manage-throughput.adoc#monitor-client-throughput[metrics] are available for insight on client throughput utilization.
```

mattschumpert: @kbatuigas a little confusing in that you don't need to actually use the APIs at all as a user when using RPK / Console. We're pushing our own internal implementation detail with this wording.  Its irrelevant and confusing as an RPK/Console user trying to do it (what? APIs?,  no. CLI or UI)

Separately (as a 3rd option), one can use these Kafka APIs directly indeed.

let's call out 3 distinct options.

mattschumpert: otherwise this lgtm

modules/get-started/pages/whats-new.adoc
```diff
@@ -21,6 +21,13 @@ You can now deploy data transform functions that xref:develop:data-transforms/de
...
+
+Client throughput quotas, previously applied on a per-shard basis, are also adjusted to apply on a per-broker basis. Cluster configuration properties for managing client quotas are deprecated, including `target_quota_byte_rate` which is disabled by default with the value `0`.
```
mattschumpert: 'also adjusted'?? no, they are now defined and enforced on a per-broker basis.  even if there is something going on internally at a core level lets not bother the user with that detail.

@kbatuigas

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -6,11 +6,13 @@ Redpanda supports applying throughput throttling on both ingress and egress inde
...
 
 == Broker-wide throughput limits
 
-Broker-wide throughput limits account for all Kafka API traffic going in to or out of the broker. The limit values represent the allowed rate of data in bytes per second passing through in each direction. Redpanda also provides administrators the ability to exclude clients from throughput throttling and to fine-tune which Kafka request types are subject to throttling limits.
+Broker-wide throughput limits account for all Kafka API traffic going into or out of the broker. The limit values represent the allowed rate of data in bytes per second passing through in each direction. Redpanda also provides administrators the ability to exclude clients from throughput throttling and to fine-tune which Kafka request types are subject to throttling limits.
```
mattschumpert: Maybe for clarity for newbies:  "in each direction, as data is Produced to or Consumed from a topic"

@pgellert exclusions?  really?   Are these still supported with the new quotas (I dont see it below)?

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
 
 == Client throughput limits
 
-Similar to <<broker-wide-throughput-limits>> but for clients, Redpanda provides configurable throughput quotas that apply to an individual client or a group of clients.
+Redpanda provides configurable throughput quotas that apply to an individual client or a group of clients. You can apply a quota for an individual client based on an exact match with its `client_id`, or a group of clients based on IDs that start with a given prefix. 
+
+As of v24.2, client throughput quotas are compatible with the https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client[Kafka API^] (AlterClientQuotas and DescribeClientQuotas), and are separate from quotas configured through cluster configuration in earlier Redpanda versions. The client throughput quotas no longer apply on a per-shard basis, and now limit the rates across a Redpanda broker's node. The quotas are neither shared nor balanced between brokers.
```
mattschumpert: perfect!

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+| Limit throughput of fetch requests
+
+| controller_mutation_rate
+| Limit rate of topic mutation requests, including create, add, and delete partition, in number of partitions per second
+
+|===
+
+You can also apply a default quota for all other client requests that don't have a specific quota configured based on the exact or prefix match on `client_id`. 
+
+It is possible to create conflicting quotas if you configure the same quotas through both the Kafka API and a cluster configuration. Redpanda resolves these conflicts by following an order of preference in finding a matching quota for a request:
```
mattschumpert, line 70: Nice work @pgellert , @michael-redpanda !!

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+. Quota configured through the Kafka API for an exact match on `client_id`
+. Quota configured through the Kafka API for a prefix match on `client_id`
+. Quota configured through cluster configuration properties (xref:reference:cluster-properties.adoc#kafka_client_group_byte_rate_quota[`kafka_client_group_byte_rate_quota`], xref:reference:cluster-properties.adoc#kafka_client_group_fetch_byte_rate_quota[`kafka_client_group_fetch_byte_rate_quota`]) for a prefix match on `client_id`
+. Default quota configured through the Kafka API on `client_id`
+. Default quota configured through cluster configuration properties on `client_id`
 
-NOTE: The client throughput quotas limit the rates within each shard (logical CPU core) of a Redpanda broker's node. The quotas are neither shared nor balanced between shards and brokers.
+Redpanda Data recommends <<migrate,migrating>> over from cluster configuration-managed quotas to managing the quotas through the Kafka API. You can recreate the configuration-based quotas through the Kafka API using `rpk`, and then remove the quota cluster configs.
```
mattschumpert: let's just use 'Redpanda' for company name here, right @kbatuigas , @Feediver1 ?

@kbatuigas seems worthwhile to provide a simple example here for getting the config using RPK and setting a similar quota via the new RPK commands.  Especially considering the units have changed, I think we need at least a small example here.  We can then link this from the upgrade guide (considering this is a required step we're calling out for upgrade, we should help out a bit in docs)

modules/get-started/pages/whats-new.adoc
Feediver1:
```suggestion
Client throughput quotas, previously applied on a per-shard basis, are now apply on a per-broker basis. Cluster configuration properties for managing client quotas are deprecated, including `target_quota_byte_rate` which is disabled by default with the value `0`.
```

Feediver1: Can you link to the deprecation entry in https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/upgrade/deprecated/ ? Wondering why it's not listed here? If it is deprecated, then we should always list in this table (along with all other deprecated properties).

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -6,11 +6,13 @@ Redpanda supports applying throughput throttling on both ingress and egress inde
 
 == Throughput throttling enforcement
 
+NOTE: As of v24.2, Redpanda enforces all throughput limits per broker, including client throughput.
```
Feediver1, line 9: Do you know the RC/build number in which this first appears and will be released? I've been seeing `24.2.1` and `24.2.2` and am wondering which will be accurate for the release.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
Feediver1: The link says **Kafka API**, which is what I was expecting to land on, but it actually takes you to a KIP. Maybe make AlterClientQuotas and DescribeClientQuotas the linked text?

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+| consumer_byte_rate
+| Limit throughput of fetch requests
+
+| controller_mutation_rate
+| Limit rate of topic mutation requests, including create, add, and delete partition, in number of partitions per second
+
+|===
+
+You can also apply a default quota for all other client requests that don't have a specific quota configured based on the exact or prefix match on `client_id`.
```
Feediver1: Not sure how to reword because I'm can figure out what your intent is here.
Perhaps:
You can also apply a default quota on all other client requests that don't have a specific quota based on an exact match or `client_id` prefix.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+|===
+
+You can also apply a default quota for all other client requests that don't have a specific quota configured based on the exact or prefix match on `client_id`. 
+
+It is possible to create conflicting quotas if you configure the same quotas through both the Kafka API and a cluster configuration. Redpanda resolves these conflicts by following an order of preference in finding a matching quota for a request:
+
+. Quota configured through the Kafka API for an exact match on `client_id`
+. Quota configured through the Kafka API for a prefix match on `client_id`
+. Quota configured through cluster configuration properties (xref:reference:cluster-properties.adoc#kafka_client_group_byte_rate_quota[`kafka_client_group_byte_rate_quota`], xref:reference:cluster-properties.adoc#kafka_client_group_fetch_byte_rate_quota[`kafka_client_group_fetch_byte_rate_quota`]) for a prefix match on `client_id`
```
Feediver1: These links do not take you to the actual property descriptions, but to the Properties page. Looks like the properties have not yet been pulled into the docs yet.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+. Default quota configured through the Kafka API on `client_id`
+. Default quota configured through cluster configuration properties on `client_id`
 
-NOTE: The client throughput quotas limit the rates within each shard (logical CPU core) of a Redpanda broker's node. The quotas are neither shared nor balanced between shards and brokers.
+Redpanda Data recommends <<migrate,migrating>> over from cluster configuration-managed quotas to managing the quotas through the Kafka API. You can recreate the configuration-based quotas through the Kafka API using `rpk`, and then remove the quota cluster configs.
+
+You can also use Redpanda Console to view client quotas that are enforced.
```
mattschumpert: @kbatuigas where is the Console part of this doc.  I don't see anything under https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/manage/console/ and we don't have a screenshot here (cause it might not exist yet :)).  

I think we either need a mention in the Console docs somewhere (that we link to here), or at least a screenshot here showing viewing quotas in console.

To be clear, this is still under development under https://github.com/redpanda-data/console-enterprise/issues/314,  but will be part of 24.2 GA. Not sure what the best page is to put it under for console.  maybe we just add a paragraph and screenshot here when available (I know it's been tested so @weeco  may be able to provide a screenshot already)

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
-To set throughput quota for a single client, configure the cluster properties xref:reference:cluster-properties.adoc#target_quota_byte_rate[`target_quota_byte_rate`] and xref:reference:cluster-properties.adoc#target_fetch_quota_byte_rate[`target_fetch_quota_byte_rate`].
+You can view current throughput quotas set through the Kafka API by running the `rpk cluster quotas describe` command.
+
+For example, to see the quotas for client ID `consumer-1`:
 
-* The `target_quota_byte_rate` property applies to a producer client that isn't a member of a client group configured by xref:reference:cluster-properties.adoc#kafka_client_group_byte_rate_quota[`kafka_client_group_byte_rate_quota`]. It sets the maximum throughput quota of a client sending to a Redpanda broker.
-* The `target_fetch_quota_byte_rate` property applies to a consumer client that isn't a member of a client group configured by xref:reference:cluster-properties.adoc#kafka_client_group_fetch_byte_rate_quota[`kafka_client_group_fetch_byte_rate_quota`]. It sets the maximum throughput quota of a client fetching from a Redpanda broker.
+[,bash]
+----
+rpk cluster quotas describe --name client-id=consumer-1
```
mattschumpert, line 88: Should we show a sample output here @kbatuigas ?

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+[,bash]
+----
+rpk cluster quotas alter --add consumer_byte_rate=200000 --name client-id=consumer-1
+----
 
 === Group of clients throughput limit
 
-To set throughput quota for a group of clients, configure the cluster properties xref:reference:cluster-properties.adoc#kafka_client_group_byte_rate_quota[`kafka_client_group_byte_rate_quota`] and xref:reference:cluster-properties.adoc#kafka_client_group_fetch_byte_rate_quota[`kafka_client_group_fetch_byte_rate_quota`].
+Alternatively, you can view or configure throughput quotas for a group of clients based on a match on client ID prefix. The following example sets the `consumer_byte_rate` quota to a group of clients that have IDs starting with `consumer-`:
```
Feediver1: Not sure--is "starting with `consumer-` different than "client IDs prefixed with `consumer-`"? Maybe just used "prefixed"?

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
mattschumpert: Nevermind, I see we have it below

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
 
-* The `kafka_client_group_byte_rate_quota` property applies to producer clients. It sets a maximum throughput quota of traffic sent to Redpanda from each producer in the group named by the property.
-* The `kafka_client_group_fetch_byte_rate_quota` property applies to consumer clients. It sets a maximum throughput quota of traffic fetched from Redpanda by each consumer in the group named by the property.
+[,bash]
+----
+rpk cluster quotas alter --add consumer_byte_rate=200000 --name client-id-prefix=consumer-
+----
+
+NOTE: The client group that matches on `client-id-prefix` does not denote a Kafka consumer group. Instead, it matches all requests with a `client_id` field that has the configured prefix. The `client_id` field is typically a configurable parameter in Kafka libraries when you create a client.
```
Feediver1:
```suggestion
NOTE: A client group specified in `client-id-prefix` is not the equivalent of a Kafka consumer group. Instead, it matches all requests configured with the prefix `client_id`. The `client_id` field is typically a configurable parameter available when you create a client in Kafka libraries.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
 
-Both `kafka_client_group_byte_rate_quota` and `kafka_client_group_fetch_byte_rate_quota` have the configuration value fields:
+=== Default client throughput limit
 
-* `group_name`: a name for the group of clients.
-* `clients_prefix`: a client belonging to the group has this prefix in its `client_id` property.
-* `quota`: the maximum throughput rate of *each client in the group* in bytes per second.
+You can apply default throughput limits to clients. Redpanda applies the default limits if no quotas have been configured specific to a client ID or ID prefix.
```
Feediver1:
```suggestion
You can apply default throughput limits to clients. Redpanda applies the default limits if no quotas are configured for a specific `client_id` or prefix.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+The example below sets a produce quota of 1GBps through the Kafka API, which means a 1GBps limit across all produce requests sent to a single Redpanda broker.
```
Feediver1:
```suggestion
To specify a produce quota of 1GBps through the Kafka API, which means a 1GBps limit across all produce requests sent to a single Redpanda broker: 
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+----
+rpk cluster quotas alter --default client-id --add producer_byte_rate=1000000000
+----
 
-NOTE: For `kafka_client_group_fetch_byte_rate_quota`, a group of consumer clients is not a Kafka consumer group, and the `group_name` field is not a Kafka consumer group ID. Instead, it is the set of clients fetching from a Redpanda broker that are configured by the property to be throughput limited.
+[[migrate]]
+=== Migrate cluster configuration quotas to Kafka API-based quotas
 
-An example configuration of `kafka_client_group_byte_rate_quota` for two groups of producers:
+. Use `rpk cluster config` to view current client quotas managed with cluster configuration. The example below shows how to retrieve the `kafka_client_group_byte_rate_quota` for two groups of producers:
```
Feediver1:
```suggestion
. Use `rpk cluster config` to view current client quotas managed with cluster configuration. The following example shows how to retrieve the `kafka_client_group_byte_rate_quota` for two groups of producers:
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
-* `clients_prefix`: a client belonging to the group has this prefix in its `client_id` property.
-* `quota`: the maximum throughput rate of *each client in the group* in bytes per second.
+You can apply default throughput limits to clients. Redpanda applies the default limits if no quotas have been configured specific to a client ID or ID prefix.
+
+The example below sets a produce quota of 1GBps through the Kafka API, which means a 1GBps limit across all produce requests sent to a single Redpanda broker. 
+
+[,bash]
+----
+rpk cluster quotas alter --default client-id --add producer_byte_rate=1000000000
+----
```
mattschumpert, line 130: @kbatuigas we also added an 'import' capability (just last week) to the RPK quotas support, which allows you to bulk import from a file to make it easier to manage a larger number of quotas.

This should also be documented.  It's mentioned here: https://github.com/redpanda-data/devex/issues/22

@r-vasquez has the details on the syntax.  But one import example here would be nice.

mattschumpert (approved):

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+| xref:reference:properties/cluster-properties.adoc#kafka_client_group_byte_rate_quota[`kafka_client_group_byte_rate_quota`]          
+| `--name client-id-prefix=<prefix> --add producer_byte_rate=<group-producer-byte-rate-value>`
+
+| xref:reference:properties/cluster-properties.adoc#kafka_client_group_fetch_byte_rate_quota[`kafka_client_group_fetch_byte_rate_quota`]    
+| `--name client-id-prefix=<prefix> --add consumer_byte_rate=<group-consumer-byte-rate-value>`
+
+|===
++
+The client throughput quotas set through the Kafka API apply per broker, so you need to convert the cluster configuration values that were applied on a per-shard (logical CPU core) basis. For example, if you set `target_fetch_quota_byte_rate` to 100MBps/shard, and you run Redpanda on 16-core brokers, you can set the new consumer_byte_rate quota to 100 * 16 = 1600MBps.
```
Feediver1:
```suggestion
The client throughput quotas set through the Kafka API apply per broker, so you must convert the cluster configuration values that were applied on a per-shard (logical CPU core) basis. For example, if you set `target_fetch_quota_byte_rate` to 100MBps/shard, and you run Redpanda on 16-core brokers, you can set the new consumer_byte_rate quota to 100 * 16 = 1600MBps.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,166 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+rpk cluster quotas alter --name client-id-prefix=producer_group_multiple --add producer_byte_rate=<group-2-producer-byte-rate-value> 
+----
++
+Replace the placeholder values with the new quota values, accounting for the conversion to per-broker limits, for example 10240 * broker core count = new quota.
+
+. Use `rpk cluster config` to remove the configuration-based quotas:
++
+[,bash]
+----
+rpk cluster config kafka_client_group_byte_rate_quota=
```
Feediver1: should there be a placeholder here for to the user to replace with their own value (inline edit of the example)?

modules/upgrade/pages/deprecated/index.adoc
```diff
@@ -9,6 +9,16 @@ This index helps you to identify deprecated features in Redpanda releases and pl
 |===
 | Deprecated in  | Feature | Details
 
+| 24.2.1
+a| Client throughput quota cluster configuration
```
Feediver1:
```suggestion
a| Client throughput quota cluster configuration:
```

modules/upgrade/pages/deprecated/index.adoc
```diff
@@ -9,6 +9,16 @@ This index helps you to identify deprecated features in Redpanda releases and pl
...
+- `kafka_client_group_byte_rate_quota`
+- `kafka_client_group_fetch_byte_rate_quota`
+- `target_quota_byte_rate`
+- `target_fetch_quota_byte_rate`
+- `kafka_admin_topic_api_rate`
+| Use xref:reference:rpk/rpk-cluster/rpk-cluster-quotas.adoc[`rpk cluster quotas`] to manage xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput limits] based on the Kafka API.
+
 | 23.3.2
 | `space_management_enable_override`
 | This property was a workaround in older version of Redpanda. It is no longer required and the property is ignored.
```
Feediver1:
```suggestion
| This property was a workaround in an older version of Redpanda. It is no longer required and the property is ignored.
```

modules/upgrade/pages/rolling-upgrade.adoc
```diff
@@ -23,6 +23,8 @@ include::partial$rolling-upgrades/upgrade-limitations.adoc[]
 
 Patch releases in https://github.com/redpanda-data/redpanda/discussions/9522[22.3.14^] and https://github.com/redpanda-data/redpanda/discussions/9523[23.1.2^] changed the behavior when remote read is disabled and the requested Raft term falls below the local log's beginning. In earlier versions, Redpanda returned an offset -1. With the patch, when you request a value older than the lowest offset, Redpanda returns the lowest offset, not -1.
 
+Version 24.2 introduced xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput quota management] using the AlterClientQuotas and DescribeClientQuotas https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client[Kafka APIs^]. Redpanda determines client throughput limits on a per-broker basis. In earlier versions, client throughput quotas were applied from cluster configurations on a per-shard basis.
```
Feediver1:
```suggestion
Starting in version 24.2,  xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput quota management] is available using the AlterClientQuotas and DescribeClientQuotas https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client[Kafka APIs^]. Redpanda determines client throughput limits on a per-broker basis. In earlier versions, client throughput quotas were applied from cluster configurations on a per-shard basis. 
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
r-vasquez: Hi @kbatuigas, the PR includes some examples: https://github.com/redpanda-data/redpanda/pull/21311

// With a full valid JSON
```
rpk cluster quotas import --from '{"quotas":[{"entity":[{"name":"foo","type":"client-id"}],"values":[{"key":"consumer_byte_rate","values":"12123123"}]},{"entity":[{"name":"foo-","type":"client-id-prefix"}],"values":[{"key":"producer_byte_rate","values":"12123123"},{"key":"consumer_byte_rate","values":"4444444"}]}]}'
```

Something important to document, is that we mention in the help text that `import` takes the output of `rpk cluster quotas describe --format <json/yaml>`, so `describe --format` acts like an `export` command.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
kbatuigas: I don't think I've seen anything more specific or granular, @pgellert could you confirm?

modules/get-started/pages/whats-new.adoc
```diff
@@ -21,6 +21,13 @@ You can now deploy data transform functions that xref:develop:data-transforms/de
 
 The docs now also include an xref:develop:data-transforms/index.adoc[expanded guide] designed to help you master the creation, deployment, and management of data transforms in Redpanda.
 
+
+== Client throughput management
+
+Redpanda now includes `rpk` and Redpanda Console support for managing xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput limits]. You can set client throughput quotas using `rpk` for individual client based on client ID, as well as a group of clients based on ID prefix matching. New xref:manage:cluster-maintenance/manage-throughput.adoc#monitor-client-throughput[metrics] are available for insight on client throughput utilization.
```
kbatuigas: @mattschumpert I removed the specific mention of AlterClientQuotas and DescribeClientQuotas. Is this less confusing? With that said, the advice that I got from @pgellert is to try and push users to `rpk` for simplicity. In terms of using the APIs directly, a tool like `kafka-configs.sh` doesn't support client-id-prefix quotas since those are specific to Redpanda. Console doesn't yet support editing quotas in the UI as far as I'm aware -- maybe @weeco can confirm again.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,198 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+[,bash]
+----
+rpk cluster quotas describe --name client-id=consumer-1
+----
+
+[,bash,role=no-copy]
+----
+ENTITY                QUOTA-TYPE          QUOTA-VALUE
+client-id=consumer-1  producer_byte_rate  140000
```
kbatuigas: @r-vasquez could you confirm if this output is correct? Since the example command doesn't include default or a user I wasn't sure if this is correct under ENTITY

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,198 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+To more easily manage multiple quotas, you can use the `cluster quotas describe` and `cluster quotas import` commands to do a bulk export and update.
 
-Both `kafka_client_group_byte_rate_quota` and `kafka_client_group_fetch_byte_rate_quota` have the configuration value fields:
+For example, to export all client quotas in JSON format:
 
-* `group_name`: a name for the group of clients.
-* `clients_prefix`: a client belonging to the group has this prefix in its `client_id` property.
-* `quota`: the maximum throughput rate of *each client in the group* in bytes per second.
+[,bash]
+----
```
kbatuigas, line 139: @r-vasquez does `cluster quotas describe --format json` output pure JSON and not, say, to a JSON file? If it's pure JSON, should we tell users to save the command output to a file so they can just provide the filepath as in the `import` example below? Or is it preferable to provide the `--source` argument as a string?

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
kbatuigas: @Feediver1 This syntax "clears out" the config, but actually @r-vasquez could you confirm if the explanation is correct? Based on the existing reference doc I'm not sure that the command actually removes the quota (it says "reset to default").

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,198 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+. Use `rpk cluster config` to remove the configuration-based quotas:
++
+[,bash]
+----
+rpk cluster config kafka_client_group_byte_rate_quota=
+----
+
+=== View throughput limits in Redpanda Console
+
+You can also use Redpanda Console to view enforced limits. In the menu, go to **Quotas**.
```
kbatuigas, line 221: @weeco Is there anything else with the Console view that needs to be called out here?

@mattschumpert It looks like the Console-specific content for this might be super light at the moment, so I chose to add it in this doc instead of creating a new standalone Console doc. However, if you have a suggestion for where it could go in the [existing Console docs](https://deploy-preview-592--redpanda-docs-preview.netlify.app/current/manage/console/) I can move it there too. I hesitate to add a screenshot here because I don't know how quickly it will become out of date.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,198 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+|===
+
+You can also apply a default quota for all other client requests that don't have a specific quota based on an exact match or `client_id` prefix. 
+
+It is possible to create conflicting quotas if you configure the same quotas through both the Kafka API and a cluster configuration. Redpanda resolves these conflicts by following an order of preference in finding a matching quota for a request:
+
+. Quota configured through the Kafka API for an exact match on `client_id`
+. Quota configured through the Kafka API for a prefix match on `client_id`
+. Quota configured through cluster configuration properties (`kafka_client_group_byte_rate_quota`, `kafka_client_group_fetch_byte_rate_quota`, xref:upgrade:deprecated/index.adoc[deprecated starting in v24.2]) for a prefix match on `client_id`
```
kbatuigas, line 74: @Feediver1 is it all right to link to the Deprecated page here? These properties are removed from the reference in this PR so we no longer have any xrefs to user specifically for these individual properties.

modules/get-started/pages/whats-new.adoc
weeco: Correct, no editing available - just viewing. We are considering working on this as a stretch but I'm almost confident that we won't be able to include this in the next release, so not relevant for this PR.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
pgellert: This is going out with the v24.2 major version release, so it is already in the beta releases v24.2.1-rc1 through v24.2.1-rc4 and the first non beta release v24.2.1 will contain the change. I think phrasing it as either v24.2 or v24.2.1 makes sense.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
pgellert: This section is about the node-wide (not client id specific) ingress/egress quotas. They do support exclusions but the client quotas didn't and still don't.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
r-vasquez: @kbatuigas No, the output is not in table format, here is the output for this example:

```
$ rpk cluster quotas describe --name client-id=consumer-1
client-id=consumer-1
	producer_byte_rate=140000
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
r-vasquez: `rpk cluster quotas import --from` accepts either a JSON, a YAML, or a file containing a JSON/YAML. 

`rpk cluster quotas describe --format json` returns the pure JSON.

So, I think it's fine to tell users to use that string, but we can mention that can be saved to files and used accordingly

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,198 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+rpk cluster quotas describe --format json
+----
 
-NOTE: For `kafka_client_group_fetch_byte_rate_quota`, a group of consumer clients is not a Kafka consumer group, and the `group_name` field is not a Kafka consumer group ID. Instead, it is the set of clients fetching from a Redpanda broker that are configured by the property to be throughput limited.
+`rpk cluster quotas import` accepts the output from `rpk cluster quotas describe --format <json/yaml>`. :
 
-An example configuration of `kafka_client_group_byte_rate_quota` for two groups of producers:
+[,bash]
+----
+rpk cluster quotas import --source <path-to-output-file>
```
r-vasquez: I'm sorry I gave you an old example, during development we changed the flag `--source` to `from`

```suggestion
rpk cluster quotas import --from <string-or-path-to-file>
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
r-vasquez: Yes, the default is an empty list and should be fine.

However I think it's missing the `set`:
```suggestion
rpk cluster config set kafka_client_group_byte_rate_quota=
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
mattschumpert: I think I agree.  @weeco  is actually working on this now and the screenshot may change this week as its finalized.

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
mattschumpert: my bad, thanks. But are these Kafka API configs or just still cluster configs?  

Might get confusing if we dont specify.

Deflaimun:
rpk command included with https://github.com/redpanda-data/docs/pull/611

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
pgellert: Still just cluster configs. The one controlling exclusions specifically is [kafka_throughput_control](https://docs.redpanda.com/current/reference/properties/cluster-properties/#kafka_throughput_control). This one, as well as the other ingress/egress cluster configs are listed in the section below the one this comment links to. Maybe the title of it would be better as "Broker-wide throughput limit ~~properties~~ *cluster configs*"

modules/get-started/pages/whats-new.adoc
```diff
@@ -25,6 +25,13 @@ You can now deploy data transform functions that xref:develop:data-transforms/de
...
```
micheleRP:
```suggestion
Redpanda now includes `rpk` and Redpanda Console support for managing xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput limits]. You can set client throughput quotas using `rpk` for an individual client based on client ID, as well as for a group of clients based on ID prefix matching. New xref:manage:cluster-maintenance/manage-throughput.adoc#monitor-client-throughput[metrics] are available for insights on client throughput utilization.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+|===
+
+You can also apply a default quota for all other client requests that don't have a specific quota based on an exact match or `client_id` prefix. 
+
+It is possible to create conflicting quotas if you configure the same quotas through both the Kafka API and a cluster configuration. Redpanda resolves these conflicts by following an order of preference in finding a matching quota for a request:
 
-NOTE: The client throughput quotas limit the rates within each shard (logical CPU core) of a Redpanda broker's node. The quotas are neither shared nor balanced between shards and brokers.
+. Quota configured through the Kafka API for an exact match on `client_id`
+. Quota configured through the Kafka API for a prefix match on `client_id`
+. Quota configured through cluster configuration properties (`kafka_client_group_byte_rate_quota`, `kafka_client_group_fetch_byte_rate_quota`, xref:upgrade:deprecated/index.adoc[deprecated starting in v24.2]) for a prefix match on `client_id`
```
micheleRP, line 74:
```suggestion
. Quota configured through cluster configuration properties (`kafka_client_group_byte_rate_quota`, `kafka_client_group_fetch_byte_rate_quota`, xref:upgrade:deprecated/index.adoc[deprecated in v24.2]) for a prefix match on `client_id`
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+. Default quota configured through the Kafka API on `client_id`
+. Default quota configured through cluster configuration properties (`target_quota_byte_rate`, `target_fetch_quota_byte_rate`, `kafka_admin_topic_api_rate`, xref:upgrade:deprecated/index.adoc[deprecated starting in v24.2]) on `client_id`
```
micheleRP, line 76:
```suggestion
. Default quota configured through cluster configuration properties (`target_quota_byte_rate`, `target_fetch_quota_byte_rate`, `kafka_admin_topic_api_rate`, xref:upgrade:deprecated/index.adoc[deprecated in v24.2]) on `client_id`
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
micheleRP: There might be a better way to point out that steps 3 & 5 are deprecated. Maybe start the step with: Deprecated:? Or put just (deprecated) next to the properties and add a note at the end of the list saying that the steps are included for completeness, but since the properties in steps 3 & 5 are deprecated in 24.2, steps 3 and 5 aren't considered in 24.2+ deployments?

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+Redpanda recommends <<migrate,migrating>> over from cluster configuration-managed quotas to Kafka-compatible quotas. You can recreate the configuration-based quotas with `rpk`, and then remove the quota cluster configs.
```
micheleRP:
```suggestion
Redpanda recommends <<migrate,migrating>> over from cluster configuration-managed quotas to Kafka-compatible quotas. You can re-create the configuration-based quotas with `rpk`, and then remove the cluster configurations.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
 
 === Individual client throughput limit
 
-To set throughput quota for a single client, configure the cluster properties xref:reference:cluster-properties.adoc#target_quota_byte_rate[`target_quota_byte_rate`] and xref:reference:cluster-properties.adoc#target_fetch_quota_byte_rate[`target_fetch_quota_byte_rate`].
+You can view current throughput quotas set through the Kafka API by running the xref:reference:rpk/rpk-cluster/rpk-cluster-quotas-describe.adoc[`rpk cluster quotas describe`] command.
```
micheleRP:
```suggestion
To view current throughput quotas set through the Kafka API, run xref:reference:rpk/rpk-cluster/rpk-cluster-quotas-describe.adoc[`rpk cluster quotas describe`].
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+	producer_byte_rate=140000
+----
+
+To set a throughput quota for a single client, use the xref:reference:rpk/rpk-cluster/rpk-cluster-quotas-alter.adoc[`rpk cluster quotas alter`] command. 
+
+[,bash]
+----
+rpk cluster quotas alter --add consumer_byte_rate=200000 --name client-id=consumer-1
+----
++
```
micheleRP: @kbatuigas I think this is a typo. Please check the rendering [here](https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/manage/cluster-maintenance/manage-throughput/#individual-client-throughput-limit).

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
 
-* The `kafka_client_group_byte_rate_quota` property applies to producer clients. It sets a maximum throughput quota of traffic sent to Redpanda from each producer in the group named by the property.
-* The `kafka_client_group_fetch_byte_rate_quota` property applies to consumer clients. It sets a maximum throughput quota of traffic fetched from Redpanda by each consumer in the group named by the property.
+[,bash]
+----
+rpk cluster quotas alter --add consumer_byte_rate=200000 --name client-id-prefix=consumer-
+----
 
-Both `kafka_client_group_byte_rate_quota` and `kafka_client_group_fetch_byte_rate_quota` have the configuration value fields:
+NOTE: A client group specified with `client-id-prefix` is not the equivalent of a Kafka consumer group. It is used only to match requests based on the `client_id` prefix. The `client_id` field is typically a configurable parameter when you create a client with Kafka libraries.
```
micheleRP:
```suggestion
NOTE: A client group specified with `client-id-prefix` is not the equivalent of a Kafka consumer group. It is used only to match requests based on the `client_id` prefix. The `client_id` field is typically a configurable property when you create a client with Kafka libraries.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
-* `group_name`: a name for the group of clients.
-* `clients_prefix`: a client belonging to the group has this prefix in its `client_id` property.
-* `quota`: the maximum throughput rate of *each client in the group* in bytes per second.
+=== Default client throughput limit
 
-NOTE: For `kafka_client_group_fetch_byte_rate_quota`, a group of consumer clients is not a Kafka consumer group, and the `group_name` field is not a Kafka consumer group ID. Instead, it is the set of clients fetching from a Redpanda broker that are configured by the property to be throughput limited.
+You can apply default throughput limits to clients. Redpanda applies the default limits if no quotas are configured for a specific `client_id` or prefix.
 
-An example configuration of `kafka_client_group_byte_rate_quota` for two groups of producers:
+To specify a produce quota of 1GBps through the Kafka API, which means a 1GBps limit across all produce requests sent to a single Redpanda broker:
```
micheleRP:
```suggestion
To specify a produce quota of 1 GBps through the Kafka API (which means a 1 GBps limit across all produce requests sent to a single Redpanda broker), run: 
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+----
+
+`rpk cluster quotas import` accepts the output string from `rpk cluster quotas describe --format <json/yaml>`:
+
+[,bash]
+----
+rpk cluster quotas import --from '{"quotas":[{"entity":[{"name":"foo","type":"client-id"}],"values":[{"key":"consumer_byte_rate","values":"12123123"}]},{"entity":[{"name":"foo-","type":"client-id-prefix"}],"values":[{"key":"producer_byte_rate","values":"12123123"},{"key":"consumer_byte_rate","values":"4444444"}]}]}'
+----
+
+You may also save the JSON or YAML output to a file and pass the file path in the `--from` flag.
```
micheleRP:
```suggestion
You can also save the JSON or YAML output to a file and pass the file path in the `--from` flag.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+| xref:reference:properties/cluster-properties.adoc#kafka_admin_topic_api_rate[`kafka_admin_topic_api_rate`]                  
+| `--default client-id --add controller_mutation_rate=<controller-mutation-rate-value>`
+
+| xref:reference:properties/cluster-properties.adoc#kafka_client_group_byte_rate_quota[`kafka_client_group_byte_rate_quota`]          
+| `--name client-id-prefix=<prefix> --add producer_byte_rate=<group-producer-byte-rate-value>`
+
+| xref:reference:properties/cluster-properties.adoc#kafka_client_group_fetch_byte_rate_quota[`kafka_client_group_fetch_byte_rate_quota`]    
+| `--name client-id-prefix=<prefix> --add consumer_byte_rate=<group-consumer-byte-rate-value>`
+
```
micheleRP, line 197: @kbatuigas: Suggest removing all these links and adding a reminder note about how they're deprecated in 24.2

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+|===
++
+The client throughput quotas set through the Kafka API apply per broker, so you must convert the cluster configuration values that were applied on a per-shard (logical CPU core) basis. For example, if you set `target_fetch_quota_byte_rate` to 100MBps/shard, and you run Redpanda on 16-core brokers, you can set the new consumer_byte_rate quota to 100 * 16 = 1600MBps.
```
micheleRP:
```suggestion
The client throughput quotas set through the Kafka API apply per broker, so you must convert the cluster configuration values that were applied on a per-shard (logical CPU core) basis. For example, if you set `target_fetch_quota_byte_rate` to 100 MBps/shard, and you run Redpanda on 16-core brokers, you can set the new consumer_byte_rate quota to 100 * 16 = 1600 MBps.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+. Use `rpk cluster quotas alter` to set the corresponding client throughput quotas based on the Kafka API:
++
+[,bash]
+----
+rpk cluster quotas alter --name client-id-prefix=producer_group_alone_producer --add producer_byte_rate=<group-1-producer-byte-rate-value> 
+rpk cluster quotas alter --name client-id-prefix=producer_group_multiple --add producer_byte_rate=<group-2-producer-byte-rate-value> 
+----
++
+Replace the placeholder values with the new quota values, accounting for the conversion to per-broker limits, for example 10240 * broker core count = new quota.
```
micheleRP:
```suggestion
Replace the placeholder values with the new quota values, accounting for the conversion to per-broker limits. For example, 10240 * broker core count = new quota.
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+rpk cluster config set kafka_client_group_byte_rate_quota=
+----
+
+=== View throughput limits in Redpanda Console
+
+You can also use Redpanda Console to view enforced limits. In the menu, go to **Quotas**.
+
+=== Monitor client throughput
+
+The following metrics are available on both the `/public_metrics` and `/metrics` endpoints to help provide insight into client throughput quota usage:
```
micheleRP:
```suggestion
The following metrics are available on both the `/public_metrics` and `/metrics` endpoints to provide insights into client throughput quota usage:
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+
+* Client quota throughput per rule and quota type
```
micheleRP:
```suggestion
* Client quota throughput per rule and quota type:
```

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
```diff
@@ -43,48 +45,200 @@ The properties for broker-wide throughput quota balancing are configured at the
...
+** `/public_metrics` - xref:reference:public-metrics-reference.adoc#redpanda_kafka_quotas_client_quota_throughput[`redpanda_kafka_quotas_client_quota_throughput`]
+** `/metrics` - xref:reference:internal-metrics-reference.adoc#vectorized_kafka_quotas_client_quota_throughput[`vectorized_kafka_quotas_client_quota_throughput`]
+* Client quota throttling delay, in seconds, per rule and quota type
```
micheleRP:
```suggestion
* Client quota throttling delay per rule and quota type, in seconds:
```

modules/upgrade/pages/rolling-upgrade.adoc
```diff
@@ -23,6 +23,8 @@ include::partial$rolling-upgrades/upgrade-limitations.adoc[]
 
 Patch releases in https://github.com/redpanda-data/redpanda/discussions/9522[22.3.14^] and https://github.com/redpanda-data/redpanda/discussions/9523[23.1.2^] changed the behavior when remote read is disabled and the requested Raft term falls below the local log's beginning. In earlier versions, Redpanda returned an offset -1. With the patch, when you request a value older than the lowest offset, Redpanda returns the lowest offset, not -1.
 
+Starting in version 24.2, xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput limits] are compatible with the https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client[AlterClientQuotas and DescribeClientQuotas^] Kafka APIs. Redpanda determines client throughput limits on a per-broker basis. In earlier versions, client throughput quotas were applied from cluster configurations on a per-shard basis. 
+
```
micheleRP, line 25: I think we should start the list with the most recent version, so moving this paragraph above the previous one.

modules/reference/pages/public-metrics-reference.adoc
```diff
@@ -497,6 +497,32 @@ Latency of produce/consume requests per broker. This duration measures from when
...
+
+Client quota throttling delay, in seconds, per rule and quota type.
+
+*Type*: histogram
+
+*Labels*:
+
+* quota_rule=("not_applicable" | "kafka_client_default" | "cluster_client_default" | "kafka_client_prefix" | "cluster_client_prefix" | "kafka_client_id")
+* quota_type=("produce_quota" | "fetch_quota" | "partition_mutation_quota")
+
```
micheleRP, line 510: @kbatuigas please check that these render consistent with the other metrics on this page. See [here](https://deploy-preview-592--redpanda-docs-preview.netlify.app/24.2/reference/public-metrics-reference/#redpanda_kafka_quotas_client_quota_throttle_time) compared to the one above it. 
And same comment for internal metrics page!

modules/manage/pages/cluster-maintenance/manage-throughput.adoc
kbatuigas: I think they would still be considered if they happen to be set @micheleRP (those properties are [still editable in 24.2](https://github.com/redpanda-data/docs/pull/592#discussion_r1675655005)) -- @pgellert can correct me if I've misunderstood, but I'm not sure we can say that 3 and 5 won't be considered.

pgellert: They are still considered in v24.2. After the migration period, in v25.1, these cluster configs are going to be removed and those steps won't be considered anymore.

micheleRP (approved):
LGTM

Feediver1:
Anything remaining that is holding up this merge @kbatuigas ?

PR #609, opened 2024-07-18
Intra broker partition balancing

kbatuigas:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2425
Review deadline: 24 July 2024

Related to https://github.com/redpanda-data/docs/pull/615 (Admin API reference update)

## Page previews

[Cluster balancing
](https://deploy-preview-609--redpanda-docs-preview.netlify.app/24.2/manage/cluster-maintenance/cluster-balancing/)
Internal metrics reference > [vectorized_reactor_utilization](https://deploy-preview-609--redpanda-docs-preview.netlify.app/24.2/reference/internal-metrics-reference/#vectorized_reactor_utilization)
[What's New](https://deploy-preview-609--redpanda-docs-preview.netlify.app/24.2/get-started/whats-new/#intra-broker-partition-balancing)

## Checks

- [x] New feature
- [ ] Content gap
- [ ] Support Follow-up


modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,34 @@ This requires an Enterprise license.
 This mode is not recommended for production clusters. Only set to `off` if you need to move partitions manually.
 |===
 
+== Intra-broker partition balancing
+
+In Redpanda, every partition replica is assigned to a core on a broker. While Redpanda's default <<partition-replica-balancing,partition balancing>> monitors cluster-level events such as the addition of new brokers or broker failure to balance partition assignments, it does not account for the distribution of partitions _within_ an individual broker. That means that some cores on a broker may inadvertently host many partitions of heavily used topics and cause the CPU to be xref:manage:monitoring.adoc#cpu-usage[overburdened].
```
kbatuigas: @ztlpn are there other metrics or other things users could do to monitor and then determine whether they should manually trigger a rebalance?

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,34 @@ This requires an Enterprise license.
...
+
+Additionally, when the partition rebalance moves some partitions away from a broker, the remaining partitions do not necessarily rebalance across the broker's cores. Or, if a broker's core count is increased, Redpanda does not assign any partitions to the new cores until new partitions are created or old partitions are moved out.
+
+Intra-broker partition balancing allows for dynamically reassigning partitions within a broker. If a broker's core counts increases, when the broker starts back up, Redpanda can check partition assignments across the broker's cores, and reassign partitions so that a balanced assignment is maintained across all cores. Redpanda can also check partition assignments when partitions are removed from a broker, and rebalance the remaining partitions between cores.
+
+To enable intra-broker partition balancing, configure the following broker properties:
```
kbatuigas: @ztlpn Are these enabled by default for new 24.2 clusters (and should we add a note regarding this)? Will these properties be made available for earlier versions as well?

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,34 @@ This requires an Enterprise license.
...
+[,bash]
+----
+curl -X POST http://localhost:9644/v1/partitions/rebalance_cores
+----
+
+To check the new partition assignments, run:
+
+[,bash]
+----
+
```
kbatuigas: @ztlpn should this be another Admin API call or should it be an rpk command?

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
ztlpn: > Are these enabled by default for new 24.2 clusters (and should we add a note regarding this)?

There is a feature flag "node_local_core_assignment" that gets enabled automatically for new 24.2 clusters. For upgraded clusters this flag will be disabled and the properties below will have no effect. There is a way to enable it manually, though I don't see it described elsewhere in the docs. Perhaps we can add a note along the lines of "contact support if you want to enable it for an upgraded cluster".

> Will these properties be made available for earlier versions as well?

No.

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,34 @@ This requires an Enterprise license.
...
+Intra-broker partition balancing allows for dynamically reassigning partitions within a broker. If a broker's core counts increases, when the broker starts back up, Redpanda can check partition assignments across the broker's cores, and reassign partitions so that a balanced assignment is maintained across all cores. Redpanda can also check partition assignments when partitions are removed from a broker, and rebalance the remaining partitions between cores.
+
+To enable intra-broker partition balancing, configure the following broker properties:
+
+* `core_balancing_on_core_count_change` - Set to `true` to rebalance partition assignments across cores after startup, if core count increases. Default: `true`.
+* `core_balancing_continuous` - Set to `true` to rebalance partition assignments across cores in runtime. Default: `true`. 
+** `core_balancing_debounce_timeout` - If balancing partition assignments across cores in runtime, set a window in which Redpanda will wait to balance assignments after core balancing is triggered. Default: 10000 ms (10 seconds).
```
ztlpn: Not sure we'd want to mention this one here. It is just a tunable, not supposed to be changed in most cases.

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
ztlpn: Admin API for now:
```
curl http://localhost:9644/v1/partitions
```

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
ztlpn: That would be the `vectorized_reactor_utilization` metric. Here are some relevant internal docs: https://redpandadata.atlassian.net/wiki/x/EYC7Gw (CPU section)

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,34 @@ This requires an Enterprise license.
...
+
+Additionally, when the partition rebalance moves some partitions away from a broker, the remaining partitions do not necessarily rebalance across the broker's cores. Or, if a broker's core count is increased, Redpanda does not assign any partitions to the new cores until new partitions are created or old partitions are moved out.
+
+Intra-broker partition balancing allows for dynamically reassigning partitions within a broker. If a broker's core counts increases, when the broker starts back up, Redpanda can check partition assignments across the broker's cores, and reassign partitions so that a balanced assignment is maintained across all cores. Redpanda can also check partition assignments when partitions are removed from a broker, and rebalance the remaining partitions between cores.
+
+To enable intra-broker partition balancing, configure the following broker properties:
+
+* `core_balancing_on_core_count_change` - Set to `true` to rebalance partition assignments across cores after startup, if core count increases. Default: `true`.
```
ztlpn: if core count decreases as well

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,34 @@ This requires an Enterprise license.
...
+
+Additionally, when the partition rebalance moves some partitions away from a broker, the remaining partitions do not necessarily rebalance across the broker's cores. Or, if a broker's core count is increased, Redpanda does not assign any partitions to the new cores until new partitions are created or old partitions are moved out.
+
+Intra-broker partition balancing allows for dynamically reassigning partitions within a broker. If a broker's core counts increases, when the broker starts back up, Redpanda can check partition assignments across the broker's cores, and reassign partitions so that a balanced assignment is maintained across all cores. Redpanda can also check partition assignments when partitions are removed from a broker, and rebalance the remaining partitions between cores.
```
ztlpn: Perhaps we should emphasize that these are problems with the old behavior and the new feature solves them.

Also we should mention that with this feature enabled, core count decrease is now possible.

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,33 @@ This requires an Enterprise license.
...
+
+To enable intra-broker partition balancing, configure the following broker properties:
+
+* `core_balancing_on_core_count_change` - Set to `true` to rebalance partition assignments across cores after startup, if core count increases or decreases. Default: `true`.
+* `core_balancing_continuous` - Set to `true` to rebalance partition assignments across cores in runtime. Default: `true`.
```
ztlpn: It is false by default

modules/reference/pages/internal-metrics-reference.adoc
```diff
@@ -221,7 +221,7 @@ High value can indicate nodes failing and causing leadership changes.
 
 Redpanda process utilization.
 
-Shows the true utilization of the CPU by a Redpanda process.
+Shows the true utilization of the CPU by a Redpanda process. Check this metric on a per-shard basis. If a shard is at 100% utilization for a continuous period of time, it likely results in high latency.
```
kbatuigas: @ztlpn Is this guidance OK to provide to users? Could we provide a bit more information such as how we should define "continuous period of time", and is there a specific way to calculate per shard? Our doc says
```
sum(vectorized_reactor_utilization{instance=~"[[node]]",shard=~"[[node_shard]]"}) by ([[aggr_criteria]])
```
But I am not very familiar with the syntax inside the {} brackets. Thanks!

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
kbatuigas: @mattschumpert is there specific guidance we want to provide regarding upgrades, or is contacting Support sufficient?

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,35 @@ This requires an Enterprise license.
...
+
+In Redpanda versions earlier than 24.2, this means that some cores on a broker may inadvertently host many partitions of heavily used topics and cause the CPU to be xref:manage:monitoring.adoc#cpu-usage[overburdened]. Additionally, when the partition rebalance moves some partitions away from a broker, the remaining partitions do not necessarily rebalance across the broker's cores. Or, if a broker's core count is increased, Redpanda does not assign any partitions to the new cores until new partitions are created or old partitions are moved out.
+
+Starting from v24.2, intra-broker partition balancing allows for dynamically reassigning partitions within a broker. If a broker's core count changes, when the broker starts back up, Redpanda can check partition assignments across the broker's cores, and reassign partitions so that a balanced assignment is maintained across all cores. Redpanda can also check partition assignments when partitions are removed from a broker, and rebalance the remaining partitions between cores.
```
mattschumpert: @kbatuigas @ztlpn  this says nothing of the topic-wise distribution. We should definitely mention that here.

It also is a little hidden that we _also_ do the rebalance on adding partitions to a broker (I think maybe mentioned above). But we should maybe list all events that trigger rebalance now

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
mattschumpert: @ztlpn can we not just enable the _feature_ flag for all clusters but not the new continuous behavior?  do we really need to keep the featue flag off (which could lead to confusion for customers who want to selectively turn it on, or if we want to roll it out in cloud in a minor by enabling the behavior flag

cc @piyushredpanda

modules/reference/pages/internal-metrics-reference.adoc
ztlpn: > Is this guidance OK to provide to users?

Yes, looks OK. We can also mention that high latency will be observed only for partitions assigned to that core and mention the need for balancing.

> how we should define "continuous period of time"

I would say more than a few seconds, but it really depends on the workload. The user may not care for small latency spikes here and there if they do some kind of bulk loading. OTOH they would want it to never reach 100% if they are doing realtime processing.

> is there a specific way to calculate per shard

Inside `{}` are filters - if you want to view utilization for specific nodes and/or shards you specify them there, but usually I would start with viewing all cores of the cluster. As the metric has per-shard granularity and we want to view it per-shard, we don't really need `sum` either. So we can just mention the metric name and that it has per-shard granularity and let the user figure out the way they want to filter it themselves.

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,35 @@ This requires an Enterprise license.
 This mode is not recommended for production clusters. Only set to `off` if you need to move partitions manually.
 |===
 
+== Intra-broker partition balancing
+
+In Redpanda, every partition replica is assigned to a core on a broker. While Redpanda's default <<partition-replica-balancing,partition balancing>> monitors cluster-level events such as the addition of new brokers or broker failure to balance partition assignments, it does not account for the distribution of partitions _within_ an individual broker.
```
ztlpn: I guess we should mention early that we mean CPU cores here. I.e. "a core on a broker" -> "a CPU core on a broker"

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
ztlpn: We should probably mention the topic-aware aspect here.

kbatuigas: @mattschumpert In the [current version of this doc](https://docs.redpanda.com/current/manage/cluster-maintenance/cluster-balancing/#partition-replica-balancing) we expand on the "topic-aware" concept and provide an example in the Partition replica balancing section. Is it basically the same concept that applies to intra-broker, and should we rehash it in the new section as well?

mattschumpert: We don't have to have another example, we should just mention it.

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
mattschumpert: We probably DO want to mention here that  'core_balancing_continuous' is an Enterprise-only feature though, right @ztlpn ? cc @kbatuigas

ztlpn (approved):

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,35 @@ This requires an Enterprise license.
 This mode is not recommended for production clusters. Only set to `off` if you need to move partitions manually.
 |===
 
+== Intra-broker partition balancing
+
+In Redpanda, every partition replica is assigned to a CPU core on a broker. While Redpanda's default <<partition-replica-balancing,partition balancing>> monitors cluster-level events such as the addition of new brokers or broker failure to balance partition assignments, it does not account for the distribution of partitions _within_ an individual broker.
```
micheleRP:
```suggestion
In Redpanda, every partition replica is assigned to a CPU core on a broker. While Redpanda's default <<partition-replica-balancing,partition balancing>> monitors cluster-level events, such as the addition of new brokers or broker failure to balance partition assignments, it does not account for the distribution of partitions _within_ an individual broker. 
```

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,35 @@ This requires an Enterprise license.
 This mode is not recommended for production clusters. Only set to `off` if you need to move partitions manually.
 |===
 
+== Intra-broker partition balancing
+
+In Redpanda, every partition replica is assigned to a CPU core on a broker. While Redpanda's default <<partition-replica-balancing,partition balancing>> monitors cluster-level events such as the addition of new brokers or broker failure to balance partition assignments, it does not account for the distribution of partitions _within_ an individual broker. 
+
+In Redpanda versions earlier than 24.2, this means that some cores on a broker may inadvertently host many partitions of heavily used topics and cause the CPU to be xref:manage:monitoring.adoc#cpu-usage[overburdened]. Additionally, when the partition rebalance moves some partitions away from a broker, the remaining partitions do not necessarily rebalance across the broker's cores. Or, if a broker's core count is increased, Redpanda does not assign any partitions to the new cores until new partitions are created or old partitions are moved out.
```
micheleRP:
```suggestion
Prior to Redpanda version 24.2, this meant that some cores on a broker could inadvertently host many partitions of heavily-used topics and cause the CPU to be xref:manage:monitoring.adoc#cpu-usage[overburdened]. Additionally, when the partition rebalance moved some partitions away from a broker, the remaining partitions did not necessarily rebalance across the broker's cores. Or, if a broker's core count was increased, Redpanda did not assign any partitions to the new cores until new partitions were created or old partitions were moved out.
```

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,35 @@ This requires an Enterprise license.
...
+
+Starting from v24.2, topic-aware intra-broker partition balancing allows for dynamically reassigning partitions within a broker. If a broker's core count changes, when the broker starts back up, Redpanda can check partition assignments across the broker's cores, and reassign partitions so that a balanced assignment is maintained across all cores. Redpanda can also check partition assignments when partitions are removed from a broker, and rebalance the remaining partitions between cores.
```
micheleRP:
```suggestion
Starting in v24.2, topic-aware intra-broker partition balancing allows for dynamically reassigning partitions within a broker. If a broker's core count changes, when the broker starts back up, Redpanda can check partition assignments across the broker's cores and reassign partitions, so that a balanced assignment is maintained across all cores. Redpanda can also check partition assignments when partitions are removed from a broker, and rebalance the remaining partitions between cores.
```

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,35 @@ This requires an Enterprise license.
...
+
+Use the public metrics for CPU usage described in the xref:manage:monitoring.adoc#cpu-usage[Monitoring] guide, as well as the xref:reference:internal-metrics-reference.adoc#vectorized_reactor_utilization[`vectorized_reactor_utilization`] internal metric to determine when to use intra-broker partition balancing.
```
micheleRP:
```suggestion
To determine when to use intra-broker partition balancing, use the public metrics for CPU usage described in the xref:manage:monitoring.adoc#cpu-usage[Monitoring] guide, as well as the xref:reference:internal-metrics-reference.adoc#vectorized_reactor_utilization[`vectorized_reactor_utilization`] internal metric.
```

micheleRP (approved):
Made some minor text suggestions, otherwise looks good!

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,53 @@ This requires an Enterprise license.
...
+
+Prior to Redpanda version 24.2, this meant that some cores on a broker could inadvertently host many partitions of heavily-used topics and cause the CPU to be xref:manage:monitoring.adoc#cpu-usage[overburdened]. Additionally, when the partition rebalance moved some partitions away from a broker, the remaining partitions did not necessarily rebalance across the broker's cores. Or, if a broker's core count was increased, Redpanda did not assign any partitions to the new cores until new partitions were created or old partitions were moved out.
+
+Starting in v24.2, topic-aware intra-broker partition balancing allows for dynamically reassigning partitions within a broker. If a broker's core count changes, when the broker starts back up, Redpanda can check partition assignments across the broker's cores and reassign partitions, so that a balanced assignment is maintained across all cores. Redpanda can also check partition assignments when partitions are removed from a broker, and rebalance the remaining partitions between cores.
```
mattschumpert: @kbatuigas 
We have the word topic-aware, but we don't really say what it does anywhere.  A sentence about how we also try to distribute partition replicas of any given topic evenly across cores within in a node is probably warranted. here

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,53 @@ This requires an Enterprise license.
 This mode is not recommended for production clusters. Only set to `off` if you need to move partitions manually.
 |===
 
+== Intra-broker partition balancing
+
+[CAUTION]
+====
+This is a beta feature for v24.2 and is not recommended for use for production clusters.
```
mattschumpert: Maybe pedantic but probably good to explicitly say.  "This feature is disabled by default." before 'how to enable'

mattschumpert (approved):

modules/get-started/pages/whats-new.adoc
```diff
@@ -52,6 +52,12 @@ New tests are added to the xref:manage:cluster-maintenance/cluster-diagnostics.a
 
 See the xref:reference:rpk/rpk-cluster/rpk-cluster-self-test-status.adoc[`rpk self test`] reference for usage and output examples.
 
+== Intra-broker partition balancing (beta)
```
micheleRP:
```suggestion
== Intra-broker partition balancing: beta
```

modules/manage/pages/cluster-maintenance/cluster-balancing.adoc
```diff
@@ -112,6 +116,55 @@ This requires an Enterprise license.
...
+
+The feature is disabled by default. To unlock this feature, run this command with a superuser:
```
micheleRP: I'd move this line up to have 1 paragraph (instead of 2) in the Caution admonition. And I'd probably make it a Note instead of a Caution, unless PM requested otherwise.

PR #631, opened 2024-07-25
rpk references: add caution for user update

r-vasquez:
This is fixed already, but it's better to add the
caution for users that might be using older rpk
versions

## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2585

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up
- [X] Small fix (typos, links, copyedits, etc)

### DIFF
--- a/modules/reference/pages/rpk/rpk-security/rpk-security-user-update.adoc
+++ b/modules/reference/pages/rpk/rpk-security/rpk-security-user-update.adoc
@@ -4,6 +4,8 @@
 
 Update SASL user credentials
 
+CAUTION: The default value for the `--mechanism` flag is `SCRAM-SHA-256`. Specify the flag when using a different mechanism to avoid unexpected changes.
+
 == Usage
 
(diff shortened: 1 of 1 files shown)


modules/reference/pages/rpk/rpk-security/rpk-security-user-update.adoc
```diff
@@ -4,6 +4,8 @@
 
 Update SASL user credentials
 
+CAUTION: The default value for the `--mechanism` flag is `SCRAM-SHA-256`. Specify the flag when using a different mechanism to avoid unexpected changes.
```
Deflaimun:
```suggestion
CAUTION: The default value for the `--mechanism` flag is `SCRAM-SHA-256`. Set the flag when using a different mechanism to avoid unexpected changes.
```

Deflaimun (approved):

PR #635, opened 2024-07-26
document incompatible change in AlterConfigs behavior

micheleRP:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2604 
Review deadline: Monday, July 29

## Page previews
[Linux: Review incompatible changes](https://deploy-preview-635--redpanda-docs-preview.netlify.app/24.2/upgrade/rolling-upgrade/)
[Kubernetes: Review incompatible changes](https://deploy-preview-635--redpanda-docs-preview.netlify.app/24.2/upgrade/k-rolling-upgrade/) (duplicated here)

## Checks

- [X] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/upgrade/pages/rolling-upgrade.adoc
+++ b/modules/upgrade/pages/rolling-upgrade.adoc
@@ -19,15 +19,7 @@ include::partial$rolling-upgrades/upgrade-limitations.adoc[]
 * An understanding of the <<impact-of-broker-restarts,impact of broker restarts>> on clients, node CPU, and any alerting systems you use.
 * Review incompatible changes in new versions.
 
-=== Review incompatible changes
-
-* Starting in version 24.2, Redpanda removes topic-level configuration overrides for unspecified fields, instead of setting them to their type's default value. This aligns more closely with the AlterConfigs API. 
-+
-NOTE: An exception to this is made for the `redpanda.remote.` topic properties, such as `redpanda.remote.delete`. The remote properties are intentionally sticky to avoid the possibility of deleting data in Tiered Storage.
-
-* Starting in version 24.2, xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput limits] are compatible with the https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client[AlterClientQuotas and DescribeClientQuotas^] Kafka APIs. Redpanda determines client throughput limits on a per-broker basis. In earlier versions, client throughput quotas were applied from cluster configurations on a per-shard basis. 
-
-* Patch releases in https://github.com/redpanda-data/redpanda/discussions/9522[22.3.14^] and https://github.com/redpanda-data/redpanda/discussions/9523[23.1.2^] changed the behavior when remote read is disabled and the requested Raft term falls below the local log's beginning. In earlier versions, Redpanda returned an offset -1. With the patch, when you request a value older than the lowest offset, Redpanda returns the lowest offset, not -1.
+include::partial$incompat-changes.adoc[]
 
 == Find a new version
 
--- a/modules/upgrade/pages/rolling-upgrade.adoc
+++ b/modules/upgrade/pages/rolling-upgrade.adoc
@@ -21,9 +21,13 @@ include::partial$rolling-upgrades/upgrade-limitations.adoc[]
 
 === Review incompatible changes
 
-Starting in version 24.2, xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput limits] are compatible with the https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client[AlterClientQuotas and DescribeClientQuotas^] Kafka APIs. Redpanda determines client throughput limits on a per-broker basis. In earlier versions, client throughput quotas were applied from cluster configurations on a per-shard basis. 
+* Starting in version 24.2, Redpanda removes topic-level configuration overrides for unspecified fields, instead of setting them to their type's default value. This aligns more closely with the AlterConfigs API. 
++
+NOTE: An exception to this is made for the `redpanda.remote.` topic properties, such as `redpanda.remote.delete`. The remote properties are intentionally sticky to avoid the possibility of deleting data in Tiered Storage.
 
-Patch releases in https://github.com/redpanda-data/redpanda/discussions/9522[22.3.14^] and https://github.com/redpanda-data/redpanda/discussions/9523[23.1.2^] changed the behavior when remote read is disabled and the requested Raft term falls below the local log's beginning. In earlier versions, Redpanda returned an offset -1. With the patch, when you request a value older than the lowest offset, Redpanda returns the lowest offset, not -1.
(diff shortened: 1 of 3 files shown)


modules/upgrade/pages/rolling-upgrade.adoc
```diff
@@ -21,9 +21,13 @@ include::partial$rolling-upgrades/upgrade-limitations.adoc[]
 
 === Review incompatible changes
 
-Starting in version 24.2, xref:manage:cluster-maintenance/manage-throughput.adoc#client-throughput-limits[client throughput limits] are compatible with the https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client[AlterClientQuotas and DescribeClientQuotas^] Kafka APIs. Redpanda determines client throughput limits on a per-broker basis. In earlier versions, client throughput quotas were applied from cluster configurations on a per-shard basis. 
+* Starting in version 24.2, Redpanda removes topic-level configuration overrides for unspecified fields, instead of setting them to their type's default value. This aligns more closely with the AlterConfigs API.
```
mattschumpert: @micheleRP THis is inaccurate.  We still have 'topic-level configuration overrides for unspecified fields' in RPK cluster config set, Ansible, Kubernetes, Console, etc. etc. as they use a different API. We are certainly not removing this capability from Redpanda

We are ONLY changing the behavior of the AlterConfigs KAFKA API  to more closely align with the behavior in Apache Kafka.

mattschumpert (changes requested):

mattschumpert (approved):

twmb (approved):

modules/upgrade/partials/incompat-changes.adoc
```diff
@@ -0,0 +1,9 @@
+=== Review incompatible changes
+
+* Starting in version 24.2, when managing configuration properties using the AlterConfigs API directly or the `kafka-configs.sh` shell script without incremental flags, Redpanda now resets all unspecified values to their default values. This aligns more closely with the behavior in Apache Kafka. There is no change if you're managing configuration with tools like `rpk`, Redpanda Console, Kubernetes, Helm, or Terraform.
```
Feediver1:
```suggestion
* Starting in version 24.2, when managing configuration properties using the AlterConfigs API directly, or the `kafka-configs.sh` shell script without incremental flags, Redpanda resets all unspecified values to the default values. This aligns more closely with the behavior in Apache Kafka. There is no change if you're managing your configuration with tools like `rpk`, Redpanda Console, Kubernetes, Helm, or Terraform. 
```

modules/upgrade/partials/incompat-changes.adoc
```diff
@@ -0,0 +1,9 @@
+=== Review incompatible changes
+
+* Starting in version 24.2, when managing configuration properties using the AlterConfigs API directly or the `kafka-configs.sh` shell script without incremental flags, Redpanda now resets all unspecified values to their default values. This aligns more closely with the behavior in Apache Kafka. There is no change if you're managing configuration with tools like `rpk`, Redpanda Console, Kubernetes, Helm, or Terraform. 
++
+NOTE: An exception is made for the `redpanda.remote.` topic properties, such as `redpanda.remote.delete`. The remote properties are intentionally sticky to avoid the possibility of disabling Tiered Storage.
+
```
Feediver1:
```suggestion
NOTE: An exception occurs for `redpanda.remote.` topic properties, such as `redpanda.remote.delete`. The remote properties are intentionally sticky to avoid the possibility of disabling Tiered Storage.
```

Feediver1: I'd avoid the use of the term "sticky" here--developer slang.

Feediver1 (approved):
See feedback

modules/upgrade/partials/incompat-changes.adoc
```diff
@@ -0,0 +1,9 @@
+=== Review incompatible changes
+
+* Starting in version 24.2, when managing configuration properties using the AlterConfigs API directly, or the `kafka-configs.sh` shell script without incremental flags, Redpanda resets all unspecified values to the default values. This aligns more closely with the behavior in Apache Kafka. There is no change if you're managing your configuration with tools like `rpk`, Redpanda Console, Kubernetes, Helm, or Terraform. 
++
+NOTE: An exception is made for the `redpanda.remote.` topic properties, such as `redpanda.remote.delete`. The remote properties are intentionally sticky to avoid the possibility of disabling Tiered Storage.
```
michael-redpanda: > The remote properties are intentionally sticky to avoid the possibility of disabling Tiered Storage.

I believe the danger is that disabling these may result in the data in S3 for those topics may get deleted, am I correct here @mattschumpert ?

michael-redpanda (approved):

modules/upgrade/partials/incompat-changes.adoc
mattschumpert: Disabling may or may not actually delete the data immediately (I dont think it will) but it can certainly lead to filling disks as the retention setting now gets applied to local storage,  and  to storing topic data on storage devices that are meant to be ephemeral.

@micheleRP 

I think we can just say  something like "These properties are intentionally sticky to avoid the possibility of _unintentionally_ disabling Tiered Storage for the topic, which can cause significant operational issues in clusters designed/sized to use Tiered Storage exclusively and where object storage is the only persistent/durable storage tier"

micheleRP: Rephrased note with Matt: 
This does not pertain to the `redpanda.remote.` topic properties, such as `redpanda.remote.delete`. The remote properties are not reset to their defaults by the AlterConfigs API to avoid the possibility of unintentionally disabling Tiered Storage for a topic, which could cause significant operational issues for clusters designed to use Tiered Storage exclusively or with object storage as the only durable storage tier.

modules/upgrade/partials/incompat-changes.adoc
```diff
@@ -0,0 +1,9 @@
+=== Review incompatible changes
+
+* Starting in version 24.2, when managing configuration properties using the AlterConfigs API directly, or the `kafka-configs.sh` shell script without incremental flags, Redpanda resets all unspecified values to the default values. This aligns more closely with the behavior in Apache Kafka. There is no change if you're managing your configuration with tools like `rpk`, Redpanda Console, Kubernetes, Helm, or Terraform.
```
pgellert, line 3: > or the `kafka-configs.sh` shell script without incremental flags,

This part is incorrect as far as I am aware. I've been testing with`kafka-configs.sh` locally and it has been using the IncrementalAlterConfigs API and not the AlterConfigs API since at least version 2.5.0 and there is no incremental flag that I am aware of. Pre-2.5.0 versions require a zookeeper endpoint for topic config updates, so they can't be used with redpanda. So whoever's managing redpanda topic configs with `kafka-configs.sh` is using the IncrementalAlterConfigs API.

I am curious where the quoted part came from, but I think we should remove it from this section.

micheleRP: Thanks @pgellert! I can get this updated. 
cc @mattschumpert

mattschumpert: Ahh OK. @pgellert  I swear @twmb and I saw a version of this .sh with some --inc or --incremental flags and @twmb was the one saying at first that kafka-config.sh uses AlterConfigs.  SO would request @twmb to clarify here

twmb: My knowledge for this comes from https://cwiki.apache.org/confluence/display/KAFKA/KIP-1011%3A+Use+incrementalAlterConfigs+when+updating+broker+configs+by+kafka-configs.sh

However, I didn't read closely enough -- it looks like modifying only _broker_ configs uses the old AlterConfigs, and modifying anything else does in fact use IncrementalAlterConfigs. My mistake.

PR #669, opened 2024-08-05
k-redpanda-helm-spec: cpu.overprovisioned

ligfx:
**Old:**

Screenshot 2024-08-05 at 3 17 57 PM

**New:**

Screenshot 2024-08-05 at 3 18 24 PM


ligfx:
Closing as these are generated dynamically from the Helm repo and would get overwritten next sync.

PR #698, opened 2024-08-21
Add Nodewatcher warnings

JakeSCahill:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2668
Resolves https://github.com/redpanda-data/documentation-private/issues/2680

Review deadline: August 26

## Page previews

https://deploy-preview-698--redpanda-docs-preview.netlify.app/current/manage/kubernetes/k-scale-redpanda/#node-pvc

## Checks

- [ ] New feature
- [ ] Content gap
- [x] Support Follow-up


modules/manage/pages/kubernetes/k-scale-redpanda.adoc
```diff
@@ -282,3 +286,11 @@ kubectl get persistentvolumeclaim --namespace <namespace>
 kubectl get persistentvolume --namespace <namespace>
 ----
 
+After the Nodewatcher controller has finished, you have two options to maintain the stability of your Redpanda cluster:
+
+  - *Option 1:* xref:manage:kubernetes/k-decommission-brokers.adoc[Decommission the broker] that was removed from the node. This is necessary to prevent quorum loss and ensure cluster stability.
```
chrisseto: I don't know the answer off the top of my head but do inactive brokers need to be decommissioned with --force or is that only if they were holding partitions without any replicas?

modules/manage/pages/kubernetes/k-scale-redpanda.adoc
```diff
@@ -282,3 +286,11 @@ kubectl get persistentvolumeclaim --namespace <namespace>
...
+
+  - *Option 2:* Recover the node. The StatefulSet will automatically reschedule the broker on the node, maintaining the desired number of brokers in the cluster. This option helps to keep your cluster running at full capacity, but you should still monitor the situation closely and consider decommissioning brokers if node failures persist.
```
chrisseto: Recovering the node would be a bit of an involved process and further tricky due to having to deal with the statefulset's ordinal system. The easiest way would be to scale down the STS, manually bind the PV and PVC together, and then decommission the newly added node. We can document that processes but it's involved enough that it should be it's own change and section. WDYT?

modules/manage/pages/kubernetes/k-scale-redpanda.adoc
```diff
@@ -124,7 +124,9 @@ To scale in a Redpanda cluster, you must decommission the brokers that you want
...
+The Nodewatcher controller maintains cluster operation during node failures by managing the lifecycle of PersistentVolumes (PVs) and PersistentVolumeClaims (PVCs) for Redpanda clusters. When the controller detects that a Node resource is not available, it sets the reclaim policy of the PV to `Retain`, helping to prevent data loss. Concurrently, it orchestrates the deletion of the PVC, which allows the Redpanda broker that was previously running on the deleted worker node to be rescheduled onto new, operational nodes.
+
+CAUTION: The Nodewatcher controller is an emergency backstop to keep your Redpanda cluster running in case of unexpected node failures. *Never use this controller as a routine method for removing brokers.* If you need to remove a broker, always follow the xref:manage:kubernetes/k-decommission-brokers.adoc[decommission process] to properly scale in your cluster.
```
chrisseto: Do you think it would be beneficial to note why this shouldn't be used for routine removal?
Should we call out that cluster auto scalers that affect redpanda Pods would be considered routine removal?

JakeSCahill: We don't support autoscalers, but I can clarify why it's not for routine removal.

modules/manage/pages/kubernetes/k-scale-redpanda.adoc
JakeSCahill: Tracking here: https://github.com/redpanda-data/documentation-private/issues/2694

JakeSCahill: When this doc is available, we should update this Nodewatcher section to link out to how to recover the broker.

modules/manage/pages/kubernetes/k-scale-redpanda.adoc
JakeSCahill: Correct 👍 I'll clarify in the docs.

vuldin (approved):
LGTM :+1:

vuldin (approved):
LGTM

micheleRP (approved):
lgtm

modules/manage/pages/kubernetes/k-scale-redpanda.adoc
```diff
@@ -282,3 +293,9 @@ kubectl get persistentvolumeclaim --namespace <namespace>
 kubectl get persistentvolume --namespace <namespace>
 ----
 
+After the Nodewatcher controller has finished, xref:manage:kubernetes/k-decommission-brokers.adoc[decommission the broker] that was removed from the node. This is necessary to prevent quorum loss and ensure cluster stability.
+
+NOTE: Make sure to use the `--force` flag when decommissioning the broker with xref:reference:rpk/rpk-redpanda/rpk-redpanda-admin-brokers-decommission.adoc[`rpk redpanda admin brokers decommission`]. This flag is required when the broker is no longer running.
```
chrisseto, line 298: I don't know the ins and outs of `--force`, could there be other ramifications here? For example if there's a topic with RF=1 that had a partition on the lost node does this opt into data loss or is that a different ball game? Ignore more if you know this is safe :)

modules/manage/pages/kubernetes/k-scale-redpanda.adoc
```diff
@@ -282,3 +293,9 @@ kubectl get persistentvolumeclaim --namespace <namespace>
 kubectl get persistentvolume --namespace <namespace>
 ----
 
+After the Nodewatcher controller has finished, xref:manage:kubernetes/k-decommission-brokers.adoc[decommission the broker] that was removed from the node. This is necessary to prevent quorum loss and ensure cluster stability.
```
chrisseto: I was reviewing my knowledge of the ghost broker situation and there's a missing link in my knowledge about exactly how quorum loss happens. Would it be worth nuancing this with something like "potential loss of quorum" or "eventual loss of quorum"?

chrisseto (approved):

modules/manage/pages/kubernetes/k-scale-redpanda.adoc
JakeSCahill:
```suggestion
After the Nodewatcher controller has finished, xref:manage:kubernetes/k-decommission-brokers.adoc[decommission the broker] that was removed from the node. This is necessary to prevent a potential loss of quorum and ensure cluster stability.
```

modules/manage/pages/kubernetes/k-scale-redpanda.adoc
r-vasquez: Just chiming in, `--force` was added to avoid the safety checks performed in older clusters:

> For safety on v22.x clusters, this command will not run if the requested 
broker is in maintenance mode. As of v23.x, Redpanda supports 
decommissioning a node that is currently in maintenance mode. If you are on 
a v22.x cluster and need to bypass the maintenance mode check (perhaps your 
cluster is unreachable), use the hidden --force flag.

More info here: https://github.com/redpanda-data/redpanda/commit/85c05713a8bcf148a5fd2c05c598748ee597dc92

JakeSCahill: Thanks Rogger. So that flag isn’t required for versions after 22.x?

r-vasquez: It shouldn't be. But there are scenarios (like a cluster is unreachable) where you still want to use this.

Cc: @daisukebe since you know more about the edge cases when using this command :+1:

JakeSCahill: Since we already document this flag, we can add follow-up info about edge cases/ramifications. I'll merge this one for now.

daisukebe: > This flag is required when the broker is no longer running.

This is still true in the latest release of Redpanda. When a broker is dead and we want to decommission it, the force flag is required. Read the Nodewatcher doc and it seems like the case.

PR #725, opened 2024-08-27
DOC-287 Mountable TS topics

kbatuigas:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2504
Related: Migrations [API reference](https://github.com/redpanda-data/docs/pull/798)

Review deadline: 9 Oct.

## Page previews

24.3 beta: [Mountable Topics](https://deploy-preview-725--redpanda-docs-preview.netlify.app/24.3/manage/mountable-topics/)

## Checks

- [x] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/manage/pages/mountable-topics.adoc
+++ b/modules/manage/pages/mountable-topics.adoc
@@ -1,0 +1,7 @@
+= Mountable Topics
+:description: Relocate a topic between your Redpanda cluster and object storage.
+:page-context-links: [{"name": "Linux", "to": "manage:mountable-topics.adoc" } ]
+:page-categories: Management
+:env-linux: true
+
+include::manage:partial$mountable-topics.adoc[]
--- a/modules/manage/pages/mountable-topics.adoc
+++ b/modules/manage/pages/mountable-topics.adoc
@@ -1,5 +1,5 @@
 = Mountable Topics
-:description: Relocate a topic between your Redpanda cluster and object storage.
+:description: Safely attach and detach Tiered Storage topics to and from a Redpanda cluster.
 :page-context-links: [{"name": "Linux", "to": "manage:mountable-topics.adoc" } ]
 :page-categories: Management
 :env-linux: true
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -11,8 +11,8 @@ The following data is moved to object storage when unmounting a topic from a clu
 
 == Prerequisites
 
-. Enable Tiered Storage for specific topics, or for the entire cluster (all topics).
-. Install `rpk`, or ensure that you have access to the Admin API.
+. Enable xref:manage:tiered-storage.adoc[Tiered Storage] for specific topics, or for the entire cluster (all topics).
+. xref:get-started:rpk-install.adoc[Install `rpk`], or ensure that you have access to the Admin API.
 
 == Unmount a topic from a cluster to object storage
 
@@ -32,13 +32,10 @@ rpk topic unmount <namespace>/<topic-name>
 Admin API::
 +
 --
-To unmount a topic from a cluster using the Admin API, you must first create an outbound topic migration. The API returns the migration ID upon creation. Then, you make an API request with the migration ID to carry out the migration in steps: prepare, execute, and cut over.
+To unmount a topic from a cluster using the Admin API, you must first create an outbound topic migration. The API returns the migration ID upon creation. Then, you make an API request with the migration ID to carry out the migration in steps: prepare, execute, and cut over. You must wait for each step to complete and for the <<monitor-progress,migration state>> to update before proceeding to the next step.
 
-[[unmount-create-migration]]
-=== Create migration
-
-Make a PUT request to the `/migrations` endpoint to create a topic migration. Specify the names of the desired topics in the request body:
-
+. Make a PUT request to the `/migrations` endpoint to create a topic migration. Specify the names of the desired topics in the request body:
++
 ```
 curl -X PUT http://localhost:9644/v1/migrations -d {
   "migration_type":"outbound", 
@@ -58,41 +55,32 @@ curl -X PUT http://localhost:9644/v1/migrations -d {
   ], "consumer_groups": []
 }
 ```
-
-* `ns` is the topic namespace. This field is optional. Default: `kafka`
++
+* `ns` is the topic namespace. This field is optional. Default value: `kafka`
 * `consumer_groups` is required.
-
++
 If successful, the response returns the ID of the newly-created migration. Use the ID in subsequent API calls to run the migration.
 
-[[unmount-prepare-migration]]
-=== Prepare migration
-
-When preparing topics for migration, Redpanda copies all possible topic data, including topic manifest,s to the destination bucket or container in object storage. In this state, the cluster still accepts client reads and writes. 
-
+. When preparing topics for migration, Redpanda copies all possible topic data including topic manifests to the destination bucket or container in object storage. In this state, the cluster still accepts client reads and writes. 
++
 Make a `POST /v1/migrations/\{id}/?action=prepare` request:
-
++
 ```
 curl -X POST http://localhost:9644/v1/<migration-id>/?action=prepare
 ```
 
-[[unmount-execute-migration]]
-=== Execute migration
-
-When you execute the migration, the cluster rejects client reads and writes. Redpanda uploads any remaining topic data that has not yet been copied to object storage. 
-
+. When you execute the migration, the cluster rejects client reads and writes. Redpanda uploads any remaining topic data that has not yet been copied to object storage. 
++
 Make a `POST /v1/migrations/\{id}/?action=execute` request to execute the migration:
-
++
 ```
 curl -X POST http://localhost:9644/v1/<migration-id>/?action=execute
 ```
 
-[[unmount-cut-over]]
-=== Cut over
-
-Finally, the migration can be deleted. Redpanda deletes the unmounted topics from the cluster.
-
+. Finally, the migration can be deleted. Redpanda deletes the unmounted topics from the cluster.
++
 To cut over after executing the migration:
-
++
 ```
 curl -X POST http://localhost:9644/v1/<migration-id>/?action=finish
 ```
@@ -123,13 +111,10 @@ rpk topic mount <namespace>/<source-topic-name> --to <namespace>/<new-topic-name
 Admin API::
 +
 --
-To mount a topic to a target cluster using the Admin API, you must first create an inbound topic migration. The API returns the migration ID upon creation. Then, you make an API request with the migration ID to carry out the migration in steps: prepare, execute, and cut over.
+To mount a topic to a target cluster using the Admin API, you must first create an inbound topic migration. The API returns the migration ID upon creation. Then, you make an API request with the migration ID to carry out the migration in steps: prepare, execute, and cut over. You must wait for each step to complete and for the <<monitor-progress,migration state>> to update before proceeding to the next step.
 
-[[mount-create-migration]]
-=== Create migration
-
-Make a PUT request to the `/migrations` endpoint to create a topic migration. Specify the names of the topics in the request body:
-
+. Make a PUT request to the `/migrations` endpoint to create a topic migration. Specify the names of the topics in the request body:
++
 ```
 curl -X PUT http://localhost:9644/v1/migrations -d {
   "migration_type":"inbound", 
@@ -149,42 +134,33 @@ curl -X PUT http://localhost:9644/v1/migrations -d {
   "consumer_groups": []
 }
 ```
-
-* `ns` is the topic namespace. This field is optional. Default: `kafka`
++
+* `ns` is the topic namespace. This field is optional. Default value: `kafka`
 * To rename the topic in the target cluster, use the optional `alias` object in the request body. In the example, topics 1 and 3 are given new names in the target cluster, while topic 2 retains its original name.
 * `consumer_groups` is required.
-
++
 If successful, the response returns the ID of the newly-created migration. Use the ID in subsequent API calls to run the migration.
 
-[[mount-prepare-migration]]
-=== Prepare migration
-
-When preparing a topic for migration, Redpanda recreates the topic in a disabled state in the target cluster. The cluster allocates partitions but does not add log segments yet. Topic metadata is populated from the topic manifest found in object storage.
-
+. When preparing a topic for migration, Redpanda recreates the topic in a disabled state in the target cluster. The cluster allocates partitions but does not add log segments yet. Topic metadata is populated from the topic manifest found in object storage.
++
 Make a `POST /v1/migrations/\{id}/?action=prepare` request to prepare the migration:
-
++
 ```
 curl -X POST http://localhost:9644/v1/<migration-id>/?action=prepare
 ```
 
-[[mount-execute-migration]]
-=== Execute migration
-
-When you execute the migration, the target cluster starts the partitions with log segments downloaded from object storage.
-
+. When you execute the migration, the target cluster starts the partitions with log segments downloaded from object storage.
++
 Make a `POST /v1/migrations/\{id}/?action=execute` request to execute the migration:
-
++
 ```
 curl -X POST http://localhost:9644/v1/<migration-id>/?action=execute
 ```
 
-[[mount-cut-over]]
-=== Cut over
-
-Finally, the migration can be deleted. The target cluster starts to handle produce and consume workloads.
-
+. Finally, the migration can be deleted. The target cluster starts to handle produce and consume workloads.
++
 To cut over after executing the migration:
-
++
 ```
 curl -X POST http://localhost:9644/v1/<migration-id>/?action=finish
 ```
@@ -193,11 +169,35 @@ curl -X POST http://localhost:9644/v1/<migration-id>/?action=finish
 
 ======
 
+== Cancel a migration
+
+You can cancel a topic migration by running this command:
+
+```
+
+```
+
+You can only cancel a migration that is not in the following final <<monitor-progress,states>>:
+
+- `cut_over`
+- `finished`
+
 == Monitor progress
 
-Issue a GET request to the `/migrations` endpoint to view the status of topic migrations:
+Issue a GET request to the `/migrations` endpoint to view the state of topic migrations:
 
 ```
 curl http://localhost:9644/v1/migrations 
 ```
 
+The response returns the state of existing migrations:
+
+- `planned`
+- `preparing`
+- `prepared`
+- `executing`
+- `executed`
+- `cut_over`
+- `finished`
+- `canceling`
+- `cancelled`
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -1,0 +1,57 @@
+For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This effectively frees up unused partition space that can be reclaimed, enables you to migrate a topic to a different cluster, and allows the topic or even the entire cluster to hibernate or be decommissioned.
+
+You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
+
+== Prerequisites
+
+. Enable Tiered Storage for specific topics, or for the entire cluster (all topics).
+. Install `rpk`, or ensure that you have access to the Admin API.
+
+
+== Mount a topic to a cluster
+
+In your target cluster, run this command to mount a topic from object storage:
+
+[tabs]
+====
+rpk::
++
+--
+```
+rpk topic mount <topic-name> -u <cloud-storage-url>
+```
+--
+Admin API::
++
+--
+```
+
+```
+--
+
+====
+
+== Unmount a topic from a cluster to object storage
+
+Run this command to flush a topic from the source cluster to object storage. This deletes the topic from the cluster. Once you initiate this process, all incoming writes to the topic are blocked. Producers and consumers of the topic receive an error message indicating that the topic is no longer available. 
+
+[tabs]
+====
+rpk::
++
+--
+```
+rpk topic unmount <topic-name>
+```
+--
+Admin API::
++
+--
+```
+
+```
+--
+
+====
+
+== Monitor progress
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -2,6 +2,10 @@ For topics with Tiered Storage enabled, you can mount and unmount topics to tran
 
 You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
 
+The following data is moved to object storage when unmounting a topic from a cluster:
+* Topic definitions. 
+* The consumer offsets topic. This allows the destination cluster to also restore consumer group state.
+
 == Prerequisites
 
 . Enable Tiered Storage for specific topics, or for the entire cluster (all topics).
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -3,6 +3,7 @@ For topics with Tiered Storage enabled, you can mount and unmount topics to tran
 You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
 
 The following data is moved to object storage when unmounting a topic from a cluster:
+
 * Topic definitions. 
 * The consumer offsets topic. This allows the destination cluster to also restore consumer group state.
 
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -174,7 +174,7 @@ curl -X POST http://localhost:9644/v1/<migration-id>/?action=finish
 You can cancel a topic migration by running this command:
 
 ```
-
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=cancel
 ```
 
 You can only cancel a migration that is not in the following final <<monitor-progress,states>>:
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -1,5 +1,7 @@
 For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This effectively frees up unused partition space that can be reclaimed, enables you to migrate a topic to a different cluster, and allows the topic or even the entire cluster to hibernate or be decommissioned.
 
+An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic.
+
 You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
 
 The following data is moved to object storage when unmounting a topic from a cluster:
@@ -12,51 +14,190 @@ The following data is moved to object storage when unmounting a topic from a clu
 . Enable Tiered Storage for specific topics, or for the entire cluster (all topics).
 . Install `rpk`, or ensure that you have access to the Admin API.
 
+== Unmount a topic from a cluster to object storage
+
+When you unmount a topic, all incoming writes to the topic are blocked as Redpanda migrates the topic from the cluster to object storage. Producers and consumers of the topic receive an error message indicating that the topic is no longer available. The unmounted topic is deleted in the source cluster.
+
+[tabs]
+======
+rpk::
++
+--
+In your cluster, run this command to unmount a topic to object storage:
+
+```
+rpk topic unmount <namespace>/<topic-name>
+```
+--
+Admin API::
++
+--
+To unmount a topic from a cluster using the Admin API, you must first create an outbound topic migration. The API returns the migration ID upon creation. Then, you make an API request with the migration ID to carry out the migration in steps: prepare, execute, and cut over.
+
+[[unmount-create-migration]]
+=== Create migration
+
+Make a PUT request to the `/migrations` endpoint to create a topic migration. Specify the names of the desired topics in the request body:
+
+```
+curl -X PUT http://localhost:9644/v1/migrations -d {
+  "migration_type":"outbound", 
+  "topics": [
+    {
+      "ns": "kafka", 
+      "topic": "<topic-1-name>"
+    }, 
+    {
+      "ns": "kafka", 
+      "topic": "<topic-2-name>"
+    }, 
+    {
+      "ns": "kafka", 
+      "topic": "<topic-3-name>"
+    }
+  ], "consumer_groups": []
+}
+```
+
+* `ns` is the topic namespace. This field is optional. Default: `kafka`
+* `consumer_groups` is required.
+
+If successful, the response returns the ID of the newly-created migration. Use the ID in subsequent API calls to run the migration.
+
+[[unmount-prepare-migration]]
+=== Prepare migration
+
+When preparing topics for migration, Redpanda copies all possible topic data, including topic manifest,s to the destination bucket or container in object storage. In this state, the cluster still accepts client reads and writes. 
+
+Make a `POST /v1/migrations/\{id}/?action=prepare` request:
+
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=prepare
+```
+
+[[unmount-execute-migration]]
+=== Execute migration
+
+When you execute the migration, the cluster rejects client reads and writes. Redpanda uploads any remaining topic data that has not yet been copied to object storage. 
+
+Make a `POST /v1/migrations/\{id}/?action=execute` request to execute the migration:
+
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=execute
+```
+
+[[unmount-cut-over]]
+=== Cut over
+
+Finally, the migration can be deleted. Redpanda deletes the unmounted topics from the cluster.
+
+To cut over after executing the migration:
+
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=finish
+```
+
+--
+======
+
 
 == Mount a topic to a cluster
 
+[tabs]
+======
+rpk::
++
+--
 In your target cluster, run this command to mount a topic from object storage:
 
-[tabs]
-====
-rpk::
-+
---
 ```
-rpk topic mount <topic-name> -u <cloud-storage-url>
+rpk topic mount <source-topic-name>
+```
+
+You can also rename the topic as you mount it to the target cluster:
+
+```
+rpk topic mount <namespace>/<source-topic-name> --to <namespace>/<new-topic-name>
 ```
 --
 Admin API::
 +
 --
-```
+To mount a topic to a target cluster using the Admin API, you must first create an inbound topic migration. The API returns the migration ID upon creation. Then, you make an API request with the migration ID to carry out the migration in steps: prepare, execute, and cut over.
+
+[[mount-create-migration]]
+=== Create migration
+
+Make a PUT request to the `/migrations` endpoint to create a topic migration. Specify the names of the topics in the request body:
 
 ```
+curl -X PUT http://localhost:9644/v1/migrations -d {
+  "migration_type":"inbound", 
+  "topics": [
+    {
+      "source_topic": {"ns": "kafka", "topic": "<source-topic-1-name>"}, 
+      "alias": {"ns": "kafka", "topic": "<new-topic-1-name>"}
+    }, 
+    {
+      "source_topic": {"ns": "kafka", "topic": "<source-topic-2-name>"}
+    }, 
+    {
+      "source_topic": {"ns": "kafka", "topic": "source-topic-3-name"}, 
+      "alias": {"ns": "kafka", "topic": "<new-topic-3-name>"}
+    }
+  ], 
+  "consumer_groups": []
+}
+```
+
+* `ns` is the topic namespace. This field is optional. Default: `kafka`
+* To rename the topic in the target cluster, use the optional `alias` object in the request body. In the example, topics 1 and 3 are given new names in the target cluster, while topic 2 retains its original name.
+* `consumer_groups` is required.
+
+If successful, the response returns the ID of the newly-created migration. Use the ID in subsequent API calls to run the migration.
+
+[[mount-prepare-migration]]
+=== Prepare migration
+
+When preparing a topic for migration, Redpanda recreates the topic in a disabled state in the target cluster. The cluster allocates partitions but does not add log segments yet. Topic metadata is populated from the topic manifest found in object storage.
+
+Make a `POST /v1/migrations/\{id}/?action=prepare` request to prepare the migration:
+
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=prepare
+```
+
+[[mount-execute-migration]]
+=== Execute migration
+
+When you execute the migration, the target cluster starts the partitions with log segments downloaded from object storage.
+
+Make a `POST /v1/migrations/\{id}/?action=execute` request to execute the migration:
+
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=execute
+```
+
+[[mount-cut-over]]
+=== Cut over
+
+Finally, the migration can be deleted. The target cluster starts to handle produce and consume workloads.
+
+To cut over after executing the migration:
+
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=finish
+```
+
 --
 
-====
-
-== Unmount a topic from a cluster to object storage
-
-Run this command to flush a topic from the source cluster to object storage. This deletes the topic from the cluster. Once you initiate this process, all incoming writes to the topic are blocked. Producers and consumers of the topic receive an error message indicating that the topic is no longer available. 
-
-[tabs]
-====
-rpk::
-+
---
-```
-rpk topic unmount <topic-name>
-```
---
-Admin API::
-+
---
-```
-
-```
---
-
-====
+======
 
 == Monitor progress
+
+Issue a GET request to the `/migrations` endpoint to view the status of topic migrations:
+
+```
+curl http://localhost:9644/v1/migrations 
+```
+
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -206,7 +206,7 @@ You cannot cancel mount and unmount operations in the following <<monitor-progre
 
 == Additional considerations
 
-It is not currently possible to unmount a topic whose name matches multiple topics in the origin cluster.
+It is not possible to unmount a topic whose name matches multiple topics in the origin cluster.
 
 Redpanda prevents you from mounting the same topic to multiple clusters at once. This ensures that multiple clusters don't write to the same location in object storage and corrupt the topic.
 
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -2,14 +2,12 @@ For topics with Tiered Storage enabled, you can mount and unmount topics to tran
 
 Redpanda also transfers topic definitions when mounting or unmounting a topic.
 
-An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic.
-
-You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
+An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic. You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
 
 == Prerequisites
 
-. Enable xref:manage:tiered-storage.adoc[Tiered Storage] for specific topics, or for the entire cluster (all topics).
 . xref:get-started:rpk-install.adoc[Install `rpk`], or ensure that you have access to the Admin API.
+. Enable xref:manage:tiered-storage.adoc[Tiered Storage] for specific topics, or for the entire cluster (all topics).
 
 == Unmount a topic from a cluster to object storage
 
@@ -103,6 +101,37 @@ curl -X POST http://localhost:9644/v1/topics/mount -d {
 
 When the migration is complete, the target cluster handles produce and consume workloads for the topics.
 
+== Cancel a migration
+
+You can cancel a topic migration by running the command:
+
+[tabs]
+======
+rpk::
++
+--
+```
+rpk cluster storage cancel-mount <migration-id>
+```
+--
+
+Admin API::
++
+--
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=cancel
+```
+--
+======
+
+You cannot cancel migrations in the following <<monitor-progress,states>>:
+
+- `planned` (but you may still xref:api:ROOT:admin-api.adoc#delete-/v1/migrations/-id-[delete] a planned migration)
+- `cut_over`
+- `finished`
+- `canceling`
+- `cancelled`
+
 == Monitor progress
 
 [tabs]
@@ -171,40 +200,9 @@ The response returns the IDs and state of existing migrations:
 
 |===
 
-== Cancel a migration
-
-You can cancel a topic migration by running the command:
-
-[tabs]
-======
-rpk::
-+
---
-```
-rpk cluster storage cancel-mount <migration-id>
-```
---
-
-Admin API::
-+
---
-```
-curl -X POST http://localhost:9644/v1/<migration-id>/?action=cancel
-```
---
-======
-
-You can only cancel a migration that is not in the following <<monitor-progress,states>>:
-
-- `planned` (but you may still xref:api:ROOT:admin-api.adoc#delete-/v1/migrations/-id-[delete] a planned migration)
-- `cut_over`
-- `finished`
-- `canceling`
-- `cancelled`
-
 == Troubleshoot
 
-If Redpanda encounters errors during mounting and unmounting, it retries the operation indefinitely. If you see issues, make sure to check whether:
+If Redpanda encounters errors during mounting and unmounting, it retries the operation indefinitely. If you experience issues, check this list:
 
 * You are attempting to mount a topic that does not exist in object storage.
 * You are attempting to mount a topic that is already mounted to the current or a different cluster.
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -163,8 +163,8 @@ The response returns the IDs and state of existing migrations:
 | Migration state | Outbound migration (unmount) | Inbound migration (mount)
 
 | `planned`
-| Redpanda validates the migration definition.
-|
+2+| Redpanda validates the migration definition.
+
 
 | `preparing`
 | Redpanda copies all possible topic data including topic manifests to the destination bucket or container in object storage.
@@ -191,12 +191,11 @@ The response returns the IDs and state of existing migrations:
 | The migration is complete and then deleted. The target cluster starts to handle produce and consume workloads.
 
 | `canceling`
-| Redpanda is in the process of canceling the migration.
-|
+2+| Redpanda is in the process of canceling the migration.
+
 
 | `cancelled`
-| The migration is canceled.
-|
+2+| The migration is canceled.
 
 |===
 
--- a/modules/manage/partials/mountable-topics.adoc
+++ b/modules/manage/partials/mountable-topics.adoc
@@ -1,8 +1,8 @@
-For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This allows you to free up and reclaim unused partition space, or migrate a topic to a different cluster and hibernate or decommission the topic or even the entire cluster.
+For topics with Tiered Storage enabled, you can unmount a topic to detach segment data that is still on disk to object storage, and unmount a topic from object storage to attach the topic data to either the same origin cluster, or a different one. This allows you to hibernate a topic and free up and reclaim system resources taken up by the topic, or migrate a topic to a different cluster.
 
-Redpanda also transfers topic definitions when mounting or unmounting a topic.
+Redpanda also transfers topic manifests when mounting or unmounting a topic, so the topic can quickly accept reads and writes again and you can resume cluster workloads with ease. 
 
-An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic. You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
+An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic. You can mount a topic from object storage that has been previously unmounted from a source cluster to a different destination cluster, or back to the same source cluster.
 
 == Prerequisites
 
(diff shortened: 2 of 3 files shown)


modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,203 @@
+For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This effectively frees up unused partition space that can be reclaimed, enables you to migrate a topic to a different cluster, and allows the topic or even the entire cluster to hibernate or be decommissioned.
+
+An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic.
+
+You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
+
+The following data is moved to object storage when unmounting a topic from a cluster:
```
kbatuigas: Need SME confirmation on this list.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,203 @@
...
+    {
+      "ns": "kafka", 
+      "topic": "<topic-3-name>"
+    }
+  ], "consumer_groups": []
+}
+```
++
+* `ns` is the topic namespace. This field is optional. Default value: `kafka`
+* `consumer_groups` is required.
```
kbatuigas: Is there information we can provide on `consumer_groups`? As far as I know, mount/unmount for consumer groups is not implemented yet. Can we explain why the field is required to create the migration?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,203 @@
...
+--
+
+======
+
+== Cancel a migration
+
+You can cancel a topic migration by running this command:
+
+```
+
```
kbatuigas, line 177: Need SME to provide Admin API call. Is this available in rpk?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,203 @@
...
+The following data is moved to object storage when unmounting a topic from a cluster:
+
+* Topic definitions. 
+* The consumer offsets topic. This allows the destination cluster to also restore consumer group state.
+
+== Prerequisites
+
+. Enable xref:manage:tiered-storage.adoc[Tiered Storage] for specific topics, or for the entire cluster (all topics).
+. xref:get-started:rpk-install.adoc[Install `rpk`], or ensure that you have access to the Admin API.
+
```
kbatuigas, line 19: Are there any limitations that users should know about? For example, number of topics you can include in a migration? Amount of time that a topic can "hibernate" in object storage until it is mounted (although maybe that's more to do with their object storage configuration)?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,203 @@
...
+```
+
+```
+
+You can only cancel a migration that is not in the following final <<monitor-progress,states>>:
+
+- `cut_over`
+- `finished`
+
+== Monitor progress
```
kbatuigas: When or why might users encounter errors? Can and should they retry?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,203 @@
...
+--
+
+======
+
+== Cancel a migration
+
+You can cancel a topic migration by running this command:
+
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=cancel
```
kbatuigas, line 267: Need SME to confirm if this is available with rpk.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+Issue a GET request to the `/migrations` endpoint to view the state of topic migrations:
+
+```
+curl http://localhost:9644/v1/migrations 
+```
+
+The response returns the state of existing migrations:
+
+|===
+| Migration state | Outbound migration (unmount) | Inbound migration (mount)
```
kbatuigas: Can we fill in the missing descriptions in the [table](https://deploy-preview-725--redpanda-docs-preview.netlify.app/24.3/manage/mountable-topics/#monitor-progress)?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+| `cancelled`
+| The migration is canceled.
+|
+
+|===
+
+== Cancel a migration
+
+You can cancel a topic migration by running this command:
```
kbatuigas: Is this available in rpk?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+--
+
+======
+
+When the migration is complete, the target cluster handles produce and consume workloads for the topics.
+
+== Monitor progress
+
+Issue a GET request to the `/migrations` endpoint to view the state of topic migrations:
```
kbatuigas: Is this available in rpk?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
+For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This effectively frees up unused partition space that can be reclaimed, enables you to migrate a topic to a different cluster, and allows the topic or even the entire cluster to hibernate or be decommissioned.
+
+An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic.
+
+You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
+
+The following data is moved to object storage when unmounting a topic from a cluster:
+
+* Topic definitions. 
+* The consumer offsets topic. This allows the destination cluster to also restore consumer group state.
```
bashtanov: it has not been implemented yet

modules/manage/partials/mountable-topics.adoc
bashtanov: I'm not aware of any limitations like this. I've successfully ran it with as many topics and partitions as I had no problems to create.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+      "topic": "<topic-2-name>"
+    }, 
+    {
+      "topic": "<topic-3-name>"
+    }
+  ]
+}
+```
+
+You may optionally include the topic namespace (`ns`). Default value: `kafka`
```
bashtanov: also it is the only one supported so far

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+```
+
+The response returns the state of existing migrations:
+
+|===
+| Migration state | Outbound migration (unmount) | Inbound migration (mount)
+
+| `planned`
+| Redpanda has initialized a migration.
+|
```
bashtanov: in fact redpanda only has preliminarily validated the migration definition

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+| `preparing`
+| Redpanda copies all possible topic data including topic manifests to the destination bucket or container in object storage.
+| Redpanda recreates the topics in a disabled state in the target cluster. The cluster allocates partitions but does not add log segments yet. Topic metadata is populated from the topic manifests found in object storage.
+
+| `prepared` 
+| The migration is ready to execute. In this state, the cluster still accepts client reads and writes for the topics.
+|
```
bashtanov: topic exists in redpanda but cannot be used yet

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+| `preparing`
+| Redpanda copies all possible topic data including topic manifests to the destination bucket or container in object storage.
+| Redpanda recreates the topics in a disabled state in the target cluster. The cluster allocates partitions but does not add log segments yet. Topic metadata is populated from the topic manifests found in object storage.
+
+| `prepared` 
+| The migration is ready to execute. In this state, the cluster still accepts client reads and writes for the topics.
+| 
+
+| `executing` 
+| The cluster rejects client reads and writes for the topics. Redpanda uploads any remaining topic data that has not yet been copied to object storage.
```
bashtanov: also uncommitted transactions involving the topic are aborted

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+| The target cluster starts topic partitions with log segments downloaded from object storage.
```
bashtanov: not yet, the cluster just checks the topic data in tiered storage is not mounted by this or any other cluster

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+| `executed` 
+| All unmounted topic data from the cluster is available in object storage.
+| All mounted topic data is available in the target cluster.
```
bashtanov: Not yet. No big milestone to boast about here.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+| `cut_over`
+| The migration is complete and can be deleted. Redpanda deletes the unmounted topics from the cluster.
```
bashtanov: No, it is not complete and cannot be deleted yet. But it cannot be cancelled anymore. In fact what happens here is the cluster marks the data in tiered storage as available for further mounts. The topic metadata is also deleted from the cluster.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+| The migration is complete and can be deleted.
```
bashtanov: No, it is not complete and cannot be deleted yet. Redpanda marks the data in tiered storage as not available to be mounted by this or any other cluster.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+== Cancel a migration
+
+You can cancel a topic migration by running this command:
+
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=cancel
+```
+
+You can only cancel a migration that is not in the following final <<monitor-progress,states>>:
```
bashtanov: You also cannot cancel a migration in `planned` state (but you can delete it). Neither can you cancel a migration in `cancelled` or `canceling` state.

modules/manage/partials/mountable-topics.adoc
kbatuigas: @bashtanov I saw in the [rpk PR](https://github.com/redpanda-data/redpanda/pull/23575/files#diff-560be64b94cb849f004c7d3849bb5336f8ff1990e63ec988e8732280a373d8efR38) that a topic must have at least 3 partitions for it to be mounted from TS to a cluster, does that apply anywhere in this doc as well?

modules/manage/partials/mountable-topics.adoc
kbatuigas: > not yet, the cluster just checks the topic data in tiered storage is not mounted by this or any other cluster

Can `| The target cluster starts topic partitions with log segments downloaded from object storage.` be moved to the next row `executed`? If not I'll delete entirely

modules/manage/partials/mountable-topics.adoc
kbatuigas: > Not yet. No big milestone to boast about here.

I can leave this empty for the`Inbound migration` column but it would be nice to provide a short description

modules/manage/partials/mountable-topics.adoc
bashtanov: I never enforced or even heard about this restriction

modules/manage/partials/mountable-topics.adoc
bashtanov: I can think of the following situations:
- attempting to mount a topic that does not exist in the cloud storage
- attempting to mount a topic that is already mounted to this or another cluster
- any failures, such as tiered storage availability problems or multiple redpanda nodes going down.

All operations are retried indefinitely, so it's really unlikely that cancelling and restarting a migration would help. If there is any underlying problem fixing it should help without restarting.

modules/manage/partials/mountable-topics.adoc
bashtanov: https://www.acronymfinder.com/SME.html

modules/manage/partials/mountable-topics.adoc
bashtanov: Topic partitions are started in `preparing` state, but clients are not given access to it. Subsequent stages for a mount are only about mount manifest manipulation and eventually lifting these restrictions. I'm not sure whether log segments are physically downloaded on topic creation or when it is read from/written into, maybe storage team knows.

modules/manage/partials/mountable-topics.adoc
bashtanov: Each -ed state is only an indicator of the work in the correspondent -ing state completed. This applies to `cut_over`/`finished` pair as well. See https://github.com/redpanda-data/docs/pull/725#discussion_r1795571115 for what is done in `executing`. Stages are there not necessarily to do a lot of work in each of them, but rather to replicate the results between them in order to guard from race conditions in case cluster nodes or network between them goes down.

modules/manage/partials/mountable-topics.adoc
gene-redpanda: I saw that in the RFC for migrations, no idea if it is actually true as the testing at the time I wrote that was against a version of migrations that wasn't actually working.

modules/manage/partials/mountable-topics.adoc
kbatuigas: Sorry @bashtanov , I meant subject matter expert which would be Gene in this case for rpk! But the cancel operation is indeed available in rpk so I will go ahead and mark this as resolved.

bashtanov (approved):

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
+For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This effectively frees up unused partition space that can be reclaimed, enables you to migrate a topic to a different cluster, and allows the topic or even the entire cluster to hibernate or be decommissioned.
```
asimms41:
```suggestion
For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This allows you to free up and reclaim unused partition space, or migrate a topic to a different cluster and hibernate or decommission the topic or even the entire cluster.
```

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
...
+Redpanda also transfers topic definitions when mounting or unmounting a topic.
+
+An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic.
+
+You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
+
+== Prerequisites
+
+. Enable xref:manage:tiered-storage.adoc[Tiered Storage] for specific topics, or for the entire cluster (all topics).
+. xref:get-started:rpk-install.adoc[Install `rpk`], or ensure that you have access to the Admin API.
```
asimms41, line 10: As for previous PR, do you need to install `rpk` first?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
+For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This effectively frees up unused partition space that can be reclaimed, enables you to migrate a topic to a different cluster, and allows the topic or even the entire cluster to hibernate or be decommissioned.
+
+Redpanda also transfers topic definitions when mounting or unmounting a topic.
+
+An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic.
+
+You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
```
asimms41: Should this be in the same paragraph as the previous sentence?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+|===
+| Migration state | Outbound migration (unmount) | Inbound migration (mount)
+
+| `planned`
+| Redpanda validates the migration definition.
+|
+
+| `preparing`
+| Redpanda copies all possible topic data including topic manifests to the destination bucket or container in object storage.
```
asimms41:
```suggestion
| Redpanda copies all possible topic data, including topic manifests, to the destination bucket or container in object storage.
```

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
...
+| Redpanda is in the process of canceling the migration.
+|
+
+| `cancelled`
+| The migration is canceled.
+|
+
+|===
+
+== Cancel a migration
```
asimms41: Should this section be before the monitoring section?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
...
+Admin API::
++
+--
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=cancel
+```
+--
+======
+
+You can only cancel a migration that is not in the following <<monitor-progress,states>>:
```
asimms41: You _cannot_ cancel migrations in the following states:

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+- `planned` (but you may still xref:api:ROOT:admin-api.adoc#delete-/v1/migrations/-id-[delete] a planned migration)
+- `cut_over`
+- `finished`
+- `canceling`
+- `cancelled`
+
+== Troubleshoot
+
+If Redpanda encounters errors during mounting and unmounting, it retries the operation indefinitely. If you see issues, make sure to check whether:
```
asimms41:
```suggestion
If Redpanda encounters errors during mounting and unmounting, it retries the operation indefinitely. If experience issues, check this list:
```

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+* You are attempting to mount a topic that does not exist in object storage.
```
asimms41:
```suggestion
* Do not attempt to mount a topic that does not exist in object storage.
```

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
...
+* You are attempting to mount a topic that is already mounted to the current or a different cluster.
```
asimms41:
```suggestion
* Do not attempt to mount a topic that is already mounted to the current or a different cluster.
```

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,214 @@
...
+* Your Redpanda deployment is experiencing failures, such as Tiered Storage becoming unavailable, or multiple brokers going down.
```
asimms41:
```suggestion
* Make sure your Redpanda deployment is not experiencing failures, such as Tiered Storage becoming unavailable, or multiple brokers going down.
```

asimms41 (commented):
Just a few suggestions. Looks good!

gene-redpanda (approved):

modules/manage/pages/mountable-topics.adoc
```diff
@@ -0,0 +1,7 @@
+= Mountable Topics
+:description: Relocate a topic between your Redpanda cluster and object storage.
```
mattschumpert: description" Safely attach and detach tiered storage topics to and from a Redpanda cluster"

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
+For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This allows you to free up and reclaim unused partition space, or migrate a topic to a different cluster and hibernate or decommission the topic or even the entire cluster.
```
mattschumpert: First sentence is wrong.  this doesn't transfer any data except the small amount of data not yet uploaded due to the upload interval

By mounting, you attach topic data previously written to a Redpanda cluster and stored in cloud storage to a Redpanda cluster, potentially a different one.  This allows you to 'hibernate' a topic, freeing up system resources consumed by unused topic, or migrate a topic to a different cluster.

(we're not really supporting the 'hibernate whole cluster' yet here)

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+Redpanda also transfers topic definitions when mounting or unmounting a topic.
```
mattschumpert: What does 'definitions' mean?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic. You can mount a topic from object storage that has been previously flushed from a source cluster to a different destination cluster, or back to the same source cluster.
```
mattschumpert: "that has previously been unmounted from a source cluster ..."

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+== Prerequisites
+
+. xref:get-started:rpk-install.adoc[Install `rpk`], or ensure that you have access to the Admin API.
```
mattschumpert, line 10: This is not really true as in cloud the RPK command will not use the admin API but rather the cloud public API (these APIs are still being added but will be soon).

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+. Enable xref:manage:tiered-storage.adoc[Tiered Storage] for specific topics, or for the entire cluster (all topics).
+
+== Unmount a topic from a cluster to object storage
+
+When you unmount a topic, all incoming writes to the topic are blocked as Redpanda migrates the topic from the cluster to object storage. Producers and consumers of the topic receive an error message indicating that the topic is no longer available. The unmounted topic is deleted in the source cluster.
```
mattschumpert: 'unmounts the topic from the cluster to object storage'

we should not use the work migrate anywhere in this doc except if we're talking about moving between two clusters.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+```
+
+* `ns` is the topic namespace. This field is optional and only `kafka` is currently supported.
+* To rename a topic in the target cluster, use the optional `alias` object in the request body. In the example, topics 1 and 3 are given new names in the target cluster, while topic 2 retains its original name.
+
+--
+
+======
+
+When the migration is complete, the target cluster handles produce and consume workloads for the topics.
```
mattschumpert: When the mount is complete.  Do not use the word 'migration' anywhere here.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+== Cancel a migration
+
+You can cancel a topic migration by running the command:
```
mattschumpert: 'Cancel a mount or unmount operation'. (not migration)

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+[tabs]
+======
+rpk::
++
+--
+```
+rpk cluster storage cancel-mount <migration-id>
```
mattschumpert: We still use migration-id here for technical reasons but only because of implementation detail

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+- `cancelled`
+
+== Monitor progress
+
+[tabs]
+======
+rpk::
++
+--
+To list topic migrations, run the command:
```
mattschumpert: "list active mount & unmount operations"

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+```
+rpk cluster storage list-mount
+```
+--
+
+Admin API::
++
+--
+Issue a GET request to the `/migrations` endpoint to view the state of topic migrations:
```
mattschumpert: "view the status of topic mount and unmount operations"

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+```
+curl http://localhost:9644/v1/migrations 
+```
+--
+======
+
+The response returns the IDs and state of existing migrations:
```
mattschumpert: The response returns the IDs and state of existing mount/unmount operations ('migrations'):

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+|===
+| Migration state | Outbound migration (unmount) | Inbound migration (mount)
+
+| `planned`
+2+| Redpanda validates the migration definition.
+
+
+| `preparing`
+| Redpanda copies all possible topic data, including topic manifests, to the destination bucket or container in object storage.
```
mattschumpert: This is a bit misleading. For TS topics its not 'copying all the data' as most of the data including manifests are already in cloud storage.  

What's a better way to describe this preparing state from a user perspective @mmaslankaprv ?  Aren't we really 'flushing all data including topic manifests'...

mattschumpert: Sounds like this description was written for a migration between clusters vs a mount/unmount

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+| Redpanda recreates the topics in a disabled state in the target cluster. The cluster allocates partitions but does not add log segments yet. Topic metadata is populated from the topic manifests found in object storage.
+
+| `prepared` 
+| The migration is ready to execute. In this state, the cluster still accepts client reads and writes for the topics.
+| Topics exist in the cluster but clients do not yet have access to consume or produce.
+
+| `executing` 
+| The cluster rejects client reads and writes for the topics. Redpanda uploads any remaining topic data that has not yet been copied to object storage. Uncommitted transactions involving the topic are aborted.
```
mattschumpert, line 226: @mmaslankaprv  In the case of TS topics (where we are also just uploading data not yet uploaded), how do we differentiate this from the first step.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+| `cancelled`
+2+| The migration is canceled.
+
+|===
+
+== Troubleshoot
+
+If Redpanda encounters errors during mounting and unmounting, it retries the operation indefinitely. If you experience issues, check this list:
+
+* Do not attempt to mount a topic that does not exist in object storage.
```
mattschumpert: What happens in this case then @bashtanov .? We should explain.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+* Do not attempt to mount a topic that is already mounted to the current or a different cluster.
```
mattschumpert: @kbatuigas instead of just saying don't do this, let's explain how things work.  We have a built-in guardrail to prevent a topic from being mounted to two clusters at once to prevent multiple clusters writing to the same location in cloud storage and corrupting the topic. this is a feature.

cc @andrwng  for commentary

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+| `prepared` 
+| The migration is ready to execute. In this state, the cluster still accepts client reads and writes for the topics.
+| Topics exist in the cluster but clients do not yet have access to consume or produce.
+
+| `executing` 
+| The cluster rejects client reads and writes for the topics. Redpanda uploads any remaining topic data that has not yet been copied to object storage. Uncommitted transactions involving the topic are aborted.
+| The target cluster checks that the topic to be mounted has not already been mounted in any cluster.
+
+| `executed` 
+| All unmounted topic data from the cluster is available in object storage.
```
asimms41, line 230: Is there meant to be a description of the mounted topic as well?

asimms41 (approved):
Looks great. Just left one question but happy to approve.

modules/manage/partials/mountable-topics.adoc
kbatuigas: @mattschumpert I can't find the reference (an internal doc I think) anymore to where I came across this, but I understood it as topic configs. I see that I also used "topic manifests" later in the doc, I can rephrase with that instead, or just delete this line entirely if you prefer

modules/manage/partials/mountable-topics.adoc
kbatuigas: @mattschumpert @mmaslankaprv Is this "remaining data" that got in between when the unmount was started, until this point?

modules/manage/partials/mountable-topics.adoc
kbatuigas: @asimms41 there's [not a lot](https://github.com/redpanda-data/docs/pull/725#discussion_r1795574128) to describe. How about `The target cluster has verified that the topic has not already been mounted.` ?

modules/manage/partials/mountable-topics.adoc
kbatuigas: @mattschumpert @andrwng for my understanding - are target clusters writing to object storage during the mount?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+| `canceling`
+2+| Redpanda is in the process of canceling the operation.
+
+
+| `cancelled`
+2+| The operation is canceled.
+
+|===
+
+It is not currently possible to unmount a topic whose name matches multiple topics in the origin cluster.
```
kbatuigas: @mattschumpert I removed the "Troubleshoot" heading and whittled it down to just these few scenarios. From chatting with @bashtanov it sounds like there is additional work to do to enable the user to specify which topic or "incarnation" they want if they try to mount a topic with multiple matches on the name. And also some work to clarify log messages when running into issues. If there are other scenarios that have to be described here please let me know and we can get that into the next iteration of this doc.

modules/manage/partials/mountable-topics.adoc
andrwng: When a topic is unmounted it leaves behind an object (in code, called the topic mount manifest) indicating that it has been unmounted and is ready to be mounted. When it's mounted, we'll check if the mount manifest exists in object storage and then remove it, before proceeding with the mount. The intent here is to protect against against multiple mounts of the same topic, such that if a topic is already mounted, the mount manifest would no longer exist and a second mount would be rejected.

Deflaimun:
Should we ELI5 what exactly is "topic mount vs. topic unmount"? Or do we assume that our users are already familiar with the subject?

Something like:
"In Redpanda mountable topics are topics that [contains these properties]. You can make a topic mountable by doing X. To unmount a topic do Y.
(Are redpanda topics mountable by default? When they become mountable?) 
"

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+An unmounted topic in object storage is detached from all clusters. The original cluster releases ownership of the topic. You can mount a topic from object storage that has been previously unmounted from a source cluster to a different destination cluster, or back to the same source cluster.
+
+== Prerequisites
+
+. xref:get-started:rpk-install.adoc[Install `rpk`], or ensure that you have access to the Admin API.
+. Enable xref:manage:tiered-storage.adoc[Tiered Storage] for specific topics, or for the entire cluster (all topics).
+
+== Unmount a topic from a cluster to object storage
+
+When you unmount a topic, all incoming writes to the topic are blocked as Redpanda unmounts the topic from the cluster to object storage. Producers and consumers of the topic receive an error message indicating that the topic is no longer available. The unmounted topic is deleted in the source cluster.
```
Deflaimun: "The unmounted topic is deleted in the source cluster."
When does that happen? After the command is successfully ran? Is the command async? If yes, can I track its progress? Is it possible that the command halts mid operation? If yes, what do I do recover?

Deflaimun: Also I believe we should add more emphasis to this part as it's super important. Suggest bold, but INFO admonition could also work

modules/manage/partials/mountable-topics.adoc
Deflaimun: How does the user get the migration id? Is it a return from the rpk/API operation?

Deflaimun: Is there a way to list all migration ids?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+Admin API::
++
+--
+```
+curl -X POST http://localhost:9644/v1/<migration-id>/?action=cancel
+```
+--
+======
+
+You cannot cancel mount and unmount operations in the following <<monitor-progress,states>>:
```
Deflaimun: How do I check those states? 
Monitor should be before cancel, no? We talk about the states before introducing them

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+--
+======
+
+The response returns the IDs and state of existing mount and unmount operations ("migrations"):
+
+|===
+| State | Unmount operation (outbound) | Mount operation (inbound)
+
+| `planned`
+2+| Redpanda validates the operation definition.
```
Deflaimun: On the preview, this presentation looks weird. Looks like it only applies to the first header. See screenshot. As a workaround, consider making this text longer.
We should look into fixing it from the UI side, by centering the row in these cases
image

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+2+| The operation is canceled.
+
+|===
+
+It is not currently possible to unmount a topic whose name matches multiple topics in the origin cluster.
+
+Redpanda prevents you from mounting a topic to multiple clusters at once. This ensures that multiple clusters don't write to the same location in object storage and corrupt the topic.
+
+If you attempt to mount a topic where the name matches a topic already in the target cluster, Redpanda fails the operation and emits a warning message in the logs.
+
```
Deflaimun, line 262: remove extra unnecessary lines

modules/manage/partials/mountable-topics.adoc
```diff
@@ -1,8 +1,8 @@
-For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This allows you to free up and reclaim unused partition space, or migrate a topic to a different cluster and hibernate or decommission the topic or even the entire cluster.
+For topics with Tiered Storage enabled, you can unmount a topic to detach segment data that is still on disk to object storage, and unmount a topic from object storage to attach the topic data to either the same origin cluster, or a different one. This allows you to hibernate a topic and free up and reclaim system resources taken up by the topic, or migrate a topic to a different cluster.
```
Deflaimun: This diff made the page more tech accurate but also harder to understand. Consider simplifying. Also see https://github.com/redpanda-data/docs/pull/725#issuecomment-2417549583 . Is unmount similar to just uploading the topic+metadata to Object Storage?

Feediver1: Seems odd that it is called Mountable, but most of the content here is about unmounting.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
```
Feediver1:
```suggestion
For topics with Tiered Storage enabled, you can unmount a topic to detach segment data that is still on disk to object storage, and unmount a topic from object storage to attach the topic data to either the same origin cluster, or a different one. This allows you to hibernate a topic and free up or reclaim system resources taken up by the topic, or migrate a topic to a different cluster.
```

modules/manage/partials/mountable-topics.adoc
Deflaimun: free up/reclaim system resources are synonymous. Keep one or the other, but not both

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+      "topic": "<topic-2-name>"
+    }, 
+    {
+      "topic": "<topic-3-name>"
+    }
+  ]
+}
+```
+
+You may optionally include the topic namespace (`ns`). Only `kafka` is currently supported.
```
Feediver1:
```suggestion
You may optionally include the topic namespace (`ns`). Only `kafka` is supported.
```

modules/manage/partials/mountable-topics.adoc
kbatuigas: @Feediver1 That's my mistake, I'll change it.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -1,8 +1,8 @@
-For topics with Tiered Storage enabled, you can mount and unmount topics to transfer the topic data between your cluster and object storage. This allows you to free up and reclaim unused partition space, or migrate a topic to a different cluster and hibernate or decommission the topic or even the entire cluster.
+For topics with Tiered Storage enabled, you can unmount a topic to detach segment data that is still on disk to object storage, and unmount a topic from object storage to attach the topic data to either the same origin cluster, or a different one. This allows you to hibernate a topic and free up and reclaim system resources taken up by the topic, or migrate a topic to a different cluster.
 
-Redpanda also transfers topic definitions when mounting or unmounting a topic.
+Redpanda also transfers topic manifests when mounting or unmounting a topic, so the topic can quickly accept reads and writes again and you can resume cluster workloads with ease.
```
Deflaimun: Wait. If the topic is deleted from the source cluster after unmounting, how does it accept read and writes again? 
Maybe this needs to be expanded.  

```suggestion
Redpanda also transfers topic manifests when mounting or unmounting a topic, making possible to quickly resume operations after re-mounting. 
```

(or something to that effect)

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+    }, 
+    {
+      "source_topic": {"ns": "kafka", "topic": "source-topic-3-name"}, 
+      "alias": {"ns": "kafka", "topic": "<new-topic-3-name>"}
+    }
+  ]
+}
+```
+
+* `ns` is the topic namespace. This field is optional and only `kafka` is currently supported.
```
Feediver1:
```suggestion
* `ns` is the topic namespace. This field is optional and only `kafka` is supported.
```

modules/manage/partials/mountable-topics.adoc
Feediver1: See line 142 below--isn't that what you are asking about here @Deflaimun ?

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+| `executed` 
+| All unmounted topic data from the cluster is available in object storage.
+| The target cluster has verified that the topic has not already been mounted.
+
+| `cut_over`
+| Redpanda deletes topic metadata from the cluster, and marks the data in object storage as available for mount operations.
+| The topic data in object storage is no longer available to mount to any clusters.
+
+| `finished`
+| The operation is complete and then deleted.
```
Feediver1:
```suggestion
| The operation is complete and deleted. 
```

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,211 @@
...
+| `finished`
+| The operation is complete and then deleted. 
+| The operation is complete and then deleted. The target cluster starts to handle produce and consume workloads.
+
+| `canceling`
+2+| Redpanda is in the process of canceling the operation.
+
+
+| `cancelled`
+2+| The operation is canceled.
```
Feediver1: you have two different spellings of canceled here.

modules/manage/partials/mountable-topics.adoc
kbatuigas: @Deflaimun (@mattschumpert to correct me if this is wrong) It's detaching the topic. So most of topic data would have already been uploaded to Tiered Storage (based on how quickly users have configured segment data to be moved to the cloud). Unmounting takes what is still on segment in the disk, moves it to the cloud (and stops reads and writes) so that it's all ready to "attach" again to a cluster.

modules/manage/partials/mountable-topics.adoc
Feediver1:
```suggestion
It is not possible to unmount a topic whose name matches multiple topics in the origin cluster.
```

modules/manage/partials/mountable-topics.adoc
kbatuigas: @Deflaimun I think in a previous commit the monitor section was actually before cancel. I'll change it back.

modules/manage/partials/mountable-topics.adoc
Deflaimun: as "cancelled" is already a state in the code, suggest that you do "cancelled" in the description too.

modules/manage/partials/mountable-topics.adoc
Deflaimun: what is deleted? the operation itself or the topic? 
if topic, consider this suggestion.

```suggestion
| The operation is complete and topic is deleted from the source cluster. 
```

modules/manage/partials/mountable-topics.adoc
Deflaimun: do we consider data singular or plural?

modules/manage/partials/mountable-topics.adoc
kbatuigas: @Deflaimun The underlying migration is deleted. I think the topic itself would have already been deleted in the prior state. @bashtanov can you confirm?

Deflaimun: The operation being deleted seems weird to me. If deleted then I can't query about it anymore, thus the state will never be accessible. What if I want to audit the operation? Can the operation be offloaded to a log or something? @bashtanov

modules/manage/partials/mountable-topics.adoc
```diff
@@ -11,7 +9,11 @@ An unmounted topic in object storage is detached from all clusters. The original
 
 == Unmount a topic from a cluster to object storage
 
-When you unmount a topic, all incoming writes to the topic are blocked as Redpanda unmounts the topic from the cluster to object storage. Producers and consumers of the topic receive an error message indicating that the topic is no longer available. The unmounted topic is deleted in the source cluster.
+When you unmount a topic, all incoming writes to the topic are blocked as Redpanda unmounts the topic from the cluster to object storage. Producers and consumers of the topic receive an error message indicating that the topic is no longer available.
```
Deflaimun: What's the expected error message? How do I differentiate the error between an unmounted topic and one that never existed?

kbatuigas: @bashtanov just to check if you have those expected messages available.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,213 @@
+For topics with Tiered Storage enabled, you can unmount a topic to detach segment data that is still on disk to object storage, and mount that topic to either the same origin cluster, or a different one. This allows you to hibernate a topic and free up system resources taken up by the topic, or migrate a topic to a different cluster.
```
mattschumpert: 'detach segment data' is implementation detail.  

Something like:

'you can unmount a topic to safely detach it from a cluster while keeping the topic's data in the cluster's cloud storage bucket/container'

modules/manage/partials/mountable-topics.adoc
mattschumpert: Will there be a page in cloud docs as well?  There it will be RPK or the Cloud API

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+Redpanda also transfers topic manifests when mounting or unmounting a topic, making it possible to quickly resume operations after mounting to the destination cluster.
```
mattschumpert: I don't know that this is a useful statement to a user @kbatuigas .  Nobody knows what topic manifests are and how they affect the user experience. No one ever interacts with a manifest directly.   To them they are just mounting and unmounting a topic.  Am I missing something @nvartolomei ?

Also, transferring is misleading. for TS we already have manifests in the bucket.  This is a 'detach/reattach' operation.

I think we can just remove this statement.

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+- `planned` (but you may still xref:api:ROOT:admin-api.adoc#delete-/v1/migrations/-id-[delete] a planned mount or unmount)
+- `cut_over`
+- `finished`
+- `canceling`
+- `cancelled`
+
+== Additional considerations
+
+It is not possible to unmount a topic whose name matches multiple topics in the origin cluster.
```
mattschumpert: I don't understand the point of this. There is no possibility for this ever to happen whatsoever. By definition the origin cluster cannot have duplicate topic names in the first place so there is no potential for this to be a problem. cc @nvartolomei

mattschumpert (changes requested):
Still some unaddressed comments

modules/manage/partials/mountable-topics.adoc
kbatuigas: @mattschumpert Yes, we'll have a page in the cloud docs. We'll share a lot of the content between the two docs but in cloud it'll show cloud API commands.

modules/manage/partials/mountable-topics.adoc
kbatuigas: @mattschumpert That's my mistake, Matt. I meant to describe a scenario that I believe will be handled by [this change](https://github.com/redpanda-data/redpanda/pull/23826) so I'll go ahead and remove this line.

modules/manage/partials/mountable-topics.adoc
bashtanov: yes, it is `failed to download manifest for topic` and it is logged with warning severity

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,230 @@
...
+  "topics": [
+    {
+      "source_topic": {"ns": "kafka", "topic": "<source-topic-1-name>"}, 
+      "alias": {"ns": "kafka", "topic": "<new-topic-1-name>"}
+    }, 
+    {
+      "source_topic": {"ns": "kafka", "topic": "<source-topic-2-name>"}
+    }, 
+    {
+      "source_topic": {"ns": "kafka", "topic": "source-topic-3-name"},
```
bashtanov: angle brackets missing?

modules/manage/partials/mountable-topics.adoc
bashtanov: @kbatuigas yes, rather between preparing and here. Originally, when designed for quite unmount+mount, this was in order to minimize topic downtime.

bashtanov (commented):
only minor suggestions

modules/manage/partials/mountable-topics.adoc
```diff
@@ -7,7 +7,7 @@ For topics with Tiered Storage enabled, you can unmount a topic to safely detach
 
 == Unmount a topic from a cluster to object storage
 
-When you unmount a topic, all incoming writes to the topic are blocked as Redpanda unmounts the topic from the cluster to object storage. Producers and consumers of the topic receive an error message indicating that the topic is no longer available. 
+When you unmount a topic, all incoming writes to the topic are blocked as Redpanda unmounts the topic from the cluster to object storage. Producers and consumers of the topic receive a warning `Failed to download manifest for topic` indicating that the topic is no longer available.
```
bashtanov: Could you change to lower case `f` in `failed` please? I would imagine them using `grep`, which is case-sensitive by default, to search for the line.

bashtanov: And change "indicating that the topic is no longer available" to "indicating that either the topic is unavailable or there are multiple topics under the specified name" or something like this. It will be clear from the rest of the message which one is the case.

kbatuigas: @bashtanov yes, thank you!

kbatuigas: @bashtanov actually, would "multiple topics under the specified name" ever apply in the case of unmount?

bashtanov: I messed up everything. This error -- `failed to download manifest for topic` -- is for mounting when the topic is not available or cannot be unambiguously defined. It will be in logs.

As for producing into a topic that is about to be unmounted, it is `invalid_topic_exception` or `resource_is_being_migrated` they will be getting. When fetching from a not-yet-ready topic it'll be `invalid_topic_exception` as well. These will be in the protocol replies.

david-yu:
Just a driveby comment since I was looking at this PR, should we also provide a disclaimer that this requires an enterprise license?

kbatuigas:
> Just a driveby comment since I was looking at this PR, should we also provide a disclaimer that this requires an enterprise license?

Thanks @david-yu -- added the note. This doc also links out to the Tiered Storage doc which does already have the license disclaimer.

gene-redpanda (approved):
LGTM

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,279 @@
...
+rpk::
++
+--
+. In your target cluster, run this command to list the topics that are available to mount from object storage:
++
+```
+rpk cluster storage list-mountable
+```
++
+The `LOCATION` value returned in the command output is in the format `<topic-name>/<cluster-uuid>/<initial-revision>`. For example:
```
mattschumpert: We should explain more here....

"This value helps to uniquely identify a topic in cloud storage even when multiple topics may have used the same name before being unmounted from different origin clusters."

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,279 @@
...
++
+```
+TOPIC      NAMESPACE  LOCATION
+testtopic  kafka      testtopic/67f5505a-32f3-4677-bcad-3c75a1a702a6/10
+```
++
+Redpanda assigns the number `initial-revision` to a topic upon creation. You can use the topic location instead of just the topic name to uniquely identify a topic to mount in the next step.
```
mattschumpert: to be even clearer, since it's referred to as 'reference' in the next step:

'you can use the topic location as a 'topic reference' instead of....'

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,279 @@
...
++
+Redpanda assigns the number `initial-revision` to a topic upon creation. You can use the topic location instead of just the topic name to uniquely identify a topic to mount in the next step. 
+
+. Mount a topic from object storage:
++
+```
+rpk cluster storage mount <source-topic-reference>
+```
++
+Replace `<source-topic-reference>` with the name of the topic to mount, or the topic location to uniquely identify a topic object storage.
```
mattschumpert: @kbatuigas  I don't think we should be using the word 'source' here.  The topic reference is the reference in the bucket regardless of what the source cluster is

I think 'topic reference' is sufficient.  It's not longer in the 'source' at this point

Also, we should explain again for clarity.  "The topic reference is required when more than 1 topic in cloud storage use the same name"

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,279 @@
...
++
+You can also rename the topic as you mount it to the target cluster:
++
+```
+rpk cluster storage mount <source-topic-reference> --to <new-topic-name>
```
mattschumpert: also just use 'topic-reference'

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,279 @@
...
+
+. Mount a topic from object storage:
++
+```
+rpk cluster storage mount <source-topic-reference>
+```
++
+Replace `<source-topic-reference>` with the name of the topic to mount, or the topic location to uniquely identify a topic object storage.
++
+You can also rename the topic as you mount it to the target cluster:
```
mattschumpert: maybe add.  "The new name acts as an alias while the topic is mounted in the target cluster and will not persist if the topic is later unmounted. This alias does not change the name of the topic in the cloud storage bucket/container"

@nvartolomei please confirm my understanding is correct

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,279 @@
...
+    {
+      "source_topic_reference": {"ns": "kafka", "topic": "<source-topic-3-name>/<cluster-B-uuid>/<initial-revision>"}, 
+      "alias": {"ns": "kafka", "topic": "<new-topic-3-name>"}
+    }
+  ]
+}
+```
+
+* `ns` is the topic namespace. This field is optional and only `kafka` is supported.
+* You may have multiple topics with the same name that are available to mount from object storage. This can happen if you have unmounted topics with this name from different clusters. To uniquely identify a source topic, append the topic name with `/<cluster-uuid>/<initial-revision>.`
```
mattschumpert: maybe at the end add....  "This fully qualified name forms the 'topic reference'"

modules/manage/partials/mountable-topics.adoc
```diff
@@ -0,0 +1,279 @@
...
+| All unmounted topic data from the cluster is available in object storage.
+| The target cluster has verified that the topic has not already been mounted.
+
+| `cut_over`
+| Redpanda deletes topic metadata from the cluster, and marks the data in object storage as available for mount operations.
+| The topic data in object storage is no longer available to mount to any clusters.
+
+| `finished`
+| The operation is complete. 
+| The operation is complete. The target cluster starts to handle produce and consume workloads.
```
mattschumpert: A nit, but 'requests' is more correct than 'workloads'

mattschumpert (approved):

modules/manage/partials/mountable-topics.adoc
kbatuigas: @mattschumpert @nvartolomei does this mean, the alias does not change the name of the _remaining mountable topics_ of the same name in cloud storage?

nvartolomei: I will defer to @bashtanov but as far as I understand alias is a better term here. We don't rename any topics we just present it with a different name.

bashtanov: It definitely does not change the name of the mountable topic. It does "change" the name of the [cluster-mounted] topic in a sense that it may have been different before, and was different originally. I'd write something like "specify a different topic name" instead of "change", as it won't even always change within this operation.

Example:
1. Create topic `foo`.
2. Unmount topic `foo`, so that it becomes a mountable called `foo`.
3. Mount the mountable `foo` as topic `bar`.
4. Unmount topic `bar`, it becomes a mountable called `foo` again, as underlying storage name always remains the same.
5. Remount the mountable `foo` as topic `bar`.

Using an alias in (5) does not make (4-5) change the topic name, but rather makes it keep it.

kbatuigas: Rephrased, checked via Slack

PR #780, opened 2024-09-26
Update get-cloud-api-spec.js

JakeSCahill:
## Description

Makes the script more generic/extensible for future use.

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


scripts/fetch-from-github/fetch.js
```diff
@@ -0,0 +1,95 @@
...
+  `);
+}
+
+if (args.includes('--help') || args.includes('-h')) {
+  printHelp();
+  process.exit(0);
+}
+
+if (args.length < 4) {
+  console.error('Error: Missing arguments.');
```
Deflaimun, line 48: Error could possibly include more information to help the user.

scripts/fetch-from-github/fetch.js
```diff
@@ -0,0 +1,95 @@
+const fs = require('fs');
+const path = require('path');
+
+// Function to load Octokit once
+let octokitInstance = null;
+
+async function loadOctokit() {
+  if (!octokitInstance) {
+    const { Octokit } = await import('@octokit/rest');
+    octokitInstance = process.env.VBOT_GITHUB_API_TOKEN
```
Deflaimun, line 10: Should this be REDPANDA_GITHUB_TOKEN to match other usages in docs-extensions-and-macros?

scripts/fetch-from-github/fetch.js
Deflaimun: not sure if mixing two different log levels in the same output is a good idea. Maybe wrap this in another function and call something like 

```
function logMessage(message, level = 'log') {
  if (level === 'error') {
    console.error(message);
  } else {
    console.log(message);
  }
}

// Print help message function
function printHelp() {
  logMessage(`...`) 
  
  }
...

if (args.length < 4) {
  logMessage('Error: Missing arguments.', 'error'); 
  printHelp();
  process.exit(1);
}
```

Deflaimun (approved):
Check minor comment

scripts/fetch-from-github/fetch.js
JakeSCahill: thanks 👍

scripts/fetch-from-github/fetch.js
JakeSCahill: this doesn't get executed in the same environment so the name isn't so important. We're just getting it from AWS, which also has a different name 😄

PR #834, opened 2024-11-01
Use csi-driver-lvm version 0.6.0

JakeSCahill:
## Description

Fixes https://redpandadata.atlassian.net/browse/DOC-695

## Page previews

[Preview](https://deploy-preview-834--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/self-hosted/kubernetes/eks-guide/#create-sc)

## Checks

- [ ] New feature
- [ ] Content gap
- [x] Support Follow-up

### DIFF
--- a/modules/deploy/partials/kubernetes/guides/create-storageclass.adoc
+++ b/modules/deploy/partials/kubernetes/guides/create-storageclass.adoc
@@ -34,7 +34,13 @@ endif::[]
 +
 The `lvm.devicePattern` property specifies the pattern that the CSI driver uses to identify available NVMe volumes on your worker nodes.
 +
-NOTE: Version 0.6.0 of `csi-driver-lvm` is used to avoid volume-mounting issues with newer versions. Later versions enable some XFS filesystem features that are incompatible with the default kernels used in {k8s-managed-service}, leading to failed mounts.
+[NOTE]
+====
+Version 0.6.0 of `csi-driver-lvm` is used to avoid volume-mounting issues with newer versions. Later versions enable some XFS filesystem features that are incompatible with the default kernels used in {k8s-managed-service}, leading to failed mounts. In recent `mkfs.xfs` updates, the `-i nrext64=1` option is enabled by default, which leads to the following error on older kernels:
+----
+XFS (dm-0): Superblock has unknown incompatible features (0x20) enabled.
+----
+====
 
 . Create the StorageClass:
 +
--- a/modules/deploy/partials/kubernetes/guides/create-storageclass.adoc
+++ b/modules/deploy/partials/kubernetes/guides/create-storageclass.adoc
@@ -13,6 +13,7 @@ ifdef::env-eks[]
 helm repo add metal-stack https://helm.metal-stack.io
 helm repo update
 helm install csi-driver-lvm metal-stack/csi-driver-lvm \
+  --version 0.6.0 \
   --namespace csi-driver-lvm \
   --create-namespace \
   --set lvm.devicePattern='/dev/nvme[1-9]n[0-9]'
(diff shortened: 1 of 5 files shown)


ligfx:
Should we note the actual error somewhere, so we know when to remove the note on "why 0.6.0"? Something like

> Newer `mkfs.xfs` versions enable `-i nrext64=1` by default, which then causes a `Superblock has unknown incompatible features (0x20) enabled` error on older kernels.

modules/deploy/partials/kubernetes/guides/create-storageclass.adoc
```diff
@@ -24,13 +25,16 @@ ifndef::env-eks[]
...
+  --version 0.6.0 \
   --namespace csi-driver-lvm \
   --create-namespace \
   --set lvm.devicePattern='/dev/nvme[0-9]n[0-9]'
 ----
 endif::[]
 +
 The `lvm.devicePattern` property specifies the pattern that the CSI driver uses to identify available NVMe volumes on your worker nodes.
++
+NOTE: Version 0.6.0 of `csi-driver-lvm` is used to avoid volume-mounting issues with newer versions. Later versions enable some XFS filesystem features that are incompatible with the default kernels used in {k8s-managed-service}, leading to failed mounts.
```
JakeSCahill:
```suggestion
[NOTE]
====
Version 0.6.0 of `csi-driver-lvm` is used to avoid volume-mounting issues with newer versions. Later versions enable some XFS filesystem features that are incompatible with the default kernels used in {k8s-managed-service}, leading to failed mounts. In recent `mkfs.xfs` updates, the `-i nrext64=1` option is enabled by default, which leads to the following error on older kernels:
----
XFS (dm-0): Superblock has unknown incompatible features (0x20) enabled.
----
====
```

modules/deploy/partials/kubernetes/guides/create-storageclass.adoc
```diff
@@ -24,13 +25,22 @@ ifndef::env-eks[]
...
   --create-namespace \
   --set lvm.devicePattern='/dev/nvme[0-9]n[0-9]'
 ----
 endif::[]
 +
 The `lvm.devicePattern` property specifies the pattern that the CSI driver uses to identify available NVMe volumes on your worker nodes.
++
+[NOTE]
+====
+Version 0.6.0 of `csi-driver-lvm` is used to avoid volume-mounting issues with newer versions. Later versions enable some XFS filesystem features that are incompatible with the default kernels used in {k8s-managed-service}, leading to failed mounts. In recent `mkfs.xfs` updates, the `-i nrext64=1` option is enabled by default, which leads to the following error on older kernels:
```
JakeSCahill:
```suggestion
Version 0.6.0 is required to avoid volume-mounting issues caused by recent `mkfs.xfs` updates. Newer versions enable the `-i nrext64=1` option, triggering the following error on default {k8s-managed-service} kernels:
```

david-yu:
Lets get this fixed first before deciding further on which CSI driver makes sense for us in the future.

david-yu (approved):

ligfx (approved):

chrisseto (approved):

modules/deploy/partials/kubernetes/guides/create-storageclass.adoc
```diff
@@ -24,13 +25,22 @@ ifndef::env-eks[]
 helm repo add metal-stack https://helm.metal-stack.io
 helm repo update
 helm install csi-driver-lvm metal-stack/csi-driver-lvm \
+  --version 0.6.0 \
```
kbatuigas, line 28: Right now the highlighted line in this code block is `--create-namespace  \`, just want to double check if we wanted to keep it there.

kbatuigas (changes requested):
@JakeSCahill One item to double check, I'll request changes just in case.

modules/deploy/partials/kubernetes/guides/create-storageclass.adoc
JakeSCahill: Good catch - thanks!

kbatuigas (approved):
Lgtm!

PR #835, opened 2024-11-04
Update with latest rpk commands from v0.0.0-20241104git4a0f859

Deflaimun:
### New commands
rpk-cluster-storage-list-mountable

### Description changes
rpk-cluster-license-info
rpk-cluster-storage-list-mountable
rpk-version

### New flags
rpk-topic-list

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/reference/pages/rpk/rpk-cluster/rpk-cluster-license-info.adoc
+++ b/modules/reference/pages/rpk/rpk-cluster/rpk-cluster-license-info.adoc
@@ -2,13 +2,11 @@
 
 Retrieve license information:
 
-----
-Organization:    Organization the license was generated for.
-Type:            Type of license: free, enterprise, etc.
-Expires:         Expiration date of the license.
-License Status:  Status of the loaded license (valid, expired, not_present).
-Violation:       Whether the cluster is using enterprise features without a valid license.
-----
+- Organization:    Organization the license was generated for.
+- Type:            Type of license (free, enterprise, etc).
+- Expires:         Expiration date of the license.
+- License Status:  Status of the loaded license (valid, expired, not_present).
+- Violation:       Whether the cluster is using enterprise features without a valid license.
 
 == Usage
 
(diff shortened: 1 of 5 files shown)


modules/reference/pages/rpk/rpk-cluster/rpk-cluster-license-info.adoc
```diff
@@ -6,6 +6,8 @@ Retrieve license information:
 Organization:    Organization the license was generated for.
 Type:            Type of license: free, enterprise, etc.
 Expires:         Expiration date of the license.
+License Status:  Status of the loaded license (valid, expired, not_present).
```
kbatuigas: In the preview this shows up as a formatted code block, just want to double check that that should be the case

modules/reference/pages/rpk/rpk-cluster/rpk-cluster-license-info.adoc
```diff
@@ -15,6 +17,14 @@ Expires:         Expiration date of the license.
 rpk cluster license info [flags]
 ----
 
+== Aliases
```
kbatuigas, line 18: Not actually sure what to make of this section, are these supposed to be an alias to the command? E.g. Use `status` instead of `info`?

kbatuigas (approved):
A couple of things to double check, but approving so that this isn't a blocker for release.

modules/reference/pages/rpk/rpk-cluster/rpk-cluster-license-info.adoc
Deflaimun: Yes. We have other instances in docs that we do this 
https://deploy-preview-835--redpanda-docs-preview.netlify.app/current/reference/rpk/rpk-cluster/rpk-cluster-txn-list/

the command support both writings

modules/reference/pages/rpk/rpk-cluster/rpk-cluster-license-info.adoc
Deflaimun: hmm we can remove this format, no strong opinion

PR #862, opened 2024-11-15
DOC-425 How to use the new debug bundle cluster configs

JakeSCahill:
## Description

Review deadline: 20 November

Cluster config reference docs will be updated in a separate PR as those are autogenerated cc @Deflaimun 
As a result, the links to those configs won't work yet.

## Checks

- [x] New feature
- [ ] Content gap
- [ ] Support Follow-up


modules/troubleshoot/partials/debug-bundle.adoc
```diff
@@ -168,550 +170,260 @@ For a description of the files and directories, see <<Contents of the debug bund
...
-// tag::inspect[]
-== Inspect the debug bundle
-
-After downloading the debug bundle files, you can inspect the contents to debug your cluster. This section provides some useful data points to check while troubleshooting.
-
-Most files in the debug bundle are JSON files. To make it easier to read these files, this section uses jq. To install jq, see the https://stedolan.github.io/jq/download/[jq downloads page^].
+== Automatically remove debug bundles
 
-=== View the version of Redpanda on all brokers
+To prevent debug bundles from accumulating in the storage directory, you can configure the xref:reference:properties/cluster-properties.adoc#debug_bundle_auto_removal_seconds[`debug_bundle_auto_removal_seconds`] property. This cluster configuration property automatically deletes debug bundles after the specified number of seconds. By default, this property is not set, meaning debug bundles are retained indefinitely.
```
michael-redpanda: nit: Only one bundle will exist at a time (attempting to run the debug bundle generation stuff will result in any previous runs being removed).

modules/troubleshoot/partials/debug-bundle.adoc
```diff
@@ -168,550 +170,260 @@ For a description of the files and directories, see <<Contents of the debug bund
...
+======
+endif::[]
 
-To check the file permissions, file size, and last modification date of the files:
+== Choose where the debug bundle is saved
 
-```bash
-cat data-dir.txt | jq
-```
+The xref:reference:properties/cluster-properties.adoc#debug_bundle_storage_dir[`debug_bundle_storage_dir`] property allows you to specify a custom directory for storing debug bundles. By default, debug bundles are stored in the Redpanda data directory. Configuring a custom storage directory can help manage storage capacity and isolate debug data from operational data.
```
michael-redpanda, line 296: praise: Great wording!

modules/troubleshoot/partials/debug-bundle.adoc
```diff
@@ -168,550 +170,260 @@ For a description of the files and directories, see <<Contents of the debug bund
...
-====
+Changes to this property take effect immediately and do not require a cluster restart.
 
-=== View cluster metadata
+Before you change this property:
 
-```bash
-cat kafka.json | jq '.[0]'
-```
+* Ensure that your chosen directory has sufficient storage capacity to handle debug bundles.
```
michael-redpanda, line 302: suggestion: Maybe make mention that certain parameters to bundle creation effect the size of the bundle (e.g. more logs being collected or more controller logs will increase the size).  So there's not like a great way to define what is "sufficient".  A couple hundred MB's should be fine in most cases.

modules/troubleshoot/partials/debug-bundle.adoc
```diff
@@ -168,550 +170,260 @@ For a description of the files and directories, see <<Contents of the debug bund
...
+* Verify the directory's permissions to ensure Redpanda can write to it.
```
michael-redpanda: praise: Good call out!  By default it is the `redpanda` user which belongs to the `redpanda` group.

modules/troubleshoot/partials/debug-bundle.adoc
```diff
@@ -168,550 +170,260 @@ For a description of the files and directories, see <<Contents of the debug bund
...
-        "Host": "redpanda-2.redpanda.<namespace>.svc.cluster.local.",
-        "Rack": null
-      }
-    ],
-    "Topics": {}
-  },
-  "Error": null
-}
-```
+Changing this path does not remove existing debug bundles in the original directory. If needed, you must manually remove these bundles.
```
michael-redpanda: nit: Actually I had to double check this, but you can still remove the previous bundle even after changing the directory (internally we store the path to the bundle when it's generated and key it by the run ID).  So you don't need this note =)

modules/troubleshoot/partials/inspect-bundle.adoc
```diff
@@ -0,0 +1,541 @@
+== Inspect the debug bundle
```
michael-redpanda, line 1: praise/question: This is great!  Is this new?  @JFlath may have some good info for this as well if you haven't already talked to him.

michael-redpanda (commented):
Just a couple of suggestions, otherwise it looks awesome!

modules/troubleshoot/partials/debug-bundle.adoc
JakeSCahill:
```suggestion
To avoid manually deleting debug bundles, you can configure the xref:reference:properties/cluster-properties.adoc#debug_bundle_auto_removal_seconds[`debug_bundle_auto_removal_seconds`] property. This cluster configuration property automatically deletes debug bundles after the specified number of seconds. By default, this property is not set, meaning debug bundles are retained indefinitely.
```

modules/troubleshoot/partials/debug-bundle.adoc
```diff
@@ -168,550 +170,260 @@ For a description of the files and directories, see <<Contents of the debug bund
...
 
-```bash
-cat admin/brokers.json | jq '.[] | .version'
-```
+Changes to this property take effect immediately and do not require a cluster restart.
 
-Example output:
+ifndef::env-kubernetes[]
+To set this property, use the xref:reference:rpk/rpk-cluster/rpk-cluster-config.adoc[`rpk cluster config`] command:
++
```
JakeSCahill:
```suggestion

```

modules/troubleshoot/partials/debug-bundle.adoc
```diff
@@ -168,550 +170,260 @@ For a description of the files and directories, see <<Contents of the debug bund
...
+[,bash]
+----
+rpk cluster config set debug_bundle_auto_removal_seconds <seconds>
+----
 
-[,json,role=no-copy,subs=attributes+]
+For example, to retain debug bundles for 1 day:
++
```
JakeSCahill:
```suggestion
```

modules/troubleshoot/partials/inspect-bundle.adoc
```diff
@@ -0,0 +1,541 @@
...
+    "expires": 0,
+    "sha256": ""
+  }
+}
+```
+====
+
+=== View metadata about the Redpanda data directory
+
+To check the size of the directories and look for anomalies:
```
Feediver1, line 136: How about a "for example,..." that points out a couple common anomalies that users should watch out for?

Feediver1 (approved):
lgtm

modules/troubleshoot/partials/inspect-bundle.adoc
JakeSCahill: @JFlath or @michael-redpanda do you have any advice we can provide here about what an anomalous RP directory size might look like?

modules/troubleshoot/partials/debug-bundle.adoc
JakeSCahill: Updated 😄

modules/troubleshoot/partials/debug-bundle.adoc
JakeSCahill: Thanks - added this detail

modules/troubleshoot/partials/debug-bundle.adoc
JakeSCahill: Removed

modules/troubleshoot/partials/inspect-bundle.adoc
JFlath: It's difficult because it's all a bit subjective and dependant on the dataset. Maybe just if one partition of a given topic is notably larger than others, or contains a lot more data that you would expect for that topic.

modules/troubleshoot/partials/inspect-bundle.adoc
JFlath: High level thoughts (happy to add more detail if needed):

* Review Redpanda's logs (in `redpanda.log` for non-K8s, `logs/` for K8s)
* Review topic definitions and configs (in `kafka.json`, but this requires a fair bit of `jq` magic)

michael-redpanda (approved):

PR #870, opened 2024-11-20
Update public-metrics-reference.adoc

Feediver1:
## Description

Resolves https://redpandadata.atlassian.net/browse/DOC-750
Review deadline: **Nov 21**

## Page previews

[Preview](https://deploy-preview-870--redpanda-docs-preview.netlify.app/current/reference/public-metrics-reference/#redpanda_memory_available_memory)

## Checks

- [ ] New feature
- [ ] Content gap
- [x] Support Follow-up

### DIFF
--- a/modules/reference/pages/public-metrics-reference.adoc
+++ b/modules/reference/pages/public-metrics-reference.adoc
@@ -208,7 +208,7 @@ The low watermark for available memory at process start.
 
 === redpanda_memory_free_memory
 
-Total memory potentially available (free plus reclaimable memory) to a CPU shard (core), in bytes.
+Total memory potentially available (includes both free and reclaimable memory) to a CPU shard (core), in bytes.
 
 *Type*: gauge
 
--- a/modules/reference/pages/public-metrics-reference.adoc
+++ b/modules/reference/pages/public-metrics-reference.adoc
@@ -208,7 +208,7 @@ The low watermark for available memory at process start.
 
 === redpanda_memory_free_memory
 
-Total memory potentially available (includes both free and reclaimable memory) to a CPU shard (core), in bytes.
+Total amount of unallocated memory for a CPU shard (core), in bytes.
 
 *Type*: gauge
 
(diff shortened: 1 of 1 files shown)


modules/reference/pages/public-metrics-reference.adoc
```diff
@@ -208,7 +208,7 @@ The low watermark for available memory at process start.
 
 === redpanda_memory_free_memory
 
-Available memory in bytes.
+Total memory potentially available (free plus reclaimable memory) to a CPU shard (core), in bytes.
```
Feediver1:
```suggestion
Total memory potentially available (includes both free and reclaimable memory) to a CPU shard (core), in bytes.
```

modules/reference/pages/public-metrics-reference.adoc
```diff
@@ -208,7 +208,7 @@ The low watermark for available memory at process start.
 
 === redpanda_memory_free_memory
 
-Available memory in bytes.
+Total memory potentially available (includes both free and reclaimable memory) to a CPU shard (core), in bytes.
```
JFlath: I don't think this is correct.

`redpanda_memory_free_memory` = Unallocated memory

`redpanda_memory_available_memory` = Unallocated + reclaimable

The problem with the original wording for this line is that it uses "Available", which is confusing when we use that in the other metric name to indicate something different.

JFlath (changes requested):

modules/reference/pages/public-metrics-reference.adoc
Feediver1:
```suggestion
Total amount of unallocated memory for a CPU shard (core), in bytes.
```

Feediver1: @JFlath Better?

JFlath: I think we should pull the word `available` out. It has a specific meaning to us with memory metrics, so using it generically here is a source of confusion. Maybe just:

> Total amount of unallocated memory for a CPU shard (core), in bytes.

kbatuigas (approved):

JFlath (approved):

PR #879, opened 2024-11-26
DOC-677 Add docs for license status in redpanda resource

JakeSCahill:
## Description

Review deadline: 27 November

## Page previews

https://deploy-preview-879--redpanda-docs-preview.netlify.app/24.3/get-started/licensing/monitor-license-status/#check-the-license-status-on-the-redpanda-resource-in-kubernetes

## Checks

- [x] New feature
- [ ] Content gap
- [ ] Support Follow-up


modules/get-started/pages/licensing/monitor-license-status.adoc
```diff
@@ -147,6 +147,41 @@ Redpanda Console displays warnings in the following scenarios:
...
+----
++
+Replace <resource-name> with the name of your Redpanda resource, such as `example-redpanda`.
+
+
+The `status.licenseStatus` field includes:
+
+- `valid`: Indicates whether the license is valid (true) or not (false).
+- `expiryDate`: The date the license will expire (format: YYYY-MM-DD).
+- `features`: A list of enterprise features currently in use.
```
JakeSCahill:
```suggestion
Replace `<resource-name>` with the name of your Redpanda resource, such as `example-redpanda`.
+
The `status.licenseStatus` field includes:
+
- `valid`: Indicates whether the license is valid (true) or not (false).
- `expiryDate`: The date the license will expire (format: YYYY-MM-DD).
- `features`: A list of enterprise features currently in use.
```

modules/get-started/pages/licensing/monitor-license-status.adoc
```diff
@@ -147,6 +147,40 @@ Redpanda Console displays warnings in the following scenarios:
...
+----
+kubectl get redpanda
+----
++
+Example output:
++
+[.no-copy]
+----
+NAME                  AGE
+example-redpanda      12m
```
andrewstucki: So, this looks like an old version of the CRDs from what I can tell. The latest, if you install from what's released, should look like this, so no real need to do the JSON path thing below:

```bash
➜  crd-demo git:(main) kubectl get redpanda -A
NAMESPACE   NAME      LICENSE       READY   STATUS
redpanda    cluster   Not Present   True    Redpanda reconciliation succeeded
```

Where the column under `LICENSE` should be, when the cluster is stable, one of `Valid`, `Expired`, or `Not Present`, and, in the unstable case either, `redpanda cluster is not yet ready or up to date` or an error explaining why the license status couldn't be fetched.
```

File: modules/get-started/pages/licensing/monitor-license-status.adoc [hunk 3]
```diff
@@ -147,6 +147,40 @@ Redpanda Console displays warnings in the following scenarios:
...
+kubectl get redpanda <resource-name> -o jsonpath='{.status.licenseStatus}'
+----
++
+Replace `<resource-name>` with the name of your Redpanda resource, such as `example-redpanda`.
++
+The `status.licenseStatus` field includes:
++
+- `valid`: Indicates whether the license is valid (true) or not (false).
+- `expiryDate`: The date the license will expire (format: YYYY-MM-DD).
+- `features`: A list of enterprise features currently in use.
```
andrewstucki: Just FYI, other than valid, we don't show this information in Kubernetes, and valid is just a status condition, there's no `licenseStatus` part of the CRD, we only publish status information under the `ClusterLicenseValid` condition in `status.conditions`.

We could add additional details if we want to, but that's not how this currently works. @david-yu thoughts about extending the status to incorporate some of these or is just publishing how to check these via execing and running `rpk cluster license info` enough since we publish the non-granular license status in the CRD?

andrewstucki (changes requested):
Going to go ahead and request changes here since these fields don't exist on the CRD as of today, but also pinging David to see if he wants me to add them.

modules/get-started/pages/licensing/monitor-license-status.adoc
david-yu: @andrewstucki It would be nice to include those fields as well. I don't know if its possible to do another release of the Operator by Tuesday that includes those fields?

david-yu: Also is this status going to be updated a lot or will only happen maybe like once an hour or so?

andrewstucki: It _may_ be possible to cut something by Tuesday, but seeing as it's a long holiday weekend and the changes will need to get reviewed, it might not land.

RE the updates -- this license check would get run every reconciliation, so whenever something on the Redpanda object (including status) change, when any of its owned resources that it's set up to watch change (I know it watches the StatefulSet, and any Deployments from things like Console), or by [default 10 hours](https://github.com/kubernetes-sigs/controller-runtime/blob/8e44a4307f70a4d5f4c5aaf0869e2bec7167f4ab/pkg/cache/cache.go#L46) defined by a cache resynchronization setting in the controller runtime.

david-yu: Ok lets see how it goes through this week. Perhaps Rafal can help review if a PR is available by tomorrow?

JakeSCahill:
Updated the PR to document how it currently works. I can follow up if/when we change the operator to display more data like expiry and enabled features..

david-yu (approved):
This looks pretty thorough based on my review. I have not tested this out myself but it looks good for tomorrow.

andrewstucki (approved):
Approving while following up in another PR to implement the desired extended status fields in the CRD

asimms41 (approved):
LGTM

modules/get-started/pages/licensing/monitor-license-status.adoc
JakeSCahill: Created a follow-up issue for docs: https://redpandadata.atlassian.net/browse/DOC-868

PR #896, opened 2024-12-05
Warning for topic recovery

Feediver1:
## Description

Resolves https://redpandadata.atlassian.net/browse/DOC-79
Review deadline: **Dec 6**

## Page previews
https://deploy-preview-896--redpanda-docs-preview.netlify.app/current/manage/tiered-storage/#data-archiving
https://deploy-preview-896--redpanda-docs-preview.netlify.app/current/manage/topic-recovery/

## Checks

- [ ] New feature
- [ X] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/manage/partials/topic-recovery-warning.adoc
+++ b/modules/manage/partials/topic-recovery-warning.adoc
@@ -1,0 +1,1 @@
+WARNING: While performing topic recovery, avoid adding additional load (such as produces, consumes, lists or additional recovery operations) to the recovering cluster. Doing so could destabilize the recovery process and result in either an unsuccessful or corrupted recovered topic.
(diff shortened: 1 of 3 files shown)


modules/manage/partials/topic-recovery-warning.adoc
```diff
@@ -0,0 +1 @@
+WARNING: While performing topic recovery, avoid adding additional load (such as produces, consumes, lists or additional recovery operations) to the recovering cluster. Doing so could destabilize the recovery process and result in either an unsuccessful or corrupted recovered topic.
```
mattschumpert: @andrwng I see you filed this a while back in 2023. Is this really still the case (we can't handle load on the target cluster during this time)?   I also think we should use the word 'target cluster' @Feediver1

mattschumpert (approved):

modules/manage/partials/topic-recovery-warning.adoc
Feediver1:
```suggestion
WARNING: While performing topic recovery, avoid adding additional load (such as produces, consumes, lists or additional recovery operations) to the target cluster. Doing so could destabilize the recovery process and result in either an unsuccessful or corrupted recovered topic.
```

andrwng: Things definitely are better now than they were then. I think it's still questionable when doing a deep recovery (redpanda.remote.read=false during recovery), so maybe if anything we can still include this warning for that scenario. Or warn against doing a deep recovery altogether.

kbatuigas (approved):
lgtm!

PR #898, opened 2024-12-06
DOC-840 Fix iceberg small details

Deflaimun:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/<add-your-issue-number-here>
Review deadline:

## Page previews

https://deploy-preview-898--redpanda-docs-preview.netlify.app/current/manage/topic-iceberg-integration/

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/manage/pages/topic-iceberg-integration.adoc
+++ b/modules/manage/pages/topic-iceberg-integration.adoc
@@ -58,7 +58,7 @@ When you enable the Iceberg integration for a Redpanda topic, Redpanda brokers s
 
 == Enable Iceberg integration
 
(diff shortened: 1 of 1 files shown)


Deflaimun:
@kbatuigas why do we mention topic creation here? 
Does iceberg work retroactively? Meaning it will apply to existing topics after we apply the topic config.

If yes, we possibly don't have to  mention topic creation

kbatuigas:
> @kbatuigas why do we mention topic creation here? 
> Does iceberg work retroactively? Meaning it will apply to existing topics. 

You can enable Iceberg on existing topics, just that it won't backfill the new table with the older topic data. I believe best practice is to just create a brand new topic at least for the purpose of trying it out in beta I guess. cc @rockwotj @andrwng @mattschumpert in case there is a specific way we want to convey this

Feediver1 (approved):

PR #913, opened 2024-12-10
Broker restart resets counters to 0

kbatuigas:
## Description

- Also includes some minor reworking of the Monitor Redpanda doc overview

Resolves https://github.com/redpanda-data/documentation-private/issues/2452
Review deadline: 12 Dec 2024

## Page previews

[Monitor Redpanda > Monitor for performance and health](https://deploy-preview-913--redpanda-docs-preview.netlify.app/current/manage/monitoring/#monitor-for-performance-and-health)

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/manage/partials/monitor-health.adoc
+++ b/modules/manage/partials/monitor-health.adoc
@@ -2,7 +2,17 @@
 
 This section provides guidelines and example queries using Redpanda's public metrics to optimize your system's performance and monitor its health.
 
-TIP: To help detect and mitigate anomalous system behaviors, capture baseline metrics of your healthy system at different stages (at start-up, under high load, in steady state) so you can set thresholds and alerts according to those baselines.
+To help detect and mitigate anomalous system behaviors, capture baseline metrics of your healthy system at different stages (at start-up, under high load, in steady state) so you can set thresholds and alerts according to those baselines.
+
+[TIP]
+====
+For counter type metrics, a broker restart causes the count in tools such as Prometheus and Grafana to reset to zero. Redpanda recommends wrapping counter metrics in a rate query to account for broker restarts, for example:
+
+[,promql]
+----
+rate(redpanda_kafka_records_produced_total[5m])
+----
+====
 
 === Redpanda architecture
 
--- a/modules/manage/partials/monitor-health.adoc
+++ b/modules/manage/partials/monitor-health.adoc
@@ -6,7 +6,7 @@ To help detect and mitigate anomalous system behaviors, capture baseline metrics
 
 [TIP]
 ====
-For counter type metrics, a broker restart causes the count in tools such as Prometheus and Grafana to reset to zero. Redpanda recommends wrapping counter metrics in a rate query to account for broker restarts, for example:
+For counter type metrics, a broker restart causes the count to reset to zero in tools like Prometheus and Grafana. Redpanda recommends wrapping counter metrics in a rate query to account for broker restarts, for example:
 
 [,promql]
 ----
--- a/modules/manage/partials/monitor-redpanda.adoc
+++ b/modules/manage/partials/monitor-redpanda.adoc
@@ -1,8 +1,8 @@
-Redpanda exports metrics through two Prometheus endpoints on the Admin API port (default: 9644) for you to monitor system health and optimize system performance.
-
-The xref:reference:internal-metrics-reference.adoc[`/metrics`] is a legacy endpoint that includes many internal metrics that are unnecessary for a typical Redpanda user to monitor. The `/metrics` endpoint is also referred to as the 'internal metrics' endpoint, and Redpanda recommends that you use it for development, testing, and analysis. Alternatively, the xref:reference:public-metrics-reference.adoc[`/public_metrics`] endpoint provides a smaller set of important metrics that can be queried and ingested more quickly and inexpensively. 
+Redpanda exports metrics through two endpoints on the Admin API port (default: 9644) for you to monitor system health and optimize system performance.
 
 include::shared:partial$metrics-usage-tip.adoc[]
+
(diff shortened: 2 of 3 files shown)


modules/manage/partials/monitor-health.adoc
```diff
@@ -4,6 +4,16 @@ This section provides guidelines and example queries using Redpanda's public met
 
 TIP: To help detect and mitigate anomalous system behaviors, capture baseline metrics of your healthy system at different stages (at start-up, under high load, in steady state) so you can set thresholds and alerts according to those baselines.
 
+[TIP]
```
kbatuigas, line 7: @wzzzrd86 I'd like to add the new admonition around here, except there is the existing tip ^ about capturing baseline metrics. This is the current preview if we put the two admonitions right next to each other: https://deploy-preview-913--redpanda-docs-preview.netlify.app/current/manage/monitoring/#monitor-for-performance-and-health can we change or downgrade one of them to just be in regular paragraph formatting, or do they both have to be called out?

modules/manage/partials/monitor-health.adoc
```diff
@@ -4,6 +4,16 @@ This section provides guidelines and example queries using Redpanda's public met
...
+====
+For counter type metrics, a broker restart causes the count in tools such as Prometheus and Grafana to reset to zero. Redpanda recommends wrapping counter metrics in a rate query to account for broker restarts, for example:
+
+[,promql]
+----
+rate(redpanda_kafka_records_produced_total)
```
kbatuigas: @wzzzrd86 is this how users would use the rate function with this metric?

modules/manage/partials/monitor-health.adoc
wzzzrd86: @kbatuigas would it be in the same area, just not marked as a Tip?  If so I think that would be fine.

modules/manage/partials/monitor-health.adoc
wzzzrd86: I think rate needs a time frame as well, but I could be wrong, something lieke rate(redpanda_kafka_records_produced_total [5m])

modules/manage/partials/monitor-health.adoc
kbatuigas: Removed the admonition for the existing tip about baseline metrics, it flows better with the first paragraph whereas the info about counter metrics resetting seems a bit more extraneous

modules/manage/partials/monitor-redpanda.adoc
```diff
@@ -1,8 +1,6 @@
-Redpanda exports metrics through Prometheus endpoints for you to monitor system health and optimize system performance.
+Redpanda exports metrics through two Prometheus endpoints on the Admin API port (default: 9644) for you to monitor system health and optimize system performance.
```
kbatuigas: @wzzzrd86 @masapanda is it correct that we call out "Prometheus" here?

kbatuigas:
Note: need to reset to discard package-lock.json changes

modules/manage/partials/monitor-health.adoc
```diff
@@ -2,7 +2,17 @@
 
 This section provides guidelines and example queries using Redpanda's public metrics to optimize your system's performance and monitor its health.
 
-TIP: To help detect and mitigate anomalous system behaviors, capture baseline metrics of your healthy system at different stages (at start-up, under high load, in steady state) so you can set thresholds and alerts according to those baselines.
+To help detect and mitigate anomalous system behaviors, capture baseline metrics of your healthy system at different stages (at start-up, under high load, in steady state) so you can set thresholds and alerts according to those baselines.
+
+[TIP]
+====
+For counter type metrics, a broker restart causes the count in tools such as Prometheus and Grafana to reset to zero. Redpanda recommends wrapping counter metrics in a rate query to account for broker restarts, for example:
```
micheleRP:
```suggestion
For counter type metrics, a broker restart causes the count to reset to zero in tools like Prometheus and Grafana. Redpanda recommends wrapping counter metrics in a rate query to account for broker restarts, for example:
```

modules/manage/partials/monitor-redpanda.adoc
```diff
@@ -1,8 +1,6 @@
...
-
-Before v22.2, a Redpanda broker provided metrics only through the xref:reference:internal-metrics-reference.adoc[`/metrics`] endpoint through the Admin API port. While Redpanda still provides this endpoint, it includes many internal metrics that are unnecessary for a typical Redpanda user to monitor. Consequently, the `/public_metrics` endpoint was added to provide a smaller set of important metrics that can be queried and ingested more quickly and inexpensively. The `/metrics` endpoint is now referred to as the 'internal metrics' endpoint, and Redpanda recommends that you use it for development, testing, and analysis.
+The xref:reference:internal-metrics-reference.adoc[`/metrics`] is a legacy endpoint that includes many internal metrics that are unnecessary for a typical Redpanda user to monitor. The `/metrics` endpoint is also referred to as the 'internal metrics' endpoint, and Redpanda recommends that you use it for development, testing, and analysis. Alternatively, the xref:reference:public-metrics-reference.adoc[`/public_metrics`] endpoint provides a smaller set of important metrics that can be queried and ingested more quickly and inexpensively.
```
micheleRP:
```suggestion
The xref:reference:internal-metrics-reference.adoc[`/metrics`] endpoint is a legacy endpoint that includes many internal metrics that are unnecessary for a typical Redpanda user to monitor. The `/metrics` endpoint is also referred to as the 'internal metrics' endpoint, and Redpanda recommends that you use it for development, testing, and analysis. Alternatively, the xref:reference:public-metrics-reference.adoc[`/public_metrics`] endpoint provides a smaller set of important metrics that can be queried and ingested more quickly and inexpensively. 
```
@kbatuigas I suggest moving the TIP with the 2 endpoints above this explanation.

micheleRP (approved):
Looks good! I left a couple suggestions to consider.

modules/manage/partials/monitor-redpanda.adoc
wzzzrd86: Yes I think that is correct

wzzzrd86 (approved):
LGTM

modules/manage/partials/monitor-redpanda.adoc
Deflaimun: I think you solved this already, but just for the future, I don't think calling it "prometheus endpoints" is accurate. They're HTTP endpoints, through the Admin API, that the response body with a prometheus format.

PR #956, opened 2025-01-17
Point config at the redpanda-operator repo

JakeSCahill:
## Description

Review deadline: 21 Jan

We now host the config for the `crd-ref-docs` tool in the `redpanda-operator` repo so that the K8s team can manage it: https://github.com/redpanda-data/redpanda-operator/pull/396/files#diff-0cb4f1303ff20a2dac37c99d821bff1eded9843e59401d605c4c8314d8762c3e

## Page previews

https://deploy-preview-956--redpanda-docs-preview.netlify.app/current/reference/k-crd/

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/.github/workflows/generate-crd.yml
+++ b/.github/workflows/generate-crd.yml
@@ -50,7 +50,7 @@ jobs:
             --source-path=./redpanda/operator/api/redpanda/v1alpha2 \
             --max-depth=10 \
             --templates-dir=./redpanda-docs/.github/crd-config/templates/asciidoctor/operator \
-            --config=./redpanda-docs/.github/crd-config/config.yaml \
+            --config=./redpanda/operator/crd-ref-docs-config.yaml \
             --renderer=asciidoctor \
             --output-path=./redpanda-docs/modules/reference/pages/k-crd.adoc
       # Check for any changes made in the documentation.


chrisseto (approved):
Preview site looks good to me!

The one maybe oddity is that you have to click twice to get to the actual references:

Screenshot 2025-01-21 at 09 39 31


Still, no `invalid type` and fewer broken links is a win 🎉  (Disclaimer I did not check all links so I'm just guessing that some are still broken)

JakeSCahill:
Thanks for this, Chris! The extra click is a scalability thing. We have a parent index page so that in the future, we can automatically list all available CRDs/versions on that page.

asimms41 (commented):
Just some observations...

- There are some parts of the spec that are not populated. For example RPControllers, PodTemplate, PodSpecApplyConfiguration. Should these be visible?
- Links to docs (and other resources) are shown as html links. For example TopicSpec. I take it that is a limitation?
- Should all config be at the same level? You need to drill down to find certain objects, fields/properties etc but the in-page navigation shows everything at the same level. Also, some fields/properties have the type shown, for example AuthorizationType (string).

JakeSCahill:
Thanks Angela, I'll add these items to a separate Jira on improving the CRD spec docs. 

https://redpandadata.atlassian.net/browse/DOC-969

asimms41 (approved):
LGTM for a v1

PR #935, opened 2024-12-19
DOC-31 add rpk tutorial

Deflaimun:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/1373
Resolves https://redpandadata.atlassian.net/browse/DOC-31
Review deadline: Jan 7th

## Page previews

https://deploy-preview-935--redpanda-docs-preview.netlify.app/current/get-started/rpk-quickstart/

## Checks

- [ ] New feature
- [x] Content gap
- [ ] Support Follow-up
- [ ] Small fix (typos, links, copyedits, etc)

PR #908, opened 2024-12-09
DOC-824 rpk with wsl

Deflaimun:
## Description

Resolves https://github.com/redpanda-data/documentation-private/issues/2802
Review deadline:

## Page previews

This partial affects multiple pages 

https://deploy-preview-908--redpanda-docs-preview.netlify.app/current/deploy/deployment-option/self-hosted/kubernetes/local-guide/

https://deploy-preview-908--redpanda-docs-preview.netlify.app/current/get-started/quick-start/#start-streaming

https://deploy-preview-908--redpanda-docs-preview.netlify.app/current/manage/kubernetes/security/tls/k-cert-manager/#test-external-connections

https://deploy-preview-908--redpanda-docs-preview.netlify.app/current/manage/kubernetes/security/tls/k-secrets/#test-external-connections

## Checks

- [ ] New feature
- [x] Content gap
- [ ] Support Follow-up
- [x] Small fix (typos, links, copyedits, etc)

PR #970, opened 2025-01-29
Fix links

JakeSCahill:
## Description

Fixes broken links that were introduced in:

- https://github.com/redpanda-data/docs/pull/908
- https://github.com/redpanda-data/docs/pull/935

The files that were changed are single-sourced into the Cloud docs and the Connect docs. As some of the links introduced do not exist in those doc-sets, the links broke.

To avoid this in the future, we can:

- Make sure to always check the deploy logs before merging.
- Always run the build locally before pushing to GitHub.
- Add more comments to single-sourced docs to clarify that care must be taken when adding new links.

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


asimms41 (approved):
LGTM

PR #1015, opened 2025-03-14
rpk: update profile docs

r-vasquez:
## Description

There are multiple changes here:

1. Update `default` -> `global`
2. Not documenting the manipulation of the redpanda.yaml (deprecated) or rpk.yaml (we use rpk profile for this instead)
3. Adding information about how to turn a flag into an env variable.

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/get-started/pages/config-rpk-profile.adoc
+++ b/modules/get-started/pages/config-rpk-profile.adoc
@@ -64,16 +64,18 @@ rpk profile use <name-of-profile-to-use>
 
 === Set or edit configuration values
 
+You can customize settings for a single profile. To set a profile's configuration:
+
+* Use xref:reference:rpk/rpk-profile/rpk-profile-set.adoc[`rpk profile set`] to set `key=value` pairs of configuration options to write to the profile's section of `rpk.yaml`.
+* Use xref:reference:rpk/rpk-profile/rpk-profile-edit.adoc[`rpk profile edit`] to edit the profile's section of the `rpk.yaml` file in your default editor.
+
 You can configure settings that apply to all profiles. To set these `globals`:
 
 * Use xref:reference:rpk/rpk-profile/rpk-profile-set-globals.adoc[`rpk profile set-globals`] to set `key=value` pairs to write to the globals section
 of `rpk.yaml`.
 * Use xref:reference:rpk/rpk-profile/rpk-profile-edit-globals.adoc[`rpk profile edit-globals`] to edit the globals section of the `rpk.yaml` file in your default editor.
 
-You can customize settings for a single profile. To set a profile's configuration:
-
-* Use xref:reference:rpk/rpk-profile/rpk-profile-set.adoc[`rpk profile set`] to set `key=value` pairs of configuration options to write to the profile's section of `rpk.yaml`.
-* Use xref:reference:rpk/rpk-profile/rpk-profile-edit.adoc[`rpk profile edit`] to edit the profile's section of the `rpk.yaml` file in your default editor.
+For a list of all the available properties that can be set in your profile, see xref:reference:rpk/rpk-x-options.adoc[`rpk -X options`]
 
 ==== Customize command prompt per profile
 
--- a/modules/get-started/pages/config-rpk-profile.adoc
+++ b/modules/get-started/pages/config-rpk-profile.adoc
@@ -75,7 +75,7 @@ You can configure settings that apply to all profiles. To set these `globals`:
 of `rpk.yaml`.
 * Use xref:reference:rpk/rpk-profile/rpk-profile-edit-globals.adoc[`rpk profile edit-globals`] to edit the globals section of the `rpk.yaml` file in your default editor.
 
-For a list of all the available properties that can be set in your profile, see xref:reference:rpk/rpk-x-options.adoc[`rpk -X options`]
+TIP: For a list of all the available properties that can be set in your profile, see xref:reference:rpk/rpk-x-options.adoc[`rpk -X options`].
 
 ==== Customize command prompt per profile
 
--- a/modules/get-started/pages/intro-to-rpk.adoc
+++ b/modules/get-started/pages/intro-to-rpk.adoc
@@ -24,7 +24,7 @@ endif::[]
 
 You can specify `rpk` command properties in the following ways:
 
-* Create an xref:get-started:config-rpk-profile.adoc[`rpk profile`]
+* Create an xref:get-started:config-rpk-profile.adoc[`rpk profile`].
 * Specify the appropriate flag on the command line.
 * Define the corresponding environment variables.
 +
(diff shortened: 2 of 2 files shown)


modules/get-started/pages/config-rpk-profile.adoc
```diff
@@ -64,16 +64,18 @@ rpk profile use <name-of-profile-to-use>
...
 
 * Use xref:reference:rpk/rpk-profile/rpk-profile-set-globals.adoc[`rpk profile set-globals`] to set `key=value` pairs to write to the globals section
 of `rpk.yaml`.
 * Use xref:reference:rpk/rpk-profile/rpk-profile-edit-globals.adoc[`rpk profile edit-globals`] to edit the globals section of the `rpk.yaml` file in your default editor.
 
-You can customize settings for a single profile. To set a profile's configuration:
-
-* Use xref:reference:rpk/rpk-profile/rpk-profile-set.adoc[`rpk profile set`] to set `key=value` pairs of configuration options to write to the profile's section of `rpk.yaml`.
-* Use xref:reference:rpk/rpk-profile/rpk-profile-edit.adoc[`rpk profile edit`] to edit the profile's section of the `rpk.yaml` file in your default editor.
+For a list of all the available properties that can be set in your profile, see xref:reference:rpk/rpk-x-options.adoc[`rpk -X options`]
```
r-vasquez: I swapped the order here as setting configuration values for single profiles are more important than setting globals.

r-vasquez: Not sure if we should instead have a `partial` that gets rendered both here and on the -X page.

modules/get-started/pages/intro-to-rpk.adoc
```diff
@@ -20,28 +20,21 @@ ifndef::env-cloud[]
 * xref:get-started:rpk-quickstart.adoc[]
 endif::[]
 
-== Specify command properties
+== Specify configuration properties
```
r-vasquez, line 23: A command property only can be specified via flags when you execute a function, for example:

```
rpk topic consume foo --group g1
```

Here, `--group` is a command property and can only be specified via flags.

However, **configuration** properties can be specified in every command; for example, if I want to set which broker rpk talks to, I could do:
```
# set it in the profile
rpk profile set brokers=foo.com
rpk topic consume foo

# Pass it with a -X flag:
rpk topic consume foo -X brokers=foo.com

# Pass it via env flags:
RPK_BROKERS=foo.com rpk topic consume
```

modules/get-started/pages/intro-to-rpk.adoc
```diff
@@ -20,28 +20,21 @@ ifndef::env-cloud[]
...
 
 You can specify `rpk` command properties in the following ways:
 
+* Create an xref:get-started:config-rpk-profile.adoc[]
 * Specify the appropriate flag on the command line.
 * Define the corresponding environment variables.
 +
 Environment variable settings only last for the duration of a shell session.
```
r-vasquez, line 32: Removing references to the `rpk` section as we want to move everything we can to profiles and the `rpk` section of the redpanda.yaml is deprecated.

It will always be maintained, but any new feature will be added to profiles and not there.

w.r.t modifying the `rpk.yaml`, we prefer the user to use `rpk profile` commands to do so and not manually edit the `rpk.yaml`

modules/get-started/pages/intro-to-rpk.adoc
```diff
@@ -64,7 +57,9 @@ Or, get a detailed description about each option with `-X help`:
 rpk -X help
 ----
 
-Some of the common configuration properties apply across all `rpk` commands as defaults. These default properties have keys with names starting with `defaults`, and they're viewable in `rpk -X list` and `rpk -X help`. For more details, see xref:reference:rpk/rpk-x-options.adoc[`rpk -X options`].
+Every `-X` option can be translated into an environment variable by prefixing with `RPK_` and replacing periods (`.`) with underscores (`_`). For example, the flag `tls.enabled` has the equivalent environment variable `RPK_TLS_ENABLED`.
```
r-vasquez: This paragraph is a copy from https://docs.redpanda.com/current/reference/rpk/rpk-x-options/

modules/get-started/pages/intro-to-rpk.adoc
```diff
@@ -20,28 +20,21 @@ ifndef::env-cloud[]
 * xref:get-started:rpk-quickstart.adoc[]
 endif::[]
 
-== Specify command properties
+== Specify configuration properties
 
 You can specify `rpk` command properties in the following ways:
 
+* Create an xref:get-started:config-rpk-profile.adoc[`rpk profile`]
```
micheleRP:
```suggestion
* Create an xref:get-started:config-rpk-profile.adoc[`rpk profile`].
```

modules/get-started/pages/intro-to-rpk.adoc
micheleRP: @r-vasquez Should this line in the next paragraph also be changed then to remove the part about `redpanda.yaml`?
_If the environment variable is not set, the value in the `rpk.yaml` configuration file is used, if that file is available, otherwise the value in the `redpanda.yaml` configuration file is used._

modules/get-started/pages/intro-to-rpk.adoc
micheleRP:
```suggestion
Every `-X` option can be translated into an environment variable by prefixing it with `RPK_` and replacing periods (`.`) with underscores (`_`). For example, the flag `tls.enabled` has the equivalent environment variable `RPK_TLS_ENABLED`.
```

modules/get-started/pages/config-rpk-profile.adoc
micheleRP:
```suggestion
TIP: For a list of all the available properties that can be set in your profile, see xref:reference:rpk/rpk-x-options.adoc[`rpk -X options`].
```
@r-vasquez I don't think so, since there are so many! What do you think about calling it out a little stronger like this, as a tip?

modules/get-started/pages/intro-to-rpk.adoc
micheleRP: @r-vasquez should we also remove the `rpk` bullet from [this page](https://docs.redpanda.com/current/manage/cluster-maintenance/node-property-configuration/)?

r-vasquez: I think it's fine to leave both as:

> If the environment variable is not set, the value in the rpk.yaml configuration file is used, if that file is available, otherwise the value in the redpanda.yaml configuration file is used.

It's useful information on how rpk parses the properties but doesn't encourage users to use the redpanda.yaml

And from the link, it's fine to keep it.

micheleRP (approved):
lgtm, thank you!

PR #1040, opened 2025-03-31
rpk support for protobuf well-known types

kbatuigas:
## Description

Previously, using rpk to produce to or consume a topic using well-known protobuf types would fail, even when using a schema registered in Schema Registry. This is now fixed as of 25.1 and has been backported to 24.3. There are no changes to actual rpk usage.

Review deadline: 1 Apr

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


mattschumpert (approved):

r-vasquez (approved):

micheleRP (approved):
looks good, but since this is a small feature, I suggest moving it down next to the Protobuf normalization item

PR #1052, opened 2025-04-04
Partition memory related changes for 25.1

StephanDollberg:
## Description

Updates docs for the partition memory requirement changes coming in 25.1

Ref ENG-33

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/deploy/partials/requirements.adoc
+++ b/modules/deploy/partials/requirements.adoc
@@ -92,7 +92,9 @@ endif::[]
 
 - 2 GB or more of memory per core.
 
-- 4 MB of memory for each topic partition replica. You can enforce this requirement in the tunable xref:reference:tunable-properties.adoc#topic_memory_per_partition[`topic_memory_per_partition`] property.
+- 2 MB of memory for each topic partition replica.
+
+The total memory available for partition replicas is determined as a percentage of the cluster's total memory, which is controlled by the xref:reference:tunable-properties.adoc#topic_partitions_memory_allocation_percent[`topic_partitions_memory_allocation_percent`] setting. Each partition replica consumes xref:reference:tunable-properties.adoc#topic_memory_per_partition[`topic_memory_per_partition`] bytes from this pool. If insufficient memory is available, topic operations will fail. You can adjust the allocation ratio using topic_partitions_memory_allocation_percent, but doing so is not recommended, as lowering it may lead to instability or degraded performance.
 
 *Recommendations*:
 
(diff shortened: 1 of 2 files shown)


modules/deploy/partials/requirements.adoc
```diff
@@ -92,7 +92,7 @@ endif::[]
 
 - 2 GB or more of memory per core.
 
-- 4 MB of memory for each topic partition replica. You can enforce this requirement in the tunable xref:reference:tunable-properties.adoc#topic_memory_per_partition[`topic_memory_per_partition`] property.
+- Approx. 2 MB of memory for each topic partition replica: Ten percent of the total cluster memory is reserved for partition replicas. Each replica takes about 200KiB of memory. You can change the partition memory reservation using the tunable xref:reference:tunable-properties.adoc#topic_partitions_memory_allocation_percent[`topic_partitions_memory_allocation_percent`] property. This is however not recommended.
```
JakeSCahill:
```suggestion
- 2 MB of memory for each topic partition replica. By default, the system reserves 10% of the total cluster memory for all partition replicas. While each replica typically uses around 200 KiB during normal operation, provisioning for 2 MB per replica ensures sufficient memory under peak load and avoids resource contention.
+
You can adjust this reservation using the xref:reference:properties/cluster-properties.adoc#topic_partitions_memory_allocation_percent[`topic_partitions_memory_allocation_percent`] cluster property. Changing this value is not recommended because under-provisioning may cause performance degradation or instability.
```

JakeSCahill (commented):
@StephanDollberg thanks for the PR 😄 Please review the suggested edit.

modules/deploy/partials/requirements.adoc
StephanDollberg: Thanks for the suggestion. So actually I wasn't really sure how to formulate it but I think this already shows that it's not a good idea to mention the 2MB.

The 2MB is only there to keep the math simple, i.e.: using the whole memory you can think of each partition replica using 2MB as that will result in the same math as doing 10% of total memory divided by 200KiB. But I think this is just confusing.

JakeSCahill: So maybe we can say:

> - 2 MB of memory for each topic partition replica.
The total pool of memory available for partition replicas is determined as a percentage of the cluster's total memory, which is controlled by the `topic_partitions_memory_allocation_percent` setting.
Each partition replica consumes `topic_memory_per_partition` bytes from this pool. If insufficient memory is available, topic operations fail.
You can adjust the allocation ratio using `topic_partitions_memory_allocation_percent`, but doing so is not recommended, as lowering it may lead to instability or degraded performance.

StephanDollberg: Yeah sounds good, have taken that. Note I split the explanation into a separate paragraph how you suggested originally but not strong feelings. Let me know what you think.

JakeSCahill (approved):

PR #1080, opened 2025-04-16
Doc-126 Revise Offset Commit topic

Feediver1:
## Description

Resolves https://redpandadata.atlassian.net/browse/DOC-126
Review deadline: **Friday, April 18**

## Page previews

[Consumer Offsets](https://deploy-preview-1080--redpanda-docs-preview.netlify.app/current/develop/consume-data/consumer-offsets/) (self-managed)
[Consumer Offsets](https://deploy-preview-1080--redpanda-docs-preview.netlify.app/redpanda-cloud/develop/consume-data/consumer-offsets/) (cloud)
This file is single sourced between self-managed and cloud. SMEs: please identify any content that is not true for Cloud. thx

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

## Summary by CodeRabbit

- **Documentation**
  - Expanded and reorganized the guide on consumer offsets, providing a detailed overview, explanations of offset commit strategies, and best practices for managing offsets in Redpanda.
  - Added sections covering offset fundamentals, commit strategies (auto, manual, external, hybrid), and operational recommendations for optimal performance.


modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
 
-When the group coordinator receives an OffsetCommitRequest, it appends the request to a special https://kafka.apache.org/documentation/#compaction[compacted^] Kafka topic named `pass:[__consumer_offsets]`.
+When a glossterm:consumer group[]) works together to consume data from topics, the partitions are divided among the consumers in the group. For example, if a topic has 12 partitions, and there are 2 consumers, each consumer would be assigned 6 partitions to consume. If a new consumer starts later and joins this consumer group, a rebalance occurs, such that each consumer ends up with 4 partitions to consume. You specify a consumer group by setting the `group.id` configuration parameter to a unique name for the group.
```
micheleRP:
```suggestion
When a consumer group works together to consume data from topics, the partitions are divided among the consumers in the group. For example, if a topic has 12 partitions, and there are 2 consumers, each consumer would be assigned 6 partitions to consume. If a new consumer starts later and joins this consumer group, a rebalance occurs, such that each consumer ends up with 4 partitions to consume. You specify a consumer group by setting the `group.id` property to a unique name for the group.
```
Suggest removing this glossterm, since it's defined in the previous paragraph. But if you keep it, there's a typo in it that needs fixing!

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+=== Do not reuse consumer groups
+
+Like many topics, the consumer group topic has multiple partitions to help with performance. When writing commit messages, Redpanda groups all of the commits for a consumer group into a specific partition to maintain ordering. Reusing a consumer group across multiple applications, even for different topics, forces all commits to use a single partition, negating the benefits of partitioning.
+
+*Best practice*: Assign a unique consumer group to each application to distribute the commit load across all partitions.
+
+=== Tune the consumer group
+
+In highly parallel applications, frequent consumer group heartbeats can create unnecessary overhead. For example, 3,200 consumers checking every 500ms generate 6,400 heartbeats per second. You can optimize this behavior by increasing the `heartbeat.interval.ms` (along with `session.timeout.ms`).
```
micheleRP:
```suggestion
In highly parallel applications, frequent consumer group heartbeats can create unnecessary overhead. For example, 3,200 consumers checking every 500 milliseconds generate 6,400 heartbeats per second. You can optimize this behavior by increasing the `heartbeat.interval.ms` (along with `session.timeout.ms`).
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+== Offset commit best practices
+
+Follow these best practices to optimize offset commits.
+
+=== Do not commit too frequently
+
+The purpose of a commit is to save consumer progress. More frequent commits reduce the amount of data to re-read after an application restart, as the commit interval directly affects the Recovery Point Objective (RPO). Because a lower RPO is desirable, application designers may believe that committing frequently is a good design choice.
+
+However, committing too frequently can result in adverse consequences. While individually small, each commit still results in a message being written to the `pass:[__consumer_offsets]` topic because the position of the consumer against every partition must be recorded. At high commit rates, this workload can become a bottleneck for both the client and the server. Additionally, many Kafka client implementations do not coalesce offset commits, meaning redundant commits in a backlog still need to be processed.
```
micheleRP:
```suggestion
However, committing too frequently can result in adverse consequences. While individually small, each commit still results in a message being written to the `pass:[__consumer_offsets]` topic, because the position of the consumer against every partition must be recorded. At high commit rates, this workload can become a bottleneck for both the client and the server. Additionally, many Kafka client implementations do not coalesce offset commits, meaning redundant commits in a backlog still need to be processed.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+In many Kafka client implementations, offset commits aren't coalesced at the client; so if a backlog of commits forms (when using the asynchronous commit API), the earlier commits still need to be processed, even though they are effectively redundant.
+
+*Best practice*: Monitor commit latency to ensure commits are timely and consider reducing commit frequency if performance issues arise.
```
micheleRP:
```suggestion
*Best practice*: Monitor commit latency to ensure commits are timely. If you notice performance issues, commit less frequently.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
 // tag::single-source[]
 :description: pass:q[Redpanda uses an internal topic, `__consumer_offsets`, to store committed offsets from each Kafka consumer that is attached to Redpanda.]
 
-Redpanda supports `pass:[__consumer_offsets]`, which is a private topic on a Redpanda broker. The `pass:[__consumer_offsets]` topic stores committed offsets from each Kafka consumer that is attached to Redpanda.
+== Overview
```
micheleRP: Most of our doc pages don't begin with an **Overview** heading, we just jump right into the introductory text.

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+However, committing too frequently can result in adverse consequences. While individually small, each commit still results in a message being written to the `pass:[__consumer_offsets]` topic because the position of the consumer against every partition must be recorded. At high commit rates, this workload can become a bottleneck for both the client and the server. Additionally, many Kafka client implementations do not coalesce offset commits, meaning redundant commits in a backlog still need to be processed.
+
+In many Kafka client implementations, offset commits aren't coalesced at the client; so if a backlog of commits forms (when using the asynchronous commit API), the earlier commits still need to be processed, even though they are effectively redundant.
+
+*Best practice*: Monitor commit latency to ensure commits are timely and consider reducing commit frequency if performance issues arise.
+
+=== Do not reuse consumer groups
```
micheleRP:
```suggestion
=== Use unique consumer groups
```
Just a suggestion to avoid "Do not"

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+The hybrid offset management strategy allows consumers to handle their own consumer rebalancing while still leveraging Redpanda's offset commit functionality. In this approach:
+
+- Consumers bypass the consumer group API and directly assign partitions instead of subscribing to a topic.
+- Offsets are committed to Redpanda.
+
+== Offset commit best practices
+
+Follow these best practices to optimize offset commits.
+
+=== Do not commit too frequently
```
micheleRP:
```suggestion
=== Avoid over-committing
```
Just suggestion because I think "Do not" sounds a bit harsh or prohibitive

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+- Consumers bypass the consumer group API and directly assign partitions instead of subscribing to a topic.
+- Offsets are not committed to Redpanda, but are instead stored in an external storage system.
+
+To implement an external offset management strategy:
+
+. Configure `enable.auto.commit` as `false`.
+. Assign the partitions to consume using `assign(Collection<TopicPartition>)`.
+. Use the offset provided with each ConsumerRecord to save your position.
+. Upon restart, restore the position of the consumer using `seek(TopicPartition, long)`.
```
micheleRP:
```suggestion
. Set `enable.auto.commit` to `false`.
. Use `assign(Collection<TopicPartition>)` to assign partitions. 
. Use the offset provided with each ConsumerRecord to save your position.
. Upon restart, use `seek(TopicPartition, long)` to restore the position of the consumer.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+        
+        // ... and at the appropriate point, call commit (not after every message)
+        consumer.commitSync();
+    }
+}
+----
+
+==== Asynchronous commit
+
+The advantage of asynchronous commits is lower latency, because the consumer does not pause to wait for the commit response. However, a disadvantage is that there is no automatic retry of the commit (`commitAsync`) if it fails. There is also increased coding complexity (due to the asynchronous callbacks) to consider. The following pseudocode shows an example of an asynchronous commit in which the consumer will not block. Instead, the commit call registers a callback, which is executed once the commit returns:
```
micheleRP:
```suggestion
The advantage of asynchronous commits is lower latency, because the consumer does not pause to wait for the commit response. However, there is no automatic retry of the commit (`commitAsync`) if it fails. There is also increased coding complexity (due to the asynchronous callbacks). The following example shows an asynchronous commit in which the consumer will not block. Instead, the commit call registers a callback, which is executed once the commit returns:
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+=== Manual offset commit
+
+The manual offset commit strategy gives consumers greater control over when commits occur. This approach is typically used when a consumer needs to align commits with an external system, such as database transactions in an RDBMS. The main advantage of manual commits is that they allow you to decide exactly when a record is considered consumed. You can use two API calls for this: `commitSync` and `commitAsync`, which differ in their blocking behavior.
+
+==== Synchronous commit
+
+The advantage of synchronous commits is that consumers can take appropriate action before continuing to consume messages, albeit at the expense of increased latency (while waiting for the commit to return). The commit (`commitSync`) will also retry automatically, until it either succeeds or receives an unrecoverable error. The following pseudocode shows an example of synchronous commit:
```
micheleRP:
```suggestion
The advantage of synchronous commits is that consumers can take appropriate action before continuing to consume messages, albeit at the expense of increased latency (while waiting for the commit to return). The commit (`commitSync`) will also retry automatically, until it either succeeds or receives an unrecoverable error. The following example shows a synchronous commit:
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
 If the offsets fail to replicate within a configurable timeout, the offset commit fails and the consumer may retry the commit after backing off.
-The brokers periodically compact the offsets topic, because it only needs to maintain the most recent offset commit for each partition.
+The brokers periodically compact the `pass:[__consumer_offsets]` topic, because it only needs to maintain the most recent offset commit for each partition.
 The coordinator also caches the offsets in an in-memory table to serve offset fetches quickly.
 
-When stored in a partition, a sequential number is assigned to each event for order. This number is the _offset_. The offset allows consumer event processing to start or continue from where it left off. So, if you have a consumer application outage that prevents your application from receiving events, you can use the consumer offset to select and send only the events that occurred when the application was down. By default, the offset starts at 0, but you can manually specify a specific start value.
+== Commit strategies
+
+There are multiple commit strategies available. The most common follow.
```
micheleRP:
```suggestion
There are several strategies for managing offset commits:
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+=== Auto commit
+
+Auto commit is the default commit strategy, where the client automatically commits offsets at regular intervals. You can enable this by setting `enable.auto.commit` to `true`. The client will then commit offsets every `auto.commit.interval.ms` milliseconds.
+
+The primary advantage of the auto commit approach is its simplicity. Once configured, the consumer requires no additional effort, as commits are managed in the background. However, this "hands-off" approach has a drawback: the consumer is unaware of what was committed or when. As a result, after an application restart, some messages may be reprocessed (since consumption resumes from the last committed offset, which may include already processed messages). The behavior provides an “at least once” delivery guarantee.
```
micheleRP:
```suggestion
The primary advantage of the auto commit approach is its simplicity. After it is configured, the consumer requires no additional effort. Commits are managed in the background. However, the consumer is unaware of what was committed or when. As a result, after an application restart, some messages may be reprocessed (since consumption resumes from the last committed offset, which may include already processed messages). The strategy guarantees at-least-once delivery.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
 The coordinator also caches the offsets in an in-memory table to serve offset fetches quickly.
 
-When stored in a partition, a sequential number is assigned to each event for order. This number is the _offset_. The offset allows consumer event processing to start or continue from where it left off. So, if you have a consumer application outage that prevents your application from receiving events, you can use the consumer offset to select and send only the events that occurred when the application was down. By default, the offset starts at 0, but you can manually specify a specific start value.
+== Commit strategies
+
+There are multiple commit strategies available. The most common follow.
+
+=== Auto commit
+
+Auto commit is the default commit strategy, where the client automatically commits offsets at regular intervals. You can enable this by setting `enable.auto.commit` to `true`. The client will then commit offsets every `auto.commit.interval.ms` milliseconds.
```
micheleRP: This is a little confusing to me. If it's the default strategy, why do you need to enable it?

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
 // tag::single-source[]
 :description: pass:q[Redpanda uses an internal topic, `__consumer_offsets`, to store committed offsets from each Kafka consumer that is attached to Redpanda.]
 
-Redpanda supports `pass:[__consumer_offsets]`, which is a private topic on a Redpanda broker. The `pass:[__consumer_offsets]` topic stores committed offsets from each Kafka consumer that is attached to Redpanda.
+== Overview
 
-Redpanda exposes the `pass:[__consumer_offsets]` key to enable the many tools in the Kafka ecosystem that rely on this value for their operation, providing greater ecosystem interoperability with environments and applications.
+All messages within Redpanda are organized by glossterm:topic[]. Messages within a topic are divided between partitions, where each glossterm:partition[] holds a subset of the messages. For example, if a topic has 5 partitions, it would be likely (though not guaranteed) that each partition holds about 20% of the messages.
```
micheleRP:
```suggestion
In Redpanda, all messages are organized by glossterm:topic[] and distributed across multiple partitions. Each glossterm:partition[] holds a subset of the messages. For example, if a topic has five partitions, it would be likely (though not guaranteed) that each partition holds about 20% of the messages.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
 
-NOTE: A consumer group is a set of consumers that cooperate to consume data from some topics. The partitions of all the topics are divided among the consumers in the group. As new group members arrive and old members leave, the partitions are re-assigned so that each member receives a proportional share of the partitions.
+As a consumer reads messages from Redpanda, it can save its progress by “committing the offset” (known as an glossterm:offset commit[]), an action initiated by the consumer, not by Redpanda. Kafka client libraries provide an API for committing offsets, which communicates with Redpanda using the glossterm:consumer group[] API. Each committed offset is stored as a message in the `pass:[__consumer_offsets]` topic, which is a private Redpanda broker topic that stores committed offsets from each Kafka consumer attached to Redpanda, allowing the consumer to resume processing from the last committed point. Note that Redpanda exposes the `pass:[__consumer_offsets]` key to enable the many tools in the Kafka ecosystem that rely on this value for their operation, providing greater ecosystem interoperability with environments and applications.
```
micheleRP: If you took out the **Overview** heading above, you could break up this text by adding a new heading above "As a consumer reads messages...", like **How consumers use offsets**

micheleRP:
```suggestion
As a consumer reads messages from Redpanda, it can save its progress by “committing the offset” (known as an glossterm:offset commit[]), an action initiated by the consumer, not Redpanda. Kafka client libraries provide an API for committing offsets, which communicates with Redpanda using the glossterm:consumer group[] API. Each committed offset is stored as a message in the `pass:[__consumer_offsets]` topic, which is a private Redpanda topic that stores committed offsets from each Kafka consumer attached to Redpanda, allowing the consumer to resume processing from the last committed point. Redpanda exposes the `pass:[__consumer_offsets]` key to enable the many tools in the Kafka ecosystem that rely on this value for their operation, providing greater ecosystem interoperability with environments and applications.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
 
-When the group coordinator receives an OffsetCommitRequest, it appends the request to a special https://kafka.apache.org/documentation/#compaction[compacted^] Kafka topic named `pass:[__consumer_offsets]`.
+When a glossterm:consumer group[]) works together to consume data from topics, the partitions are divided among the consumers in the group. For example, if a topic has 12 partitions, and there are 2 consumers, each consumer would be assigned 6 partitions to consume. If a new consumer starts later and joins this consumer group, a rebalance occurs, such that each consumer ends up with 4 partitions to consume. You specify a consumer group by setting the `group.id` configuration parameter to a unique name for the group.
+
+Kafka consumer tracks the maximum offset it has consumed in each partition and can commit offsets to ensure it can resume processing from the same point in the event of a restart. Kafka allows offsets for a consumer group to be stored on a designated broker, known as the "group coordinator". All consumers in the group send their offset commits and fetch requests to this group coordinator.
```
micheleRP:
```suggestion
Kafka tracks the maximum offset it has consumed in each partition and can commit offsets to ensure it can resume processing from the same point in the event of a restart. Kafka allows offsets for a consumer group to be stored on a designated broker, known as the group coordinator. All consumers in the group send their offset commits and fetch requests to this group coordinator.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+NOTE: More advanced consumers can read data from Redpanda without using a consumer group, by requesting to read a specific topic + partition + offset range. This pattern is often used by stream processing systems such as Apache Spark and Apache Flink, which have their own mechanisms for assigning work (/partitions) to consumers.
```
micheleRP:
```suggestion
NOTE: More advanced consumers can read data from Redpanda without using a consumer group, by requesting to read a specific topic + partition + offset range. This pattern is often used by stream processing systems such as Apache Spark and Apache Flink, which have their own mechanisms for assigning work (/partitions) to consumers.
```
I don't really understand the "(/partitions)" part

modules/develop/pages/consume-data/consumer-offsets.adoc
Feediver1: Agree, but we do have other topics like this one, which have a longish intro/overview, and I think it would simply help frame the introduction more clearly.

modules/develop/pages/consume-data/consumer-offsets.adoc
Feediver1: You still need to understand what it means, how it compares to other strategies, and what the pros and cons are.

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
```
Feediver1:
```suggestion
NOTE: More advanced consumers can read data from Redpanda without using a consumer group, by requesting to read a specific topic + partition + offset range. This pattern is often used by stream processing systems such as Apache Spark and Apache Flink, which have their own mechanisms for assigning work to consumers.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
 // tag::single-source[]
 :description: pass:q[Redpanda uses an internal topic, `__consumer_offsets`, to store committed offsets from each Kafka consumer that is attached to Redpanda.]
 
-Redpanda supports `pass:[__consumer_offsets]`, which is a private topic on a Redpanda broker. The `pass:[__consumer_offsets]` topic stores committed offsets from each Kafka consumer that is attached to Redpanda.
+== Overview
 
-Redpanda exposes the `pass:[__consumer_offsets]` key to enable the many tools in the Kafka ecosystem that rely on this value for their operation, providing greater ecosystem interoperability with environments and applications.
+In Redpanda, all messages are organized by glossterm:topic[] and distributed across multiple partitions. Each glossterm:partition[] holds a subset of the messages. For example, if a topic has five partitions, it would be likely (though not guaranteed) that each partition holds about 20% of the messages.
```
c4milo: ... and distributed across multiple partitions, according to a [partition strategy](https://www.redpanda.com/guides/kafka-tutorial-kafka-partition-strategy). For example, when using the Round-robin strategy, a producer writing to a topic with five partitions would approximately distribute 20% of messages per partition.

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
 
-When the group coordinator receives an OffsetCommitRequest, it appends the request to a special https://kafka.apache.org/documentation/#compaction[compacted^] Kafka topic named `pass:[__consumer_offsets]`.
+When a consumer group works together to consume data from topics, the partitions are divided among the consumers in the group. For example, if a topic has 12 partitions, and there are 2 consumers, each consumer would be assigned 6 partitions to consume. If a new consumer starts later and joins this consumer group, a rebalance occurs, such that each consumer ends up with 4 partitions to consume. You specify a consumer group by setting the `group.id` property to a unique name for the group.
```
c4milo: nit: we may want to be consistent with how we express numbers, either using words or numbers. Further above, there is an example of a topic with five partitions, and this paragraph's example is using a topic with 12 (a number instead of words) partitions.

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+== Commit strategies
+
+There are several strategies for managing offset commits:
+
+=== Auto commit
+
+Auto commit is the default commit strategy, where the client automatically commits offsets at regular intervals. You can enable this by setting `enable.auto.commit` to `true`. The client will then commit offsets every `auto.commit.interval.ms` milliseconds.
+
+The primary advantage of the auto commit approach is its simplicity. After it is configured, the consumer requires no additional effort. Commits are managed in the background. However, the consumer is unaware of what was committed or when. As a result, after an application restart, some messages may be reprocessed (since consumption resumes from the last committed offset, which may include already processed messages). The strategy guarantees at-least-once delivery.
```
c4milo: Side note. Many people new to Kafka are often bitten by this default behavior; they were supposed to consume and write to another data store, however, if the write to that datastore fails, the consumer might not recover it if it was auto-committed. I do not recommend that people use this default behavior unless they know the tradeoff well. So, it can not only duplicate messages, but even worse, it can also drop messages intended to be in another datastore.

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+However, committing too frequently can result in adverse consequences. While individually small, each commit still results in a message being written to the `pass:[__consumer_offsets]` topic, because the position of the consumer against every partition must be recorded. At high commit rates, this workload can become a bottleneck for both the client and the server. Additionally, many Kafka client implementations do not coalesce offset commits, meaning redundant commits in a backlog still need to be processed.
+
+In many Kafka client implementations, offset commits aren't coalesced at the client; so if a backlog of commits forms (when using the asynchronous commit API), the earlier commits still need to be processed, even though they are effectively redundant.
+
+*Best practice*: Monitor commit latency to ensure commits are timely. If you notice performance issues, commit less frequently.
```
c4milo, line 118: great advise 👍🏻

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+
+There are several strategies for managing offset commits:
+
+=== Auto commit
+
+Auto commit is the default commit strategy, where the client automatically commits offsets at regular intervals. You can enable this by setting `enable.auto.commit` to `true`. The client will then commit offsets every `auto.commit.interval.ms` milliseconds.
+
+The primary advantage of the auto commit approach is its simplicity. After it is configured, the consumer requires no additional effort. Commits are managed in the background. However, the consumer is unaware of what was committed or when. As a result, after an application restart, some messages may be reprocessed (since consumption resumes from the last committed offset, which may include already processed messages). The strategy guarantees at-least-once delivery.
+
+=== Manual offset commit
```
c4milo, line 41: Manual commits of offsets IMO is a best practice, although, others might disagree.

c4milo (approved):
LGTM, this all should apply to Redpanda Cloud as well 👍🏻

modules/develop/pages/consume-data/consumer-offsets.adoc
Feediver1: Good catch. Typically, we write out numbers under 10, and use actual numerals when greater than 10.

Feediver1:
```suggestion
When a consumer group works together to consume data from topics, the partitions are divided among the consumers in the group. For example, if a topic has 12 partitions, and there are two consumers, each consumer would be assigned six partitions to consume. If a new consumer starts later and joins this consumer group, a rebalance occurs, such that each consumer ends up with four partitions to consume. You specify a consumer group by setting the `group.id` property to a unique name for the group.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
Feediver1:
```suggestion
In Redpanda, all messages are organized by glossterm:topic[] and distributed across multiple partitions, based on a https://www.redpanda.com/guides/kafka-tutorial-kafka-partition-strategy[partition strategy^]. For example, when using the Round Robin strategy, a producer writing to a topic with five partitions would approximately distribute 20% of messages per glossterm:partition[].
```

modules/develop/pages/consume-data/consumer-offsets.adoc
Feediver1: Actually @micheleRP, this should be KafkaConsumer (which is a client that consumes records from a Kafka cluster), so changing it to just say Kafka would be misleading.

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,125 @@
...
+== Commit strategies
+
+There are several strategies for managing offset commits:
+
+=== Auto commit
+
+Auto commit is the default commit strategy, where the client automatically commits offsets at regular intervals. You can enable this by setting `enable.auto.commit` to `true`. The client will then commit offsets every `auto.commit.interval.ms` milliseconds.
+
+The primary advantage of the auto commit approach is its simplicity. After it is configured, the consumer requires no additional effort. Commits are managed in the background. However, the consumer is unaware of what was committed or when. As a result, after an application restart, some messages may be reprocessed (since consumption resumes from the last committed offset, which may include already processed messages). The strategy guarantees at-least-once delivery.
```
Feediver1:
```suggestion
The primary advantage of the auto commit approach is its simplicity. After it is configured, the consumer requires no additional effort. Commits are managed in the background. However, the consumer is unaware of what was committed or when. As a result, after an application restart, some messages may be reprocessed (since consumption resumes from the last committed offset, which may include already processed messages). The strategy guarantees at-least-once delivery.

NOTE: If your consume configuration is set up to consume and write to another data store, and the write to that datastore fails, the consumer might not recover when it is auto-committed. It may not only duplicate messages, but could also drop messages intended to be in another datastore. Make sure you understand the trade-off possibilities associated with this default behavior.
```

Feediver1:
Also got the approval from @pmw-rp via Slack.

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,127 @@
...
+- Consumers bypass the consumer group API and directly assign partitions instead of subscribing to a topic.
+- Offsets are committed to Redpanda.
+
+== Offset commit best practices
+
+Follow these best practices to optimize offset commits.
+
+=== Avoid over-committing
+
+The purpose of a commit is to save consumer progress. More frequent commits reduce the amount of data to re-read after an application restart, as the commit interval directly affects the Recovery Point Objective (RPO). Because a lower RPO is desirable, application designers may believe that committing frequently is a good design choice.
```
micheleRP:
```suggestion
The purpose of a commit is to save consumer progress. More frequent commits reduce the amount of data to re-read after an application restart, as the commit interval directly affects the recovery point objective (RPO). Because a lower RPO is desirable, application designers may believe that committing frequently is a good design choice.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,127 @@
...
-The brokers periodically compact the offsets topic, because it only needs to maintain the most recent offset commit for each partition.
+The brokers periodically compact the `pass:[__consumer_offsets]` topic, because it only needs to maintain the most recent offset commit for each partition.
 The coordinator also caches the offsets in an in-memory table to serve offset fetches quickly.
 
-When stored in a partition, a sequential number is assigned to each event for order. This number is the _offset_. The offset allows consumer event processing to start or continue from where it left off. So, if you have a consumer application outage that prevents your application from receiving events, you can use the consumer offset to select and send only the events that occurred when the application was down. By default, the offset starts at 0, but you can manually specify a specific start value.
+== Commit strategies
+
+There are several strategies for managing offset commits:
+
+=== Auto commit
```
micheleRP:
```suggestion
=== Automatic offset commit
```
just to be consistent with the other headings

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,127 @@
 // tag::single-source[]
 :description: pass:q[Redpanda uses an internal topic, `__consumer_offsets`, to store committed offsets from each Kafka consumer that is attached to Redpanda.]
 
-Redpanda supports `pass:[__consumer_offsets]`, which is a private topic on a Redpanda broker. The `pass:[__consumer_offsets]` topic stores committed offsets from each Kafka consumer that is attached to Redpanda.
+== Overview
 
-Redpanda exposes the `pass:[__consumer_offsets]` key to enable the many tools in the Kafka ecosystem that rely on this value for their operation, providing greater ecosystem interoperability with environments and applications.
+In Redpanda, all messages are organized by glossterm:topic[] and distributed across multiple partitions, based on a https://www.redpanda.com/guides/kafka-tutorial-kafka-partition-strategy[partition strategy^]. For example, when using the Round Robin strategy, a producer writing to a topic with five partitions would approximately distribute 20% of messages per glossterm:partition[].
```
micheleRP:
```suggestion
In Redpanda, all messages are organized by glossterm:topic[] and distributed across multiple partitions, based on a https://www.redpanda.com/guides/kafka-tutorial-kafka-partition-strategy[partition strategy^]. For example, when using the round robin strategy, a producer writing to a topic with five partitions would distribute approximately 20% of the messages to each glossterm:partition[].
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,127 @@
...
+
+Kafka tracks the maximum offset it has consumed in each partition and can commit offsets to ensure it can resume processing from the same point in the event of a restart. Kafka allows offsets for a consumer group to be stored on a designated broker, known as the group coordinator. All consumers in the group send their offset commits and fetch requests to this group coordinator.
+
+NOTE: More advanced consumers can read data from Redpanda without using a consumer group, by requesting to read a specific topic + partition + offset range. This pattern is often used by stream processing systems such as Apache Spark and Apache Flink, which have their own mechanisms for assigning work to consumers.
```
micheleRP:
```suggestion
NOTE: More advanced consumers can read data from Redpanda without using a consumer group by requesting to read a specific topic, partition, and offset range. This pattern is often used by stream processing systems such as Apache Spark and Apache Flink, which have their own mechanisms for assigning work to consumers.
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,127 @@
...
 The coordinator also caches the offsets in an in-memory table to serve offset fetches quickly.
 
-When stored in a partition, a sequential number is assigned to each event for order. This number is the _offset_. The offset allows consumer event processing to start or continue from where it left off. So, if you have a consumer application outage that prevents your application from receiving events, you can use the consumer offset to select and send only the events that occurred when the application was down. By default, the offset starts at 0, but you can manually specify a specific start value.
+== Commit strategies
+
+There are several strategies for managing offset commits:
+
+=== Auto commit
+
+Auto commit is the default commit strategy, where the client automatically commits offsets at regular intervals. You can enable this by setting `enable.auto.commit` to `true`. The client will then commit offsets every `auto.commit.interval.ms` milliseconds.
```
micheleRP:
```suggestion
Auto commit is the default commit strategy, where the client automatically commits offsets at regular intervals. This is set with the `enable.auto.commit` property. The client then commits offsets every `auto.commit.interval.ms` milliseconds.
```
Just my preference, because I still thought it sounded like you had to enable it.

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,127 @@
...
+
+The primary advantage of the auto commit approach is its simplicity. After it is configured, the consumer requires no additional effort. Commits are managed in the background. However, the consumer is unaware of what was committed or when. As a result, after an application restart, some messages may be reprocessed (since consumption resumes from the last committed offset, which may include already processed messages). The strategy guarantees at-least-once delivery.
```
micheleRP:
```suggestion
The primary advantage of the auto commit approach is its simplicity. After it is configured, the consumer requires no additional effort. Commits are managed in the background. However, the consumer is unaware of what was committed or when. As a result, after an application restart, some messages may be reprocessed (since consumption resumes from the last committed offset, which may include already-processed messages). The strategy guarantees at-least-once delivery.
```

micheleRP (approved):
no blockers, but please see my suggestions!

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,127 @@
 // tag::single-source[]
 :description: pass:q[Redpanda uses an internal topic, `__consumer_offsets`, to store committed offsets from each Kafka consumer that is attached to Redpanda.]
 
-Redpanda supports `pass:[__consumer_offsets]`, which is a private topic on a Redpanda broker. The `pass:[__consumer_offsets]` topic stores committed offsets from each Kafka consumer that is attached to Redpanda.
+== Overview
```
Feediver1:
```suggestion
```

modules/develop/pages/consume-data/consumer-offsets.adoc
```diff
@@ -4,20 +4,127 @@
...
 
-NOTE: A consumer group is a set of consumers that cooperate to consume data from some topics. The partitions of all the topics are divided among the consumers in the group. As new group members arrive and old members leave, the partitions are re-assigned so that each member receives a proportional share of the partitions.
+As a consumer reads messages from Redpanda, it can save its progress by “committing the offset” (known as an glossterm:offset commit[]), an action initiated by the consumer, not Redpanda. Kafka client libraries provide an API for committing offsets, which communicates with Redpanda using the glossterm:consumer group[] API. Each committed offset is stored as a message in the `pass:[__consumer_offsets]` topic, which is a private Redpanda topic that stores committed offsets from each Kafka consumer attached to Redpanda, allowing the consumer to resume processing from the last committed point. Redpanda exposes the `pass:[__consumer_offsets]` key to enable the many tools in the Kafka ecosystem that rely on this value for their operation, providing greater ecosystem interoperability with environments and applications.
```
Feediver1, line 15:
```suggestion

== How consumers use offsets

As a consumer reads messages from Redpanda, it can save its progress by “committing the offset” (known as an glossterm:offset commit[]), an action initiated by the consumer, not Redpanda. Kafka client libraries provide an API for committing offsets, which communicates with Redpanda using the glossterm:consumer group[] API. Each committed offset is stored as a message in the `pass:[__consumer_offsets]` topic, which is a private Redpanda topic that stores committed offsets from each Kafka consumer attached to Redpanda, allowing the consumer to resume processing from the last committed point. Redpanda exposes the `pass:[__consumer_offsets]` key to enable the many tools in the Kafka ecosystem that rely on this value for their operation, providing greater ecosystem interoperability with environments and applications.
```

PR #1126, opened 2025-05-14
iotune: Stop recommending to run iotune during setup

StephanDollberg:
We have found it to cause more harm than good. Hence, stop recommending people to run it.

Leave the docs page for `rpk iotune` but add a note that it's not recommended to run.

Ref CORE-10020

## Checks

- [ ] New feature
- [ ] Content gap
- [x] Support Follow-up


modules/deploy/pages/deployment-option/self-hosted/kubernetes/k-production-deployment.adoc
```diff
@@ -98,22 +96,16 @@ spec:
...
-      extraVolumes: |-
-        - name: redpanda-io-config
-          configMap:
-            name: redpanda-io-config
-      extraVolumeMounts: |-
-        - name: redpanda-io-config
-          mountPath: /etc/redpanda-io-config
-      additionalRedpandaCmdFlags:
-        - "--io-properties-file=/etc/redpanda-io-config/io-config.yaml"
+    #statefulset:
```
StephanDollberg, line 99: @chrisseto or @david-yu do you guys have a recommendation of what is the best way to expose these now empty options?

chrisseto: Remove the entire statefulset block (Everything after line 99 in this PR).

modules/deploy/pages/deployment-option/self-hosted/kubernetes/k-production-deployment.adoc
```diff
@@ -98,22 +96,16 @@ spec:
...
+    #  extraVolumes: |-
+    #  extraVolumeMounts: |-
+    #  additionalRedpandaCmdFlags:
```
chrisseto, line 102:
```suggestion
```

chrisseto (approved):

modules/deploy/pages/deployment-option/self-hosted/kubernetes/k-production-deployment.adoc
StephanDollberg: Thank Chris, so that's what I had originally but then noticed that there is the paragraph below in the bullet point list which refers to the `statefulset` so it would be a bit without context otherwise. I'd expect there to be other things one can set here?

chrisseto: Oh, I didn't see that. Thank you for pointing it out. I think it'd be best to remove that line, otherwise we could specify `statefulset: {}` here. @JakeSCahill what do you think?

modules/reference/pages/rpk/rpk-iotune.adoc
```diff
@@ -5,6 +5,11 @@
 
 include::reference:partial$unsupported-os-rpk.adoc[]
 
+[NOTE]
+====
+It's not required to run iotune for optimial performance.
```
Feediver1:
```suggestion
It's not required to run iotune for optimal performance.
```

Feediver1:
Is this ready for final doc review and approval @StephanDollberg?

StephanDollberg:
@Feediver1 No not at this point, it's likely going to change.

StephanDollberg:
We will be going in a different direction.

PR #1135, opened 2025-05-25
Fix indentation for secretRef in authentication docs

avinashupadhya99:
## Description

The indentation for the `secretRef` in few examples in the [Kubernetes authentication docs page](https://docs.redpanda.com/current/manage/kubernetes/security/authentication/k-authentication) is broken. The `secretRef` key should be under `auth.sasl` and not `auth` (See [helm values](https://github.com/redpanda-data/helm-charts/blob/d16c638447ac045eadd07a57bb2de5f20d8d3f93/charts/redpanda/values.yaml#L162)). There are other examples in the page that have the right reference.

This PR fixes the indentation for `secretRef`s which are under `auth` instead of `auth.sasl`

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/manage/partials/authentication.adoc
+++ b/modules/manage/partials/authentication.adoc
@@ -753,7 +753,7 @@ spec:
     auth:
       sasl:
         enabled: true
-      secretRef: redpanda-superusers
+        secretRef: redpanda-superusers
     config:
       cluster:
         sasl_mechanisms:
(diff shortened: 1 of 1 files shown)


JakeSCahill (approved):
Thanks 🙏

PR #1195, opened 2025-07-03
Rename extraClusterConfig to extraClusterConfiguration

JakeSCahill:
## Description

Based on https://github.com/redpanda-data/redpanda-operator/blob/371f59cfe2cb4605c274c28dff56a29446a67f62/charts/redpanda/values.yaml#L927 I believe the config should be `extraClusterConfiguration` rather than the documented `extraClusterConfig`.

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


chrisseto (approved):
I'm so sorry about that! Thank you for catching that.

PR #1219, opened 2025-07-18
Dev mode

paulohtb6:
## Description

Introduces dev-mode. Speeds up Antora processing to a maximum, making local builds take up to 40s (from 3-4min+).

Should be used to focus on syntax issues and quick checks. 

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


local-dev-playbook.yml
```diff
@@ -0,0 +1,37 @@
...
+    # Only watch specific directories you're working on
+    # start_paths: [modules/get-started, modules/develop]  # Uncomment and customize
+ui:
+  bundle:
+    url: https://github.com/redpanda-data/docs-ui/releases/latest/download/ui-bundle.zip
+    snapshot: true
+asciidoc:
+  attributes:
+    # Aggressive error suppression for dev mode
+    attribute-missing: warn
```
JakeSCahill, line 24: I think this is usually `info`. `warn` seems more agressive since that's the default log level for Antora.

local-dev-playbook.yml
```diff
@@ -0,0 +1,37 @@
...
+    sectlinks: false
+    # Skip heavy validations
+    skip-front-matter: true
+  extensions:
+  - '@asciidoctor/tabs'
+  # Minimal extensions only
+antora:
+  extensions:
+  # Only the absolute minimum for dev mode
+  - require: '@redpanda-data/docs-extensions-and-macros/extensions/modify-redirects'
```
JakeSCahill, line 37: How do we decide on the absolute minimum? This is subjective depending on what you're working on I think.

local-dev-playbook.yml
```diff
@@ -0,0 +1,37 @@
...
+  html_extension_style: indexify
+output:
+  clean: false  # Don't clean on every build
+  dir: ./build/dev-site  # Separate dev output
+content:
+  sources:
+  - url: .
+    branches: HEAD
+    # Only watch specific directories you're working on
+    # start_paths: [modules/get-started, modules/develop]  # Uncomment and customize
```
JakeSCahill, line 16: With this setup you'll miss out on link checking across components and versions. I think this mode should only be used to check the rendered output while you iterate.

We should still recommend running the usual (non-dev mode) before pushing to get all the validations.

local-dev-playbook.yml
paulohtb6: Yes. I believe whoever is using should edit according to what's being tested. There's always local-antora to get the correct info from.

local-dev-playbook.yml
paulohtb6: Yes, absolutely. This should only be using during dev, but the writer must check with `npm run build` before proceeding.

PR #1269, opened 2025-08-01
metrics: Recommend better metrics for global produce/consume TP

StephanDollberg:
We added these metrics a long while ago for better monitoring of global produce/consume throughput.

They don't suffer from various issues (prometheus bugs, including non-returned bytes) like the per partition metrics.

See https://github.com/redpanda-data/redpanda/pull/14836 for more background.

Hence, recommend them for total produce/consume TP monitoring.

## Description

Review deadline:

## Checks

- [ ] New feature
- [ ] Content gap
- [x] Support Follow-up

### DIFF
--- a/modules/manage/partials/monitor-health.adoc
+++ b/modules/manage/partials/monitor-health.adoc
@@ -89,7 +89,7 @@ While maximizing the rate of messages moving from producers to brokers then to c
 
 The total throughput of a cluster can be measured by the producer and consumer rates across all topics.
 
-To observe the total producer and consumer rates of a cluster, monitor xref:reference:public-metrics-reference.adoc#redpanda_kafka_request_bytes_total[`redpanda_kafka_request_bytes_total`] with the `produce` and `consume` labels, respectively.
+To observe the total producer and consumer rates of a cluster, monitor xref:reference:public-metrics-reference.adoc#redpanda_rpc_received_bytes[`redpanda_rpc_received_bytes`] for producer traffic and xref:reference:public-metrics-reference.adoc#redpanda_rpc_sent_bytes[`redpanda_rpc_sent_bytes`] for consumer traffic. Filter both metrics using the `redpanda_server` label with the value `kafka`.
 
 ==== Producer throughput
 
@@ -97,7 +97,7 @@ For the produce rate, create a query to get the produce rate across all topics:
 
 [,promql]
 ----
-sum(rate(redpanda_kafka_request_bytes_total{redpanda_request="produce"} [5m] )) by (redpanda_request)
+rate(redpanda_rpc_received_bytes{redpanda_server="kafka"}[$__rate_interval])
 ----
 
 ==== Consumer throughput
@@ -106,7 +106,7 @@ For the consume rate, create a query to get the total consume rate across all to
 
 [,promql]
 ----
-sum(rate(redpanda_kafka_request_bytes_total{redpanda_request="consume"} [5m] )) by (redpanda_request)
+rate(redpanda_rpc_sent_bytes{redpanda_server="kafka"}[$__rate_interval])
 ----
 
 === Latency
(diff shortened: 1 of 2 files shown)


Feediver1 (approved):
lgtm

modules/manage/partials/monitor-health.adoc
```diff
@@ -89,15 +89,15 @@ While maximizing the rate of messages moving from producers to brokers then to c
...
+To observe the total producer and consumer rates of a cluster, monitor xref:reference:public-metrics-reference.adoc#redpanda_rpc_sent_bytes[`redpanda_rpc_sent_bytes`] and xref:reference:public-metrics-reference.adoc#redpanda_rpc_received_bytes[`redpanda_rpc_received_bytes`] with the `kafka` `redpanda_server` label.
 
 ==== Producer throughput
 
 For the produce rate, create a query to get the produce rate across all topics:
 
 [,promql]
 ----
-sum(rate(redpanda_kafka_request_bytes_total{redpanda_request="produce"} [5m] )) by (redpanda_request)
+rate(redpanda_rpc_received_bytes{redpanda_server="kafka"}[$__rate_interval])
```
travisdowns, line 100: Any comment on dropping `sum`? All series will be present in the result, I think it's consistent with line 96 right above?

modules/manage/partials/monitor-health.adoc
```diff
@@ -89,15 +89,15 @@ While maximizing the rate of messages moving from producers to brokers then to c
 
 The total throughput of a cluster can be measured by the producer and consumer rates across all topics.
 
-To observe the total producer and consumer rates of a cluster, monitor xref:reference:public-metrics-reference.adoc#redpanda_kafka_request_bytes_total[`redpanda_kafka_request_bytes_total`] with the `produce` and `consume` labels, respectively.
+To observe the total producer and consumer rates of a cluster, monitor xref:reference:public-metrics-reference.adoc#redpanda_rpc_sent_bytes[`redpanda_rpc_sent_bytes`] and xref:reference:public-metrics-reference.adoc#redpanda_rpc_received_bytes[`redpanda_rpc_received_bytes`] with the `kafka` `redpanda_server` label.
```
travisdowns: Technical question: these are all send/recv bytes on the kafka API, not limited to produce and consume API keys, right?

travisdowns: I think if we say "produce" and "consume" we should put the metrics in the same order, and say "respetively" as before (now they are swapped I think).

travisdowns (commented):
Review complete with comments.

modules/manage/partials/monitor-health.adoc
StephanDollberg: > All series will be present in the result 

Just to clarify, there is no topic here so there will be one series per broker.

> Any comment on dropping sum?

I mean this example query seems fairly arbitrary to me. You'll likely have to modify it anyway to add a cluster label or something. 

We could just do any empty sum by which again would aggregate everything but that seems kinda wrong to me. No strong feelings though.

modules/manage/partials/monitor-health.adoc
StephanDollberg: Yes that's correct. I think that's less of any issue than the issues we have with the metrics we use right now. Unfortunately we don't have the per handler metrics on public metrics right now (only the latency ones).

StephanDollberg: Fixed the order. The respectively referred to the labels so it doesn't apply anymore, no?

travisdowns: I mean it's still two things to monitor for the two API calls: produce and consume use metrics X and Y respectively (instead of labels X and Y respecitvely).

Kinda feels like it is a bit more important than before even since the two things were really obvious from their naming, while now that isn't true (you have to know that produce will primary increment the `received` metric). Makes the sentence more awkward though!

modules/manage/partials/monitor-health.adoc
travisdowns: I like it how you have it, not sure what I was even thinking here (as it's consistent with the other queries I can see in the doc).

travisdowns (approved):

modules/manage/partials/monitor-health.adoc
StephanDollberg: Ok, I have rephrased.

travisdowns: Thanks.

PR #1302, opened 2025-08-11
docs: clarify OIDC principal format in authentication

sago2k8:
## Description
Update the authentication documentation to better explain how OIDC principals should be formatted in superusers configuration and ACL rules. The changes make it clear that the sub claim value is used directly without prefixes, unlike SASL users.

Review deadline:

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


modules/manage/partials/authentication.adoc
```diff
@@ -63,9 +63,11 @@ You assign ACLs and superuser roles to principals, not users. Even if a user exi
 
 For example, with OIDC:
 
-- If your token's `sub` claim is `example@company.com`, then:
-  * In the `superusers` list, specify the principal as `example@company.com`
-  * In ACL rules, specify the principal as `user:OIDC:example@company.com`
+- If your OIDC token's sub (subject) claim contains example@company.com, then:
```
BenPope: I'm not sure "contains" is clearer than "is", to me, contains implies that there could be other characters preceding or following it.

modules/manage/partials/authentication.adoc
```diff
@@ -63,9 +63,11 @@ You assign ACLs and superuser roles to principals, not users. Even if a user exi
...
+* For superusers configuration: Use the exact sub claim value: example@company.com
+* For ACL rules: Use the exact sub claim value: example@company.com
+Note: Unlike SASL users which require the User: prefix, OIDC principals use the value extracted from the token's claims (according to the configured principal mapping) without any prefix
```
JakeSCahill:
```suggestion
If your OIDC token's sub (subject) claim is example@company.com, use that exact sub claim value for superusers configuration and ACL rules.

Unlike SASL users which require the `User:` prefix, OIDC principals use the value extracted from the token's claims (according to the configured principal mapping) without any prefix
```

BenPope: I don't really know where this prefix comes in, or is displayed, or where the user specifies it.

I think it's pretty much in `:8081/security/acls`, and what's returned from `rpk security acl list`?

sago2k8: can you maybe help me here @r-vasquez , both are valid in my experience (User:<principal>, <principal>), I am not sure what's preferable or which one should be in the documentation ?

r-vasquez: My 2c with this doc change is that the document talks about configuring principals, and then we mention that we can use sub claims, I think we should be more explicit, something like:

```
If your OIDC token's sub (subject) claim is example@company.com, use that exact sub claim value as your principal for superusers configuration and ACL rules.
	
Unlike SASL users which require the `User:` principal type prefix, OIDC principals use the value extracted from the token's claims (according to the configured principal mapping) without any prefix
```

> I think it's pretty much in :8081/security/acls, and what's returned from rpk security acl list?

Yes, also when creating, we have an example in this same page below 
```
Grant the new user describe privileges for a topic called myfirsttopic:

rpk security acl create --allow-principal User:myuser \
```
_Which is actually not needed, but it doesn't hurt. 😅_

BenPope: > Grant the new user describe privileges for a topic called myfirsttopic:
> 
> rpk security acl create --allow-principal User:myuser \
> ```
> 
> _Which is actually not needed, but it doesn't hurt. 😅_

It's not needed, and shouldn't be in the docs. 

Oh, I do not like this UX:

Without prefix:
```sh
rpk security acl create --allow-principal myuser ...
rpk security acl list
PRINCIPAL                   HOST  RESOURCE-TYPE  RESOURCE-NAME  RESOURCE-PATTERN-TYPE  OPERATION  PERMISSION  ERROR
User:myuser                 *     TOPIC          model-         PREFIXED               ALL        ALLOW
```
With prefix
```sh
rpk security acl create --allow-principal User:myuser ...
rpk security acl list
PRINCIPAL                   HOST  RESOURCE-TYPE  RESOURCE-NAME  RESOURCE-PATTERN-TYPE  OPERATION  PERMISSION  ERROR
User:myuser                 *     TOPIC          model-         PREFIXED               ALL        ALLOW
```
Principal that contains the `User:` prefix:
```sh
rpk security acl create --allow-principal User:User:myuser ...
rpk security acl list
PRINCIPAL                   HOST  RESOURCE-TYPE  RESOURCE-NAME  RESOURCE-PATTERN-TYPE  OPERATION  PERMISSION  ERROR
User:User:myuser            *     TOPIC          model-         PREFIXED               ALL        ALLOW
```
The prefix isn't needed because to get a role, you ask for a role:
```sh
rpk security acl create --allow-role myrole ...
rpk security acl list
PRINCIPAL                   HOST  RESOURCE-TYPE  RESOURCE-NAME  RESOURCE-PATTERN-TYPE  OPERATION  PERMISSION  ERROR
RedpandaRole:myrole         *     TOPIC          model-         PREFIXED               ALL        ALLOW
```

Feediver1:
@sago2k8 Do you plan to update this PR? It has been sitting for a while with no activity. Please let us know how you wish to proceed. Thanks.

modules/manage/partials/authentication.adoc
sago2k8: Should we open a follow up ticket for this, it is indeed confusing.

sago2k8:
> @sago2k8 Do you plan to update this PR? It has been sitting for a while with no activity. Please let us know how you wish to proceed. Thanks.

I have updated it could you review it ?

modules/manage/partials/authentication.adoc
```diff
@@ -63,9 +63,10 @@ You assign ACLs and superuser roles to principals, not users. Even if a user exi
 
 For example, with OIDC:
 
-- If your token's `sub` claim is `example@company.com`, then:
-  * In the `superusers` list, specify the principal as `example@company.com`
-  * In ACL rules, specify the principal as `user:OIDC:example@company.com`
+If your OIDC token's sub (subject) claim is example@company.com, use that exact sub claim value for superusers configuration and ACL rules.
+
+Unlike SASL users which require the `User:` prefix, OIDC principals use the value extracted from the token's claims (according to the configured principal mapping) without any prefix
```
paulohtb6:
```suggestion
Configure superusers and ACL rules using the exact value from your OIDC token's sub (subject) claim. For example, if your token's sub claim is example@company.com, use that exact value in your configuration.

OIDC principals use the value extracted from the token's claims according to your configured principal mapping. SASL users require the User: prefix, but OIDC principals do not use any prefix.
```

paulohtb6 (approved):

PR #1315, opened 2025-08-15
http-proxy: Conditionalize links to Self-Managed topics

JakeSCahill:
## Description

This pull request updates the HTTP Proxy documentation to improve clarity and tailor content for different deployment environments. The main changes reorganize and conditionally display authentication details based on whether the documentation is being built for a cloud environment.

Documentation improvements:

* Clarified the explanation of how authentication methods are determined, and made the reference to the `authentication_method` broker property conditional so it only appears in non-cloud builds.
* Made the reference to OIDC authentication configuration details conditional, so it is only included in non-cloud builds.

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


Feediver1 (approved):
lgtm

PR #1356, opened 2025-09-12
node_id_overrides can be removed after use

vuldin:
## Description

Review deadline:

## Page previews

https://deploy-preview-1356--redpanda-docs-preview.netlify.app/current/reference/properties/broker-properties/#node_id_overrides

## Checks

- [ ] New feature
- [ ] Content gap
- [X] Support Follow-up

### DIFF
--- a/modules/reference/pages/properties/broker-properties.adoc
+++ b/modules/reference/pages/properties/broker-properties.adoc
@@ -475,9 +475,9 @@ Broker IDs are immutable. After a broker joins the cluster, its `node_id` *canno
 
 === node_id_overrides
 
-List of broker IDs and UUID to override at broker startup. Each entry includes the current UUID and desired ID and UUID. Each entry applies to a given broker only if 'current' matches that broker's current UUID.
+List of node ID and UUID overrides applied at broker startup. Each entry includes the current UUID, desired new ID and UUID, and an ignore flag. An entry applies only if 'current_uuid' matches the broker's actual UUID.
 
-Once a broker has been restarted and the cluster is up and healthy with this section in place, then this section is not needed for subsequent restarts and can be removed from the config if desired (for example, if you need to have identical config files across each broker within a cluster).
+Remove this property after the cluster restarts successfully and operates normally. This prevents reapplication and maintains consistent configuration across brokers.
 
 *Visibility:* `user`
 
--- a/modules/reference/pages/properties/broker-properties.adoc
+++ b/modules/reference/pages/properties/broker-properties.adoc
@@ -475,7 +475,7 @@ Broker IDs are immutable. After a broker joins the cluster, its `node_id` *canno
 
 === node_id_overrides
 
-List of node ID and UUID overrides applied at broker startup. Each entry includes the current UUID, desired new ID and UUID, and an ignore flag. An entry applies only if 'current_uuid' matches the broker's actual UUID.
+List of node ID and UUID overrides applied at broker startup. Each entry includes the current UUID, the desired new ID and UUID, and an ignore flag. An entry applies only if `current_uuid` matches the broker's actual UUID.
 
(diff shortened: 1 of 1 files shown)


modules/reference/pages/properties/broker-properties.adoc
```diff
@@ -477,6 +477,8 @@ Broker IDs are immutable. After a broker joins the cluster, its `node_id` *canno
 
 List of broker IDs and UUID to override at broker startup. Each entry includes the current UUID and desired ID and UUID. Each entry applies to a given broker only if 'current' matches that broker's current UUID.
 
+Once a broker has been restarted and the cluster is up and healthy with this sectio in place, then this section is not needed for subsequent restarts and can be removed from the config if desired (for example, if you need to have identical config files across each broker within a cluster).
```
vuldin:
```suggestion
Once a broker has been restarted and the cluster is up and healthy with this section in place, then this section is not needed for subsequent restarts and can be removed from the config if desired (for example, if you need to have identical config files across each broker within a cluster).
```

vuldin (commented):
Thanks @paulohtb6, this is more readable than the initial changes.

modules/reference/pages/properties/broker-properties.adoc
```diff
@@ -475,7 +475,9 @@ Broker IDs are immutable. After a broker joins the cluster, its `node_id` *canno
 
 === node_id_overrides
 
-List of broker IDs and UUID to override at broker startup. Each entry includes the current UUID and desired ID and UUID. Each entry applies to a given broker only if 'current' matches that broker's current UUID.
+List of node ID and UUID overrides applied at broker startup. Each entry includes the current UUID, desired new ID and UUID, and an ignore flag. An entry applies only if 'current_uuid' matches the broker's actual UUID.
```
micheleRP:
```suggestion
List of node ID and UUID overrides applied at broker startup. Each entry includes the current UUID, the desired new ID and UUID, and an ignore flag. An entry applies only if `current_uuid` matches the broker's actual UUID.
```

modules/reference/pages/properties/broker-properties.adoc
```diff
@@ -1346,4 +1348,4 @@ Replace the following placeholders with your values:
 * `<kafka-broker-1>`, `<kafka-broker-2>`: IP addresses or hostnames of your Kafka brokers
 * `<kafka-port>`: Port number for the Kafka API (typically 9092)
 
-NOTE: Audit Log Client uses the same configuration properties as HTTP Proxy Client but with different defaults optimized for audit logging operations. The client uses immediate batching (0ms delay, 0 record count) with compression enabled (zstd) and acknowledgment level 1 for reliable audit log delivery.
\ No newline at end of file
+NOTE: Audit Log Client uses the same configuration properties as HTTP Proxy Client but with different defaults optimized for audit logging operations. The client uses immediate batching (0ms delay, 0 record count) with compression enabled (zstd) and acknowledgment level 1 for reliable audit log delivery.
```
micheleRP:
```suggestion
NOTE: Audit log client uses the same configuration properties as HTTP Proxy client but with different defaults optimized for audit logging operations. The client uses immediate batching (0 ms delay, 0 record count) with compression enabled (zstd) and acknowledgment level 1 for reliable audit log delivery.
```

micheleRP: @paulohtb6 just noticed that this page doesn't follow our standard for sentence case headings. I think we should update headings to HTTP-based authentication, HTTP Proxy client, Schema Registry client, and Audit log client.

paulohtb6: Thanks. I'll create a new issue to tackle that.

paulohtb6 (approved):

PR #1391, opened 2025-10-16
[25.3] Import mode for Schema Registry

kbatuigas:
## Description

Related: [API reference PR](https://github.com/redpanda-data/api-docs/pull/28)

This pull request updates the Schema Registry API documentation to clarify and expand the information about supported modes, including new instructions and use cases for disaster recovery and migration. The changes introduce the new `IMPORT` mode, reorganize documentation for clarity, and add examples for both `rpk` and `curl` commands.

**Schema Registry Modes and Usage**

* Added support and documentation for the new `IMPORT` mode, explaining its use for migration scenarios, including bypassing compatibility checks and specifying schema IDs and versions. [[1]](diffhunk://#diff-0de0ae8002f8769f123c123aece06a0c4611f6f8a03aaf31499811d6a2bcefceL979-R1054) [[2]](diffhunk://#diff-0de0ae8002f8769f123c123aece06a0c4611f6f8a03aaf31499811d6a2bcefceR1066-R1142)
* Clarified the distinction between `READWRITE`, `READONLY`, and `IMPORT` modes, and reorganized the documentation to provide clear guidance on when to use each mode.

**Command Examples and Instructions**

* Introduced tabbed examples for both `rpk` and `curl` commands throughout the documentation, making it easier for users to find and use the correct commands for their setup. [[1]](diffhunk://#diff-0de0ae8002f8769f123c123aece06a0c4611f6f8a03aaf31499811d6a2bcefceL979-R1054) [[2]](diffhunk://#diff-0de0ae8002f8769f123c123aece06a0c4611f6f8a03aaf31499811d6a2bcefceR1066-R1142)
* Added detailed instructions for setting global and subject-level modes, including how to force enable import mode and register schemas with specific IDs and versions.

**Disaster Recovery and Migration Guidance**

* Provided dedicated sections explaining how to use read-only mode for disaster recovery and import mode for migration, including the necessary permissions and version-specific behavior changes.

Review deadline: 16 Oct

## Page previews

Use the Schema Registry API > [Set Schema Registry mode](https://deploy-preview-1391--redpanda-docs-preview.netlify.app/25.3/manage/schema-reg/schema-reg-api/#set-schema-registry-mode)

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


Feediver1 (approved):
lgtm

modules/manage/pages/schema-reg/schema-reg-api.adoc
```diff
@@ -976,38 +976,82 @@ Curl::
...
+--
+
+Curl::
++
 ```bash
 curl http://localhost:8081/mode
 ```
+====
 
 === Set global mode
```
mattschumpert, line 1020: @kbatuigas Suggest for more clarity:

"Set the mode for the registry at a global level.  This mode will apply to all subjects"

modules/manage/pages/schema-reg/schema-reg-api.adoc
```diff
@@ -1019,9 +1063,83 @@ This request returns the mode that is enforced. If the subject is set to a speci
...
+Curl::
++
 ```bash
 curl -X PUT -H "Content-Type: application/vnd.schemaregistry.v1+json" --data '{"mode": "READONLY"}' http://localhost:8081/mode/<subject>
 ```
+====
+
+=== Use READONLY mode for disaster recovery
+
+A read-only Schema Registry does not accept direct writes. An active production cluster can replicate schemas to a read-only Schema Registry to keep it in sync, for example using Redpanda's https://github.com/redpanda-data/schema-migration/[Schema Migration tool^]. Users in the disaster recovery (DR) site cannot update schemas directly, so the DR cluster has an exact replica of the schemas in production. In a failover due to a disaster or outage, you can set Schema Registry to read-write mode, taking over for the failed cluster and ensuring availability.
```
mattschumpert, line 1099: Maybe we rather want to recommend Redpanda Migrator here @pgellert @hcoyote ? 

https://docs.redpanda.com/redpanda-connect/cookbooks/redpanda_migrator/

modules/manage/pages/schema-reg/schema-reg-api.adoc
```diff
@@ -1019,9 +1063,83 @@ This request returns the mode that is enforced. If the subject is set to a speci
...
+
+=== Use IMPORT mode for migration
+
+Set the target Schema Registry to import mode to:
+
+- Bypass compatibility checks when registering schemas.
+- Specify a specific schema ID and version for the registered schema, so you can retain the same IDs and version from the original Schema Registry and keep topic data associated with the correct schema.
+
+To enable import mode, you need either superuser access, or a Schema Reigstry ACL with the `alter_configs` operation on the `registry` resource. See xref:manage:schema-reg/schema-reg-authorization.adoc#enable-schema-registry-authorization[Enable Schema Registry Authorization] to learn how to enable schema registry authorization for your cluster.
```
mattschumpert: Nice!!

modules/manage/pages/schema-reg/schema-reg-api.adoc
```diff
@@ -1019,9 +1063,83 @@ This request returns the mode that is enforced. If the subject is set to a speci
...
+
+[CAUTION]
+====
+In Redpanda versions before 25.3, you could specify a schema ID or version when registering a schema in read-write mode.
+
+Starting with 25.3, read-write mode returns an error when you try to register a schema with a specific ID or version. If you have custom scripts that rely on the ability to specify an ID or version with Redpanda 25.2 and earlier, you must do either of the following:
```
mattschumpert: @kbatuigas  this also needs to be included in the 'upgrade guide' noting incompatible changes in 25.3.

mattschumpert: cc @Feediver1 @Jeffail @david-yu @hcoyote  we need to ensure Migrator adapts to this as cleanly as possible (Migrator knows _when_ to start using IMPORT mode so imports don't just start failing in readwrite mode when customers with active Migrator pipelines upgrade).

modules/manage/pages/schema-reg/schema-reg-api.adoc
```diff
@@ -1019,9 +1063,83 @@ This request returns the mode that is enforced. If the subject is set to a speci
...
+
+* Omit the ID and version fields when registering a schema. The schema will be registered under a new ID and version.
+* Change the Schema Registry or the subject to import mode.
+====
+
+To force enable import mode:
```
mattschumpert: @kbatuigas the meaning of force isn't explained here. What's the different between force and not?

modules/manage/pages/schema-reg/schema-reg-api.adoc
```diff
@@ -1019,9 +1063,83 @@ This request returns the mode that is enforced. If the subject is set to a speci
...
+
+To register a schema with a specific ID and version:
+
+[tabs]
+====
+rpk::
++
+--
+```bash
+rpk registry schema create <subject-name> --schema order.proto --id 1 --schema-version 4
```
mattschumpert, line 1140: @kbatuigas I think this is the more important piece and the base case for how to use import mode, so probably should go above the 'force' option.  I was a bit lost following this with the big yellow box and force in the middle of it all.   I think we should clarify the workflow here 'With import mode enabled, you can now register a schema explicitly setting the schema ID and version:"

mattschumpert (approved):

modules/get-started/pages/release-notes/redpanda.adoc
```diff
@@ -7,68 +7,6 @@ This topic includes new content added in version {page-component-version}. For a
...
-- xref:reference:properties/cluster-properties.adoc#log_compaction_merge_max_ranges[`log_compaction_merge_max_ranges`]
-- xref:reference:properties/cluster-properties.adoc#log_compaction_merge_max_segments_per_range[`log_compaction_merge_max_segments_per_range`]
-- xref:reference:properties/cluster-properties.adoc#log_compaction_pause_use_sliding_window[`log_compaction_pause_use_sliding_window`]
-- xref:reference:properties/cluster-properties.adoc#max_compaction_lag_ms[`max_compaction_lag_ms`]
-- xref:reference:properties/cluster-properties.adoc#min_compaction_lag_ms[`min_compaction_lag_ms`]
-- xref:reference:properties/cluster-properties.adoc#schema_registry_enable_authorization[`schema_registry_enable_authorization`]
-- xref:reference:properties/cluster-properties.adoc#topic_label_aggregation_limit[`topic_label_aggregation_limit`]
+Redpanda Schema Registry now supports an import mode that allows you to import existing schemas and retain their current IDs and version numbers. Import mode is useful when migrating from another schema registry. See xref:manage:schema-reg/schema-reg-api.adoc#set-schema-registry-mode[Use the Schema Registry API] for more information.
```
pgellert: I think we should call out here the breaking change that previously READWRITE mode allowed specifying the id + version of the schema when creating the schema, but now READWRITE mode no longer allows that. Maybe a sentence at the end:
"Going forward, IMPORT mode must be used when importing schemas. READWRITE mode no longer allows specifying a schema id and schema version when creating a schema."

modules/manage/pages/schema-reg/schema-reg-api.adoc
```diff
@@ -976,38 +976,82 @@ Curl::
...
+rpk registry mode get <subject-name>
+```
+--
+
+Curl::
++
 ```bash
 curl http://localhost:8081/mode/<subject>
 ```
+====
```
pgellert, line 1066: `rpk registry mode get <subject-name>` already always uses the "defaultToGlobal=true" parameter under the hood. So the surrounding text now needs updating. Maybe the first command could be:

To look up the applicable mode for the subject, you can use:
```
rpk registry mode get <subject-name>
curl http://localhost:8081/mode/<subject>?defaultToGlobal=true
```


This request returns the mode that is enforced. If the subject is set to a specific mode (to override the global mode), it will return the override mode, otherwise it returns the global mode.

To retrieve the subject-level override if it exists, use:
```
curl http://localhost:8081/mode/<subject>
```

This request returns an error if there is no specific mode set for the subject

modules/manage/pages/schema-reg/schema-reg-api.adoc
pgellert: @hcoyote do we have any good customer examples of this by any chance?

I don't know how well-maintained https://github.com/redpanda-data/schema-migration/ is at the moment, that's owned by CS as far as I know. The problem with linking to RPCN migrator here is that its schema_registry output requires the target cluster to be in IMPORT mode. So this described example of the target cluster being in READONLY mode only works if the replication is done as a replication of the topic, rather than a continuous schema registry import.

We could still link to RPCN (https://docs.redpanda.com/redpanda-connect/cookbooks/redpanda_migrator/), but it might be good to have a blog post / docs page about how this is set up in practice if we have some good examples.

I think DR will also require the target cluster's SR to be in READONLY mode. So maybe it would be even better to link to some Redpanda DR docs here?

modules/manage/pages/schema-reg/schema-reg-api.adoc
pgellert: +1, @kbatuigas I think you could just include this explanation here: https://redpandadata.atlassian.net/wiki/spaces/CORE/pages/1345224705/Schema+Registry+IMPORT+mode

modules/manage/pages/schema-reg/schema-reg-api.adoc
kbatuigas: Added

modules/manage/pages/schema-reg/schema-reg-api.adoc
kbatuigas: Updated, thanks!

modules/manage/pages/schema-reg/schema-reg-api.adoc
kbatuigas: Updated to move admonition closer to the beginning of the [Set mode section](https://github.com/redpanda-data/docs/pull/1391/files?new_files_changed=true#diff-0de0ae8002f8769f123c123aece06a0c4611f6f8a03aaf31499811d6a2bcefceR989).

pgellert (approved):
Looks great, thanks!

modules/manage/pages/schema-reg/schema-reg-api.adoc
hcoyote: Responded in Slack DMs.

PR #1449, opened 2025-11-11
DOC-1197: Auto-Enable Features on First License

Feediver1:
## Description

Resolves https://redpandadata.atlassian.net/browse/DOC-1197
Review deadline: **Thursday, Nov 13th**

## Page previews

[Preview](https://deploy-preview-1449--redpanda-docs-preview.netlify.app/25.3/get-started/licensing/overview/#self-managed)

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/get-started/pages/licensing/overview.adoc
+++ b/modules/get-started/pages/licensing/overview.adoc
@@ -112,7 +112,7 @@ The following table lists the enterprise features for Redpanda and how Redpanda 
 | xref:manage:cluster-maintenance/continuous-data-balancing.adoc[Continuous Data Balancing]
 | Automatically balances partitions across a cluster to optimize resource use and performance.
 
-Continuous Data Balancing is enabled by default for all new clusters, regardless of the license type.
+Continuous Data Balancing is enabled by default for all new clusters with valid licenses.
 
 | Continuous balancing is disabled, reverting to `node_add` setting that balances partitions only after a broker is added to the cluster.
 
--- a/modules/get-started/pages/licensing/overview.adoc
+++ b/modules/get-started/pages/licensing/overview.adoc
@@ -119,9 +119,10 @@ Continuous Data Balancing is enabled by default for all new clusters, regardless
 | xref:manage:cluster-maintenance/cluster-balancing.adoc#intra-broker-partition-balancing[Continuous Intra-Broker Partition Balancing]
 (`core_balancing_continuous`)
 | Balances partition replicas across CPU cores in an individual broker to optimize disk space usage.
-| Continuous Intra-Broker Partition Balancing is disabled.
 
 Continuous Intra-Broker Partition Balancing is enabled by default for all new clusters, regardless of the license type. If your license expires and you apply a new license, this feature is re-enabled.
+
+| Continuous Intra-Broker Partition Balancing is disabled.
 
 | xref:manage:security/fips-compliance.adoc[FIPS Compliance]
 | Enables compliance with FIPS security standards for cryptography.
--- a/modules/get-started/pages/licensing/overview.adoc
+++ b/modules/get-started/pages/licensing/overview.adoc
@@ -114,7 +114,7 @@ The following table lists the enterprise features for Redpanda and how Redpanda 
 
 Continuous Data Balancing is enabled by default for all new clusters, regardless of the license type. If your license expires and you apply a new license, this feature is re-enabled.
 
-| Continuous balancing is disabled, reverting to the default `node_add` setting that balances partitions only after a broker is added to the cluster.
+| Continuous balancing is disabled, reverting to `node_add` setting that balances partitions only after a broker is added to the cluster.
 
 | xref:manage:cluster-maintenance/cluster-balancing.adoc#intra-broker-partition-balancing[Continuous Intra-Broker Partition Balancing]
 (`core_balancing_continuous`)
--- a/modules/get-started/pages/licensing/overview.adoc
+++ b/modules/get-started/pages/licensing/overview.adoc
@@ -120,7 +120,7 @@ Continuous Data Balancing is enabled by default for all new clusters with valid 
 (`core_balancing_continuous`)
 | Balances partition replicas across CPU cores in an individual broker to optimize disk space usage.
 
-Continuous Intra-Broker Partition Balancing is enabled by default for all new clusters with valid a valid license.
+Continuous Intra-Broker Partition Balancing is enabled by default for all new clusters with a valid license.
 
 | Continuous Intra-Broker Partition Balancing is disabled.
 
--- a/modules/get-started/pages/licensing/overview.adoc
+++ b/modules/get-started/pages/licensing/overview.adoc
@@ -73,7 +73,6 @@ For more information, see:
   https://docs.redpanda.com/current/get-started/licensing/overview/#license-keys
 ```
 
-
 === License keys
 
 Enterprise features require an Enterprise Edition license key, sometimes called enterprise license, license key, or license.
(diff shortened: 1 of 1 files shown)


modules/get-started/pages/licensing/overview.adoc
```diff
@@ -112,11 +111,17 @@ The following table lists the enterprise features for Redpanda and how Redpanda
 
 | xref:manage:cluster-maintenance/continuous-data-balancing.adoc[Continuous Data Balancing]
 | Automatically balances partitions across a cluster to optimize resource use and performance.
+
+Continuous Data Balancing is enabled by default for all new clusters, regardless of the license type. If your license expires and you apply a new license, this feature is re-enabled.
```
IoannisRP: > If your license expires and you apply a new license, this feature is re-enabled.

I am not sure about this sentence. It is correct, but it is the behavior for all enterprise features.
IMO, pointing it out here raises questions about the behavior of all the other enterprise features.

Same for Continuous Intra-Broker Partition Balancing.

modules/get-started/pages/licensing/overview.adoc
```diff
@@ -112,11 +111,17 @@ The following table lists the enterprise features for Redpanda and how Redpanda
...
+
 | Continuous balancing is disabled, reverting to the default `node_add` setting that balances partitions only after a broker is added to the cluster.
```
IoannisRP:
```suggestion
| Continuous balancing is disabled, reverting to `node_add` setting that balances partitions only after a broker is added to the cluster.
```
This value is no longer the default

modules/get-started/pages/licensing/overview.adoc
Feediver1: @mattschumpert We were discussing this yesterday. Did I capture this correctly? Thoughts on keeping in, revising, removing?

modules/get-started/pages/licensing/overview.adoc
Feediver1: @IoannisRP Is there a new default? Or should it just be: Continuous balancing is disabled?

IoannisRP: the new default value is `continuous`.
This corresponds to the enterprise feature "continuous balancing" being enabled, instead of just balancing on "node add"

modules/get-started/pages/licensing/overview.adoc
mattschumpert: @Feediver1  @IoannisRP the difference between 'enabled' and 'activated' is probably relevant here. All features again become available for use, but IIUC @IoannisRP continuously data balancing and partition core balancing are actually re-activated automatically (set to continuous mode) once a new license is applied (after previous expiry). correct?

IoannisRP: Being pedantic, once a valid license is re-applied, the restrictions are lifted and the behavior returns to what the configuration dictates (this applies to all enterprise features).

The wording is not wrong. But the same sentence can be used on every other enterprise feature with the exact same meaning. So, only adding it here raises more questions than it answers. At least, that is how I read it.

An important detail on how restrictions are applied, is that the configuration is never changed. If you have balancing set to `continuous`, it is never explicitly disabled by changing the config to `node_add`. While it is set to `continuous` and there is no valid license, the balancing behavior is restricted to that of `node_add`.

modules/get-started/pages/licensing/overview.adoc
```diff
@@ -112,11 +111,17 @@ The following table lists the enterprise features for Redpanda and how Redpanda
 
 | xref:manage:cluster-maintenance/continuous-data-balancing.adoc[Continuous Data Balancing]
 | Automatically balances partitions across a cluster to optimize resource use and performance.
-| Continuous balancing is disabled, reverting to the default `node_add` setting that balances partitions only after a broker is added to the cluster.
+
+Continuous Data Balancing is enabled by default for all new clusters, regardless of the license type. If your license expires and you apply a new license, this feature is re-enabled.
```
Feediver1:
```suggestion
Continuous Data Balancing is enabled by default for all new clusters, regardless of the license type.
```

modules/get-started/pages/licensing/overview.adoc
```diff
@@ -112,11 +111,17 @@ The following table lists the enterprise features for Redpanda and how Redpanda
...
+
+| Continuous balancing is disabled, reverting to `node_add` setting that balances partitions only after a broker is added to the cluster.
 
 | xref:manage:cluster-maintenance/cluster-balancing.adoc#intra-broker-partition-balancing[Continuous Intra-Broker Partition Balancing]
 (`core_balancing_continuous`)
 | Balances partition replicas across CPU cores in an individual broker to optimize disk space usage.
+
+Continuous Intra-Broker Partition Balancing is enabled by default for all new clusters, regardless of the license type. If your license expires and you apply a new license, this feature is re-enabled.
```
Feediver1:
```suggestion
Continuous Intra-Broker Partition Balancing is enabled by default for all new clusters, regardless of the license type.
```

mattschumpert (approved):

modules/get-started/pages/licensing/overview.adoc
```diff
@@ -112,11 +111,17 @@ The following table lists the enterprise features for Redpanda and how Redpanda
 
 | xref:manage:cluster-maintenance/continuous-data-balancing.adoc[Continuous Data Balancing]
 | Automatically balances partitions across a cluster to optimize resource use and performance.
-| Continuous balancing is disabled, reverting to the default `node_add` setting that balances partitions only after a broker is added to the cluster.
+
+Continuous Data Balancing is enabled by default for all new clusters, regardless of the license type.
```
paulohtb6: even invalid ones?

paulohtb6: this is a dangerous claim, we should explicitly say what types are available.

micheleRP (approved):
lgtm

modules/get-started/pages/licensing/overview.adoc
paulohtb6: Suggest one or the other here to be more defensive 

```suggestion
Continuous Data Balancing is enabled by default for all new clusters with valid licenses.
```

or being explicit 

```suggestion
Continuous Data Balancing is enabled by default for all new clusters, whether using Community or Enterprise Edition.
```

Feediver1: All new clusters come with a trial license. Not sure how it would be invalid? When the license expires, the feature (all the enterprise features) is still there, but you just cannot use that feature. Your data remains safe until you apply a valid license.

Feediver1: No--it's only for Enterprise, not Community. https://deploy-preview-1449--redpanda-docs-preview.netlify.app/25.3/get-started/licensing/overview/#community-edition

modules/get-started/pages/licensing/overview.adoc
```diff
@@ -112,11 +111,17 @@ The following table lists the enterprise features for Redpanda and how Redpanda
...
+
+Continuous Data Balancing is enabled by default for all new clusters with valid licenses.
+
+| Continuous balancing is disabled, reverting to `node_add` setting that balances partitions only after a broker is added to the cluster.
 
 | xref:manage:cluster-maintenance/cluster-balancing.adoc#intra-broker-partition-balancing[Continuous Intra-Broker Partition Balancing]
 (`core_balancing_continuous`)
 | Balances partition replicas across CPU cores in an individual broker to optimize disk space usage.
+
+Continuous Intra-Broker Partition Balancing is enabled by default for all new clusters, regardless of the license type.
```
Feediver1:
```suggestion
Continuous Intra-Broker Partition Balancing is enabled by default for all new clusters with valid a valid license.
```

modules/get-started/pages/licensing/overview.adoc
```diff
@@ -112,11 +111,17 @@ The following table lists the enterprise features for Redpanda and how Redpanda
...
+
+Continuous Data Balancing is enabled by default for all new clusters with valid licenses.
+
+| Continuous balancing is disabled, reverting to `node_add` setting that balances partitions only after a broker is added to the cluster.
 
 | xref:manage:cluster-maintenance/cluster-balancing.adoc#intra-broker-partition-balancing[Continuous Intra-Broker Partition Balancing]
 (`core_balancing_continuous`)
 | Balances partition replicas across CPU cores in an individual broker to optimize disk space usage.
+
+Continuous Intra-Broker Partition Balancing is enabled by default for all new clusters with valid a valid license.
```
Feediver1:
```suggestion
Continuous Intra-Broker Partition Balancing is enabled by default for all new clusters with a valid license.
```

PR #1513, opened 2025-12-11
Doc-1601: Specify cluster UUID to restore with Whole Cluster Recovery

Feediver1:
## Description

Resolves https://redpandadata.atlassian.net/browse/DOC-1601

IIUC, this feature is for Redpanda Support. I included more detail to help Support when using this config. If not solely for support, I can remove all the explanations and just include a paragraph with a description of the config and what it does, along with a note telling users not to use it without contacting RP Support first. 

Review deadline: Dec 19th

## Page previews
(https://deploy-preview-1513--redpanda-docs-preview.netlify.app/current/manage/disaster-recovery/whole-cluster-restore/#advanced-restore-data-when-multiple-clusters-share-data)

## Checks

- [x ] New feature
- [ ] Content gap
- [ ] Support Follow-up


mattschumpert:
@Feediver1 

1. Terminology needs adjustment here.  The heading is not 'when multiple clusters share data'  its' when data from multiple clusters exists in the same bucket/container'
2. Content is quite useful but I think this should be moved to an internal Wiki page (not in public docs), without any objection from @nvartolomei , @andrwng

nvartolomei:
@mattschumpert i'm fine with public doc unless it is a business/product concern. My argument is that our internal teams consume the same docs. There isn't anything to hide either. We already have this documented in the code.

mattschumpert:
Up to you.  We should be clear in the docs though this only occurs in extreme cases where multiple clusters have accidentally been sharing a bucket due to misconfiguration or a cluster lost quorum etc.  I leave it to you / @andrwng  to sort this out with @Feediver1

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -225,3 +226,97 @@ NODE  CONFIG-VERSION  NEEDS-RESTART  INVALID  UNKNOWN
...
 When the cluster restore is successfully completed, you can redirect your application workload to the new cluster. Make sure to update your application code to use the new addresses of your brokers.
+
+== Advanced: Restore data when multiple clusters share data
+
+[CAUTION]
+====
+This is an advanced use case that should be performed only by Redpanda support.
+====
+
+Typically, you will have a one-to-one mapping between a Redpanda cluster and its object storage bucket. However, it's possible to run multiple clusters that share the same storage bucket. Sharing an object storage bucket allows you to move tenants between clusters without moving data. For example, you might wish to mount topics to multiple clusters in the same bucket without having to move data.
```
nvartolomei: > For example, you might wish to mount topics to multiple clusters in the same bucket without having to move data.

`you might wish to move topics (unmount on cluster A, mount on cluster B)`

nvartolomei (approved):
That's beyond what I would have expected for us to have documented but looks great! 🤟

wzzzrd86:
Based on some questions and comments Dais made to me, shared, below, I am in favor of documenting it publicly. 

> Suppose we leave it undocumented. What happens if a customer wants to share a bucket across multiple clusters? That’s supported by design, and at some point we (eng / CS / support / whoever) will end up telling them to configure cloud_storage_cluster_name to properly distinguish the clusters, which is set by default in cloud clusters. At that point, the obvious question comes up: why isn’t this documented?
> Now let’s assume it does get configured. A year later, a different person needs to run WCR. They’re not aware of the earlier discussion (unfortunately realistic). They run the restore command exactly as documented and notice a log message mentioning cloud_storage_cluster_name. Naturally, they search the public docs to understand what it is but find nothing. The end result is confusion: “Why is this not documented anywhere if we were explicitly told to configure it?”

wzzzrd86 (approved):

modules/manage/partials/whole-cluster-restore.adoc
Feediver1:
```suggestion
Typically, you will have a one-to-one mapping between a Redpanda cluster and its object storage bucket. However, it's possible to run multiple clusters that share the same storage bucket. Sharing an object storage bucket allows you to move tenants between clusters without moving data. For example, you might wish to move topics (unmount on cluster A, mount on cluster B) to multiple clusters in the same bucket without having to move data.
```

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -225,3 +226,97 @@ NODE  CONFIG-VERSION  NEEDS-RESTART  INVALID  UNKNOWN
 endif::[]
 
 When the cluster restore is successfully completed, you can redirect your application workload to the new cluster. Make sure to update your application code to use the new addresses of your brokers.
+
+== Advanced: Restore data when multiple clusters share data
```
Feediver1:
```suggestion
== Advanced: Restore data from multiple clusters sharing the same bucket
```

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -53,6 +53,7 @@ By default, Redpanda uploads cluster metadata to object storage periodically. Yo
 * xref:reference:cluster-properties.adoc#enable_cluster_metadata_upload_loop[`enable_cluster_metadata_upload_loop`]: Enable metadata uploads. This property is enabled by default and is required for Whole Cluster Restore.
 * xref:reference:properties/object-storage-properties.adoc#cloud_storage_cluster_metadata_upload_interval_ms[`cloud_storage_cluster_metadata_upload_interval_ms`]: Set the time interval to wait between metadata uploads.
 * xref:reference:cluster-properties.adoc#controller_snapshot_max_age_sec[`controller_snapshot_max_age_sec`]: Maximum amount of time that can pass before Redpanda attempts to take a controller snapshot after a new controller command appears. This property affects how current the uploaded metadata can be.
+* xref:reference:properties/object-storage-properties.adoc#cloud_storage_cluster_name[`cloud_storage_cluster_name`]: *Advanced: This is an internal-only configuration and should be enabled only after consulting with Redpanda support.* Specify a custom name for cluster's metadata in object storage. For use when multiple clusters share the same storage bucket (for example, for Whole Cluster Restore).
```
paulohtb6:
```suggestion
* xref:reference:properties/object-storage-properties.adoc#cloud_storage_cluster_name[`cloud_storage_cluster_name`]: This is an internal-only configuration and should be enabled only after consulting with Redpanda support. Specify a custom name for cluster's metadata in object storage. For use when multiple clusters share the same storage bucket (for example, for Whole Cluster Restore).
```

we don't use this bolding elsewhere. Suggest to remove it.

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -12,14 +12,14 @@ endif::[]
 include::shared:partial$enterprise-license.adoc[]
 ====
 
-With xref:{link-tiered-storage}[Tiered Storage] enabled, you can use Whole Cluster Restore to restore data from a failed cluster (source cluster), including its metadata, onto a new cluster (target cluster). This is a simpler and cheaper alternative to active-active replication, for example with xref:migrate:data-migration.adoc[MirrorMaker 2]. Use this recovery method to restore your application to the latest functional state as quickly as possible.
+With xref:{link-tiered-storage}[Tiered Storage] enabled, you can use Whole Cluster Restore to restore data from a failed cluster (source cluster you are restoring from), including its metadata, onto a new cluster (target cluster you are restoring to). This is a simpler and cheaper alternative to active-active replication, for example with xref:migrate:data-migration.adoc[MirrorMaker 2]. Use this recovery method to restore your application to the latest functional state as quickly as possible.
```
paulohtb6, line 15: this change looks redudant. I would revert.

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -225,3 +226,97 @@ NODE  CONFIG-VERSION  NEEDS-RESTART  INVALID  UNKNOWN
 endif::[]
 
 When the cluster restore is successfully completed, you can redirect your application workload to the new cluster. Make sure to update your application code to use the new addresses of your brokers.
+
+== Advanced: Restore data from multiple clusters sharing the same bucket
```
paulohtb6:
```suggestion
== Restore data from multiple clusters sharing the same bucket
```

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -225,3 +226,97 @@ NODE  CONFIG-VERSION  NEEDS-RESTART  INVALID  UNKNOWN
...
+
+[CAUTION]
+====
+This is an advanced use case that should be performed only by Redpanda support.
```
paulohtb6: Suggest we keep it consistent with the phrasing that we had above 

```suggestion
This is an advanced use case that should be performed only after consulting with Redpanda support.
```

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -225,3 +226,97 @@ NODE  CONFIG-VERSION  NEEDS-RESTART  INVALID  UNKNOWN
...
+
+=== Configure cluster names for multiple source clusters
+
+To disambiguate cluster metadata from multiple clusters, use the xref:reference:properties/object-storage-properties.adoc#cloud_storage_cluster_name[`cloud_storage_cluster_name`] property (off by default), which allows you to assign a unique name to each cluster sharing the same object storage bucket. Redpanda uses this name to organize the cluster metadata within the shared object storage bucket. This ensures that each cluster's data remains distinct and prevents conflicts during recovery operations.The name must be unique within the bucket, 1-64 characters, and use only letters, numbers, underscores, and hyphens. Do not change this value once set. After setting, your object storage bucket organization may look like the following:
```
paulohtb6, line 255: > do not change this value once set. 

Or what? What problems can this present? Is it recoverable? When the new name is set is it immediately used?

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -225,3 +226,97 @@ NODE  CONFIG-VERSION  NEEDS-RESTART  INVALID  UNKNOWN
...
+|  | +- 2/cluster_manifest.json
+|  + <uuid-b>/manifests/
+|    +- 0/cluster_manifest.json
+|    +- 1/cluster_manifest.json # lost cluster
++- cluster_name/
+   +- rp-foo/uuid/<uuid-a>
+   +- rp-qux/uuid/<uuid-b>
+----
+
+During a Whole Cluster Restore, Redpanda looks for the cluster name specified in `cloud_storage_cluster_name` and only consider manifests associated with that name. Because the name specified here is `rp-qux`, Redpanda only considers manifests for the clusters `<uuid-b>` and `<uuid-c>`, ignoring cluster `<uuid-a>` entirely. In this case, your object storage bucket may look like the following:
```
paulohtb6: What name for cloud_storage_cluster_name are we using here as an example? Is it rp-qux? If yes we could make this more clear. 
UUID-C comes out of nowhere so we can explain that it's the new cluster in the narrative doc too

modules/manage/partials/whole-cluster-restore.adoc
```diff
@@ -225,3 +226,97 @@ NODE  CONFIG-VERSION  NEEDS-RESTART  INVALID  UNKNOWN
...
+   | +- 0/cluster_manifest.json
+   | +- 1/cluster_manifest.json #lost cluster
+   + <uuid-b>/manifests/
+     +- 3/cluster_manifest.json # lost again (not recovered)
+   + <uuid-d>/manifests/
+      +- 7/cluster_manifest.json # new attempt to recover uuid-b
+                                 # it does not have the data
+----
+
+In such cases, you can explicitly run a POST request using the Admin API:
```
paulohtb6, line 313:
```suggestion
In such cases, you can explicitly run a POST request using the Admin API to force Redpanda to use a specific UUID:
```

paulohtb6: this is only possible with Admin API? Any other ways to do this?

paulohtb6 (approved):

modules/manage/partials/whole-cluster-restore.adoc
Feediver1: update was requested by PM so it's a bit redundant, but super clear.

modules/manage/partials/whole-cluster-restore.adoc
Feediver1:
```suggestion
During a Whole Cluster Restore, Redpanda looks for the cluster name specified in `cloud_storage_cluster_name` and only consider manifests associated with that name. Because the cluster name specified here is `rp-qux`, Redpanda only considers manifests for the clusters `<uuid-b>` and `<uuid-c>` (another new cluster sharing the bucket), ignoring cluster `<uuid-a>` entirely. In this case, your object storage bucket may look like the following:
```

modules/manage/partials/whole-cluster-restore.adoc
Feediver1: Yes--only with the Admin API

daisukebe: Just a note: once 25.3.5 is released, we will have an equivalent rpk command available. 

```
rpk cluster storage restore start --cluster-uuid-override uuid-a
```

PR #1514, opened 2025-12-12
Add Shadowing docs for Kubernetes deployments

JakeSCahill:
## Description

Resolves https://redpandadata.atlassian.net/browse/DOC-1533
Review deadline: December 15

## Checks

- [x] New feature
- [ ] Content gap
- [ ] Support Follow-up


modules/troubleshoot/partials/errors-and-solutions.adoc
```diff
@@ -576,3 +576,249 @@ If you're using `rpk`, ensure to specify the `-X user`, `-X pass`, and `-X sasl.
...
+
+* Verify SASL configuration if source cluster requires authentication
+* Test network connectivity: `kubectl exec` into shadow pod and try connecting to source cluster
+* Check Kubernetes NetworkPolicies and firewall rules
+//end::shadow-link-monitoring[]
+
+//tag::shadow-link-failover[]
+=== Topics stuck in FAILING_OVER state
+
+When using Helm, topics may remain in `FAILING_OVER` state for extended periods.
```
andrewstucki:
```suggestion
When using the operator, topics may remain in `FAILING_OVER` state for extended periods.
```
I'm also not entirely sure about this. What is this in reference to? Is this talking about on a ShadowLink CRD? If not, is this talking about manually initialized topics? If so, it shouldn't be `Helm` or `operator` specific minus the rpk commands being wrapped in a `kubectl exec`.

modules/troubleshoot/partials/errors-and-solutions.adoc
```diff
@@ -576,3 +576,249 @@ If you're using `rpk`, ensure to specify the `-X user`, `-X pass`, and `-X sasl.
...
+
+Ensure sufficient cluster resources (CPU, memory, disk space) are available on the shadow cluster. Verify network connectivity between shadow cluster nodes and confirm that all shadow topic partitions have elected leaders.
+
+ifdef::env-kubernetes[]
+If topics remain stuck after addressing these cluster health issues and you need immediate failover, you can force delete the shadow link:
+
+[,bash]
+----
+kubectl exec --namespace <shadow-namespace> <shadow-pod-name> --container redpanda -- \
+  rpk shadow delete <shadow-link-name> --force
```
andrewstucki: So, this deletes and entire shadow link with forcible failover for any mirrored topics. Is this an instruction about a particular topic or about the link in general? If the former, you shouldn't delete the shadow topic but instead just promote/fail it over directly, but again, that's a core operation thing.

modules/troubleshoot/partials/errors-and-solutions.adoc
```diff
@@ -576,3 +576,249 @@ If you're using `rpk`, ensure to specify the `-X user`, `-X pass`, and `-X sasl.
...
+----
+
+Check for finalizers on the resource:
+
+[,bash]
+----
+kubectl get shadowlink --namespace <shadow-namespace> <shadowlink-name> -o yaml | grep finalizers -A 5
+----
+
+If necessary, remove the finalizer to force deletion:
```
andrewstucki: In general this is a bad idea because it means we won't clean up any side effects and the shadow link might still exist in the target cluster. At minimum if someone does this they should run the

```
rpk shadow delete <shadow-link-name> --force
```

bits above after this to make sure everything is pruned manually.

modules/troubleshoot/partials/errors-and-solutions.adoc
```diff
@@ -576,3 +576,249 @@ If you're using `rpk`, ensure to specify the `-X user`, `-X pass`, and `-X sasl.
...
+When creating a shadow link with TLS enabled, you may see:
+
+[.no-copy]
+----
+Cluster link unreachable, preflight check failed - { node: -1 }, { error_code: broker_not_available [8] }
+----
+
+The shadow cluster cannot verify the source cluster's TLS certificate. This is the most common issue when using TLS with self-signed certificates (the default for Kubernetes deployments with `tls.certs.default.caEnabled=true`).
+
+Ensure the `ca_path` in your shadow link configuration points to the source cluster's CA certificate, not the shadow cluster's own CA.
```
andrewstucki:
```suggestion
Ensure the `ca` in your shadow link configuration is the source cluster's CA certificate, not the shadow cluster's own CA.
```
We don't point to a path on disk, we pass the CA in as a PEM-formatted value directly to the shadow link configuration.

modules/troubleshoot/partials/errors-and-solutions.adoc
```diff
@@ -576,3 +576,249 @@ If you're using `rpk`, ensure to specify the `-X user`, `-X pass`, and `-X sasl.
...
+=== Could not read trust file
+
+When creating a shadow link with TLS enabled, you may see:
+
+[.no-copy]
+----
+Cluster link unreachable, preflight check failed - Could not read trust file /path/to/ca.crt
+----
+
+The `ca_path` in your shadow link configuration is incorrect or the file does not exist in the pod.
```
andrewstucki:
```suggestion
The `ca` in your shadow link configuration is incorrect or the default CA file does not exist in the pod.
```

modules/troubleshoot/partials/errors-and-solutions.adoc
```diff
@@ -576,3 +576,249 @@ If you're using `rpk`, ensure to specify the `-X user`, `-X pass`, and `-X sasl.
...
+
+For Helm deployments, the default CA certificate path is `/etc/tls/certs/default/ca.crt`. If you copied the source cluster's CA certificate to a custom location, ensure the path in your configuration matches where you copied the file.
```
andrewstucki: Not sure how this works out based on the above. Is this something you've seen? In general we only allow for setting this portion of the config:

https://github.com/redpanda-data/redpanda/blob/334c803b1afa8a220275702df34f649f38cfd9ec/proto/redpanda/core/common/v1/tls.proto#L51-L62

andrewstucki (changes requested):
Some initial comments

modules/troubleshoot/partials/errors-and-solutions.adoc
```diff
@@ -576,3 +576,249 @@ If you're using `rpk`, ensure to specify the `-X user`, `-X pass`, and `-X sasl.
...
+# Extract source cluster's CA
+kubectl exec --namespace <source-namespace> <source-pod-name> --container redpanda -- \
+  cat /etc/tls/certs/default/ca.crt > source-ca.crt
+
+# Copy to shadow cluster
+kubectl cp source-ca.crt <shadow-namespace>/<shadow-pod-name>:/tmp/source-ca.crt
+
+# Reference in shadow link config: ca_path: "/tmp/source-ca.crt"
+----
+endif::[]
```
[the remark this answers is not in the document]
andrewstucki: Also, just going to mention here, if you're using the CRDs rather than rpk directly, we don't allow you to configure shadow links this way (and you probably shouldn't since a pod restart will remove the file anyway). The proper way of doing it would be to either copy the file into a secret on the kubernetes cluster where the shadow cluster exists and reference it within a `staticConfiguration` block, or inline it directly in that block.

modules/troubleshoot/partials/errors-and-solutions.adoc
JakeSCahill: We document the use of `ca_path` and it worked in my testing: https://docs.redpanda.com/current/manage/disaster-recovery/shadowing/setup/#create-a-shadow-link

david-yu (approved):

andrewstucki (approved):
From the sounds of it, a number of my earlier feedback comments were oriented towards the operator whereas some of the troubleshooting stuff was for helm. Things generally look fine to me now.

paulohtb6:
Throughout prefer to use Shadowing instead of Shadow links.

JakeSCahill:
> Throughout prefer to use Shadowing instead of Shadow links.

ShadowLink is the name of the CRD. I have updated some places like the nav.

paulohtb6 (approved):

PR #1552, opened 2026-01-20
Add the Bump proxy for previewing API docs in PRs

JakeSCahill:
## Description

Allows us to preview API docs in preview builds by adding the proxy for Bump.sh.

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up


paulRbr:
Hi Jake,

Sorry to intrude, I've been warned by Kate that the preview doesn't seem to work well when following the links to the api docs. And indeed, the behavior isn't as we all expected (sorry for having suggested that “solution” last time we spoke).

Indeed, currently the bump.sh proxy only allows the api docs site to work on the production proxy url `docs.redpanda.com/api`. There are quite a few things in the bump.sh docs javascript which has same-origin restrictions (frames are loaded through XHR and we have a strong same-origin CORS policy, replacing the browser history is only allowed for same-origin, etc...). 

A potential solution (but it needs to be discussed internally @bump-sh because I'm not sure we really want to go in that direction) would be to allow you to specify a `X-BUMP-SH-DOCS-ORIGIN` which would contain the netlify preview hostname.

I'll come back to you with our restrictions on the bump.sh side and let's have a live chat if need be. Thanks!

Feediver1:
@kbatuigas When will you next meet with bump/plan to discuss this with them? Can you provide an update on that here? thx

JakeSCahill:
> Nice investigation, and happy that you found the issue! I'm wondering : how was this cache control not affecting the production site?

Both sites ran the exact same caching code, so the bug could have hit either one, but it only showed up on the preview because the preview was missing `BUMP_PROXY_SECRET` that Bump needs. On the real site, that value is set, so when the page asks Bump for a partial, Bump hands back the right piece and everything caches normally. On the preview, with no valid value, Bump just bounced those requests back to the main page instead, so the preview ended up saving and re-serving the whole main page in place of the smaller partials for five minutes.

Maybe worth a mention in https://docs.bump.sh/help/customization-options/embed-mode/

kbatuigas (approved):

PR #1557, opened 2026-01-23
Update brokers list outputs

daisukebe:
## Description

This PR updates the output of the `rpk redpanda admin brokers list` command in all pages that exists because they are still old. 

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/manage/pages/cluster-maintenance/decommission-brokers.adoc
+++ b/modules/manage/pages/cluster-maintenance/decommission-brokers.adoc
@@ -52,7 +52,8 @@ rpk cluster info
 .Example output
 [%collapsible]
 ====
-```
+[,bash,role="no-wrap"]
+----
 CLUSTER
 =======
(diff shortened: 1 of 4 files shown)


daisukebe:
@Feediver1, the new output is horizontally wide. Is it possible to make the `[%collapsible]` block horizontally scrollable?
image

JakeSCahill:
Thanks! You can get the code block to scroll by using this syntax instead of triple backticks:

```
[,bash,role=“no-wrap”]
- - - -
Code here
- - - -
```

don’t add spaces between the dashes. I’m currently on mobile and needed to do that to get it to render.

JakeSCahill:
I think the ideal scenario for the long-term is for us to build tests that save the `rpk` output and run on each release. This way our outputs will never become outdated like this.

JakeSCahill (approved):

PR #1564, opened 2026-02-03
DOC-1935 add glossterms for ADP

micheleRP:
## Description
This pull request introduces a comprehensive set of new and updated glossary terms for Redpanda and agentic systems, focusing on AI agents, protocols, authentication, and related infrastructure. 

* Added definitions for key agentic concepts, including `Agentic Data Plane (ADP)`, `Model Context Protocol (MCP)`, and many others. Each term includes a description, hover text, category, and cross-references to related terms.
* Updated `authentication` and `authorization` terms to include standard abbreviations (`AuthN`, `AuthZ`) and clarified their role in Redpanda security. 
* Added definition for `identity provider (IdP)`, explaining its function in authentication and token issuance, with references to related security terms.
* Fixed typo in `Kafka API` term

Resolves https://redpandadata.atlassian.net/browse/DOC-1935
Review deadline:

## Page previews
[Glossary](https://deploy-preview-1564--redpanda-docs-preview.netlify.app/25.3/reference/glossary)

## Checks

- [X] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/terms/partials/a2a-message-processor.adoc
+++ b/modules/terms/partials/a2a-message-processor.adoc
@@ -1,0 +1,8 @@
+=== a2a_message processor
+:term-name: a2a_message processor
+:hover-text: A Redpanda Connect processor that enables pipelines to invoke AI agents using the Agent2Agent protocol.
+:category: Redpanda Connect
+
+The `a2a_message` processor allows Redpanda Connect pipelines to send messages to AI agents and receive responses. This enables pipeline-initiated agent interactions where streaming data can trigger agent reasoning and tool execution as part of data processing workflows.
+
+See also: glossterm:processor[], glossterm:a2a-protocol[], glossterm:ai-agent[], glossterm:redpanda-connect[]
--- a/modules/terms/partials/a2a-message-processor.adoc
+++ b/modules/terms/partials/a2a-message-processor.adoc
@@ -1,6 +1,0 @@
-=== a2a_message processor
-:term-name: a2a_message processor
-:hover-text: A Redpanda Connect processor that enables pipelines to invoke AI agents using the Agent2Agent protocol.
-:category: Redpanda Connect
-
-The `a2a_message` processor allows Redpanda Connect pipelines to send messages to AI agents and receive responses. This enables pipeline-initiated agent interactions where streaming data can trigger agent reasoning and tool execution as part of data processing workflows.
--- a/modules/terms/partials/a2a-message-processor.adoc
+++ b/modules/terms/partials/a2a-message-processor.adoc
@@ -4,5 +4,3 @@
 :category: Redpanda Connect
 
 The `a2a_message` processor allows Redpanda Connect pipelines to send messages to AI agents and receive responses. This enables pipeline-initiated agent interactions where streaming data can trigger agent reasoning and tool execution as part of data processing workflows.
-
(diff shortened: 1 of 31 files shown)


modules/terms/partials/trace.adoc
kbatuigas, line 1: I think the idea behind having separate terms for transcript and trace is to explain that trace has a more general definition outside of Redpanda (i.e. with OpenTelemetry), and transcript is our take? interpretation? inspired by Anthropic, and is also closely tied to how we want to present this data in the UI. But will need Marc and others to chime in and confirm.

For example, if we tell users how to hook up their external agents to our otel ingestion pipeline, I think we would need to refer to both the otel trace and the ADP transcript, in that specific context the two are even less interchangeable

modules/terms/partials/authentication.adoc
```diff
@@ -1,6 +1,6 @@
-=== authentication
-:term-name: authentication
-:hover-text: The process of verifying the identity of a principal, user, or service account. 
+=== authentication (AuthN)
+:term-name: authentication (AuthN)
```
paulohtb6: not sure that works. It will prob no highlight just "authentication".


Do we have an "alias" feature where we can put multiple names for the same term?

modules/terms/partials/knowledge-base.adoc
```diff
@@ -0,0 +1,6 @@
+=== knowledge base
+:term-name: knowledge base
```
paulohtb6: maybe this should be capitalized because it will aalso be a feature?

modules/terms/partials/local-mcp-server.adoc
```diff
@@ -0,0 +1,6 @@
+=== local MCP server
+:term-name: local MCP server
+:hover-text: An MCP server running in a local development environment that connects AI agents to tools and resources on the developer's machine.
+:category: Agentic Data Plane
+
+Local MCP servers enable developers to test and debug AI agent integrations on their own machines before deploying to production. They provide the same MCP interface as remote servers but run locally with access to local files, databases, and development resources.
```
paulohtb6: not sure calling it local mcp server is a good idea. Maybe the glossterm could be just mcp server?

modules/terms/partials/a2a-message-processor.adoc
```diff
@@ -0,0 +1,6 @@
+=== a2a_message processor
```
JakeSCahill: This is a Redpanda connect component. We shouldn’t define this as a term.

modules/terms/partials/knowledge-base.adoc
micheleRP: i'm going to remove this for now, since it's now even supported yet

modules/terms/partials/local-mcp-server.adoc
micheleRP: removing local: we have MCP server defined

modules/terms/partials/a2a-message-processor.adoc
micheleRP: removing

modules/terms/partials/authentication.adoc
micheleRP: We have acronyms in other glossterms too and works fine like [here](https://deploy-preview-1564--redpanda-docs-preview.netlify.app/25.3/reference/glossary#access-control-list-acl)... but I've moved AuthZ and AuthN into the descriptions

paulohtb6 (approved):

PR #1576, opened 2026-02-17
Add gloss term for log

kbatuigas:
## Description

This pull request adds documentation for the term "log" in the Redpanda core module. The new partial provides a concise definition and context for how logs are used within Redpanda.

Documentation update:

* Added a new partial `log.adoc` defining "log" as an ordered, append-only, immutable sequence of records, and explaining its role in Redpanda's storage and replication.

Review deadline:

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up

### DIFF
--- a/modules/terms/partials/log.adoc
+++ b/modules/terms/partials/log.adoc
@@ -3,6 +3,6 @@
 :hover-text: An ordered, append-only, immutable sequence of records.
 :category: Redpanda core
 
-The log is Redpanda's core storage abstraction. Topics and partitions are implemented as logs. Logs are replicated across brokers within a Redpanda cluster for fault tolerance.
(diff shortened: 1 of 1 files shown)


modules/terms/partials/log.adoc
```diff
@@ -0,0 +1,8 @@
+=== log
+:term-name: log
+:hover-text: An ordered, append-only, immutable sequence of records.
+:category: Redpanda core
+
+The log is Redpanda's core storage abstraction. Topics and partitions are implemented as logs. Logs are replicated across brokers within a Redpanda cluster for fault tolerance.
```
Feediver1: From Claude Code:

Log (Redpanda Core Definition)                                                                                                                                                                                                
                                                                                                                                                                                                                                
A log is the physical append-only file on disk where events for a partition are sequentially written and stored. Each partition has its own log that is divided into segments, which are replicated across nodes using the Raft consensus algorithm to ensure data durability and fault tolerance.

Interesting that it focuses on partitions rather than topics and partitions!

Feediver1 (approved):

PR #1608, opened 2026-03-16
feat(DOC-1970): add missing rpk connect reference pages

mfernest:
Fixes https://redpandadata.atlassian.net/browse/DOC-1970
## Summary

Adds documentation for three `rpk connect` commands missing from the reference:

- `rpk connect agent` (+ `agent init`, `agent run` subcommands) — marked EXPERIMENTAL
- `rpk connect dry-run`
- `rpk connect plugin` (+ `plugin init` subcommand) — marked EXPERIMENTAL

Updates `nav.adoc` to include the new pages in alphabetical order alongside existing `rpk connect` entries.

## Test plan

- [x] Build passes locally (exit 0)
- [ ] Netlify deploy preview renders new pages correctly
- [ ] New pages appear in nav under `rpk connect`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1616, opened 2026-03-18
review fixes: chroot flags, dry-run usage, plugin-init example

mmatczuk:
Review fixes for #1608:

- `rpk-connect-agent-run.adoc`: add Linux-only `--chroot` and `--chroot-passthrough` flags
- `rpk-connect-dry-run.adoc`: fix Usage line from `[command options]` to `[OPTIONS] [files...]`
- `rpk-connect-plugin-init.adoc`: add Example section

### DIFF
--- a/modules/reference/pages/rpk/rpk-connect/rpk-connect-plugin-init.adoc
+++ b/modules/reference/pages/rpk/rpk-connect/rpk-connect-plugin-init.adoc
@@ -12,6 +12,13 @@ Generate a project on the local filesystem that can be used as a starting point 
 
  rpk connect plugin init [OPTIONS]
 
+== Example
+
+[,bash]
+----
+rpk connect plugin init ./my-plugin
+----
+
 == Options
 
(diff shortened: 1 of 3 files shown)


modules/reference/pages/rpk/rpk-connect/rpk-connect-plugin-init.adoc
```diff
@@ -12,6 +12,13 @@ Generate a project on the local filesystem that can be used as a starting point
 
  rpk connect plugin init [OPTIONS]
 
+== Example
+
+[,bash]
+----
+rpk connect plugin init ./my-plugin
+----
```
mfernest, line 20: Per our [style guide](https://github.com/redpanda-data/docs-team-standards/blob/main/resources/writing-style/style-guide.md#placeholders), we avoid possessive prefixes like `my-` in examples. Would `./example-plugin` or similar work here instead?

PR #1665, opened 2026-04-10
ci: add playbook branch validation check

micheleRP:
## Summary

- Adds a CI workflow that checks `local-antora-playbook.yml` for non-standard branch references
- Fails the check if any `branches:` value doesn't match the allowed set (`main`, `HEAD`, `v/*`, `shared`, `site-search`, `!v-end-of-life/*`)
- Prevents accidentally merging a playbook that still points to a PR branch, which would break the production build

## Context

When single-sourcing pages across repos, we temporarily override playbook branch references to point to PR branches for Netlify previews. This CI check serves as a safety net to ensure those overrides are reverted before merge.

## Test plan

- [ ] Verify the check passes on main (all standard branches)
- [ ] Verify the check would fail if a PR branch name were in the playbook

### DIFF
--- a/local-antora-playbook.yml
+++ b/local-antora-playbook.yml
@@ -16,7 +16,7 @@ content:
   - url: .
     branches: HEAD
   - url: https://github.com/redpanda-data/docs
-    branches: [test, v/*, shared, site-search,'!v-end-of-life/*']
+    branches: [v/*, shared, site-search,'!v-end-of-life/*']
   - url: https://github.com/redpanda-data/cloud-docs
     branches: 'main'
   - url: https://github.com/redpanda-data/redpanda-labs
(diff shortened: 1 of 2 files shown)


JakeSCahill (approved):

local-antora-playbook.yml
```diff
@@ -16,7 +16,7 @@ content:
   - url: .
     branches: HEAD
   - url: https://github.com/redpanda-data/docs
-    branches: [v/*, shared, site-search,'!v-end-of-life/*']
+    branches: [test, v/*, shared, site-search,'!v-end-of-life/*']
```
JakeSCahill:
```suggestion
    branches: [v/*, shared, site-search,'!v-end-of-life/*']
```

PR #1611, opened 2026-03-17
fix(DOC-1959): correct election_timeout_ms type/default; exclude raft_election_timeout_ms

mfernest:
## Summary

Fixes two issues with the `election_timeout_ms` cluster property reference (escalated from Support, P2):

- **`election_timeout_ms`** was showing type `string` and default `null` — fixed via property-overrides.json to correctly show type `integer` and default `1500` (1500 milliseconds)
- **`raft_election_timeout_ms`** was appearing as a separate cluster property — this is the internal C++ variable name, not the user-facing config key; marked `exclude_from_docs` so it no longer appears in the reference

The correct user-facing property is `election_timeout_ms` with a default of 1500 milliseconds.

## Preview

[raft_election_timeout_ms](https://deploy-preview-1611--redpanda-docs-preview.netlify.app/current/reference/properties/cluster-properties/#raft_election_timeout_ms)

[election_timeout_ms](https://deploy-preview-1611--redpanda-docs-preview.netlify.app/current/reference/properties/cluster-properties/#election_timeout_ms)

## Test plan

- [x] Build passes locally
- [ ] Verify `election_timeout_ms` shows type `integer` and default `1500` at `/reference/properties/cluster-properties#election_timeout_ms`
- [ ] Verify `raft_election_timeout_ms` no longer appears on the cluster properties page

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1712, opened 2026-05-26
fix(DOC-1959): remove orphaned raft_election_timeout_ms section from cluster properties

Feediver1:
## Summary

Follow-up to #1611. Closes out DOC-1959.

PR #1611 added an `exclude_from_docs` override for `raft_election_timeout_ms` in `docs-data/property-overrides.json`. That override succeeded at the source-JSON layer (`modules/reference/attachments/redpanda-properties-v26.1.6.json` on `main` no longer contains the property), but the rendered `modules/reference/partials/properties/cluster-properties.adoc` partial was never regenerated. As a result, the published Cluster Properties page still showed a `raft_election_timeout_ms` entry, leaving DOC-1959 unsatisfied.

This PR deletes the orphaned 47-line section directly from the partial. Future auto-docs regeneration runs will preserve the deletion because the source v26.1.6 JSON no longer contains the property.

## Verify

- [ ] On the Netlify deploy preview, navigate to *Reference > Properties > Cluster Properties* and confirm `Ctrl+F raft_election_timeout_ms` returns no hits.
- [ ] Confirm `election_timeout_ms` (the correct user-facing property) is unchanged.

## Test plan

- [ ] Netlify build passes
- [ ] `grep -rn raft_election_timeout_ms modules/` returns zero hits


jason-da-redpanda (approved):

micheleRP (commented):
## Review

Final-pass review for #1712. Fix is correct and minimal — the deletion accomplishes the user-facing goal and breaks no xrefs. Two notes below on the PR description and a UX question.

### Suggestions

**1. PR description has two inaccurate technical claims**

- *"the override succeeded at the source-JSON layer (v26.1.6 JSON no longer contains the property)"* — the current source JSON on `main` is `redpanda-properties-v26.1.8.json` and it still contains `raft_election_timeout_ms` (line 10740). `exclude_from_docs` doesn't strip properties from the source JSON; it operates at the partial-generation layer.

- *"Future auto-docs regeneration runs will preserve the deletion because the source v26.1.6 JSON no longer contains the property."* — the next auto-docs run will re-add the section to the partial, but wrapped in `// tag::exclude-from-docs[]` / `// end::exclude-from-docs[]` markers (see `docs-extensions-and-macros/tools/property-extractor/templates/property.hbs`). The consumer page (`modules/reference/pages/properties/cluster-properties.adoc:13`) filters tagged content out at include time via `[tags=!deprecated;!exclude-from-docs]`, so the published page stays clean — but the partial file itself will show the section reappearing in the next auto-docs diff.

The current partial on `main` has the section *without* tag wrappers because auto-docs hasn't run since #1611 was merged (last auto-docs run was for v26.1.9 earlier that day, before #1611 landed). That's why the manual deletion was needed.

Suggested rewording for the description so the next reviewer doesn't think auto-docs has regressed:

> Future auto-docs regeneration runs will re-add the section, but wrapped in `// tag::exclude-from-docs[]` markers that the consumer page filters out via `[tags=!exclude-from-docs]`. The deletion is a stopgap until that regeneration runs.

**2. Stale-bookmark UX**

Commit fc1fcc48 added a soft-redirect anchor on the parent page, then 67f68bc4 reverted it with no reasoning in the commit message. Worth a one-line note on why the anchor was pulled — or consider re-adding it so users who bookmarked `cluster-properties#raft_election_timeout_ms` land on a brief note pointing to `election_timeout_ms` instead of the page top.

### Impact on other files

No updates required. Confirmed via `git ls-files | xargs grep -l raft_election_timeout_ms` — the only remaining references are the source JSON (correctly retained), `docs-data/property-overrides.json` (the override entry), and the partial itself. No `xref:` to the removed anchor exists in `docs`, `cloud-docs`, or `rp-connect-docs`. PR #1611 already corrected the user-facing `election_timeout_ms` entry, so no related-page updates are needed and no What's New entry is warranted (removing incorrect content, not announcing a change).

### What works well

- Minimal, targeted change (single file, 0 additions / 47 deletions).
- No code paths or behavior touched.
- Leaves the override in `property-overrides.json` so the long-term mechanism keeps working.
- Concrete Verify and Test plan checklists in the PR body.

micheleRP (approved):
Approving — no critical issues. Suggestions in the prior review comment are optional polish (PR description wording + revert reasoning); feel free to merge.

PR #1495, opened 2025-12-02
Upgrade Doc Detective and fix test timing issues

JakeSCahill:
## Description

This PR upgrades Doc Detective from v3.5.0 to v3.6.1, which resolves all timing and selector parsing issues discovered during testing.

This upgrade happened in stages as issues were discovered, reported, and fixed:

### Initial upgrade to v3.5.0

- Gained improved page load detection via network traffic monitoring
- Encountered critical bugs:
  - `detectSteps` config option was being ignored ([#159](https://github.com/doc-detective/doc-detective/issues/159))
  - CSS selectors with `data-testid` attributes were being parsed incorrectly

### Upgrade to v3.5.1

- Fixed the `detectSteps` bug ([#159](https://github.com/doc-detective/doc-detective/issues/159))
- Still had issues with `data-testid` selector parsing

### Upgrade to v3.6.0

- Fixed XML attribute parsing bug that was breaking `data-testid` selectors
- All selector issues resolved

### Final upgrade to v3.6.1

- **Fixed timing issues with element detection after navigation**
- Tests now pass reliably without workaround wait or screenshot steps
- Element timeout handling works as expected

### Test reliability improvements

All tests now pass consistently:

- Element detection works reliably after navigation
- No timing-related failures
- Cleaner, more maintainable test code

Related upstream issues we reported:

- [#165](https://github.com/doc-detective/doc-detective/issues/165) - Element timing issue
- [#164](https://github.com/doc-detective/doc-detective/issues/164) - Navigation reliability

## Testing

All tests pass successfully with v3.6.1:

```bash
npm run test-quickstart
```

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up
- [ ] Small fix (typos, links, copyedits, etc)

PR #1743, opened 2026-06-10
chore(netlify): force clean node_modules on every preview build

Feediver1:
## Summary

Adds a `netlify.toml` to the root of the docs repo that forces a clean `node_modules` install on every preview build. Closes the cache-poisoning class of failures that has been blocking PRs against `v/25.1` and `v/25.2`.

## Problem

Netlify's persistent build cache keeps a poisoned `node_modules` snapshot from early June 2026. With `npm 10.9.3` (the version Node 22.20.0 ships with), reconciling that cached state against the current `package-lock.json` fails with:

```
npm error code ERR_INVALID_ARG_TYPE
npm error The "from" argument must be of type string. Received undefined
```

The build crashes within ~5 seconds of `npm install`, before Antora loads. Symptoms observed across this session:

- **PR 1739** (v/24.3, DOC-2237 backport): npm install crash → manual cache clear via "Retry without cache" → next push restored the same snapshot → crash returned.
- **PR 1737, 1738** (v/25.2, v/25.1): same fingerprint, multiple cycles of Retry-without-cache required to get even one green build per push.

Clean `npm install` always works (verified by cloning fresh and running locally). The problem is exclusively the cached state Netlify restores.

## Fix

Override the build command in `netlify.toml`:

```toml
[build]
  command = "rm -rf node_modules && npm install --prefer-online --no-audit --no-fund && npm run build"
```

- `rm -rf node_modules` discards Netlify's restored cache.
- `npm install --prefer-online` forces npm to bypass its own metadata cache.
- `npm run build` then runs the existing Antora playbook unchanged.

Cost: ~30-60 seconds per build for the install (was effectively free when cache wasn't poisoned). Benefit: no more per-push manual cache clears.

## Scope

- Affects only the `redpanda-docs-preview` Netlify site (this repo's PR previews).
- The production `docs.redpanda.com` site is built from `redpanda-data/docs-site`, not from this repo, and is **not affected** by this change.
- Adding `netlify.toml` means PRs against `main` pick up the fix automatically once this lands. PRs against version branches won't get the fix until either (a) this commit is backported to those branches, or (b) each PR explicitly includes the file.

## Not included

- A post-build grep that fails the build on `ERROR (asciidoctor)` lines (mentioned earlier in DOC-2237 discussion). That's a separate consideration — there are currently a number of pre-existing cross-component errors from cloud-docs/main and rp-connect-docs/main that would need cleanup before strict-error mode can land. Worth doing as a follow-up after those are resolved.

## Verification

Once merged, the next preview build on any PR against `main` should:
1. Skip the cache restore (or restore it and immediately delete it).
2. Do a fresh `npm install`.
3. Build cleanly without `ERR_INVALID_ARG_TYPE`.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,27 @@
+# Netlify configuration for the redpanda-docs-preview site (PR previews of
+# redpanda-data/docs). The production site docs.redpanda.com is built from
+# redpanda-data/docs-site, not from this repo, so this file does not affect
+# production.
+#
+# Why this file exists
+# --------------------
+# Netlify preserves node_modules across builds via its persistent cache.
+# A poisoned cache snapshot taken in early June 2026 causes npm 10.9.3
+# to fail with ERR_INVALID_ARG_TYPE ("The 'from' argument must be of
+# type string. Received undefined") when reconciling the cached
+# node_modules against the current package-lock.json on certain
+# v/* base branches (notably v/25.1 and v/25.2). The error fires within
+# ~5 seconds of npm install starting, before Antora even loads.
+#
+# Manually clicking "Retry without cache" in the Netlify UI works for a
+# single deploy, but the next push to the same PR restores the same
+# poisoned snapshot and the build fails again. The cache is keyed at
+# the project / deploy-context level and survives across builds even
+# when individual deploys clear it.
+#
+# Forcing a clean node_modules every build (rm + fresh npm install)
+# bypasses the cached state entirely. This costs ~30-60s per build
+# but eliminates the cache poisoning problem for every PR.
+
+[build]
+  command = "rm -rf node_modules && npm install --prefer-online --no-audit --no-fund && npm run build"


kbatuigas (commented):
@Feediver1 @JakeSCahill — Before merging, I wanted to flag some context from the last time this happened:

**This is the second round of the same `ERR_INVALID_ARG_TYPE` crash.** Back on December 3, 2025, the identical error hit preview builds after Jake's Doc Detective upgrade (PR #1495). That time, Jake fixed it by updating the Node.js version in the Netlify UI settings for the preview site.

**I'm not sure a `netlify.toml` belongs in this repo.** This is a content source repo — the production site is built from `docs-site`, which owns the `netlify.toml` and already pins `NODE_VERSION = "22"` there. Build infrastructure config for the preview site has historically lived in the Netlify UI, not in this repo.

**Would the simpler fix be to update the Node version in the Netlify UI for `redpanda-docs-preview` again?** That's what resolved it last time, it keeps build config out of the content repo, and it avoids the 30-60s per-build cost of nuking `node_modules` on every run.

PR #1747, opened 2026-06-15
Devprod 4222 update setup urls v23.3

PrzemekZglinicki:
## Description
Update the domain used to download Redpanda setup scripts in the install documentation. The scripts configure the system package manager (APT/YUM) to pull Redpanda packages from the new Artifact Registry.

Resolves https://redpandadata.atlassian.net/browse/DEVPROD-4222
Review deadline: 22.06.2026

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up
- [x] Small fix (typos, links, copyedits, etc)

PR #1709, opened 2026-05-20
Update setup script URLs from Cloudsmith to linux.pkg.redpanda.com

PrzemekZglinicki:
Update the domain used to download Redpanda setup scripts in the install documentation. The scripts configure the system package manager (APT/YUM) to pull Redpanda packages from the new Artifact Registry.

## Checks

- [ ] New feature
- [ ] Content gap
- [ ] Support Follow-up
- [x] Small fix (typos, links, copyedits, etc)

PR #1755, opened 2026-06-16
Add Linux package repository migration guide (DEVPROD-4222)

JakeSCahill:
## Description

The setup-script URL PRs (#1709 and the per-version backports #1747–#1753) update the **fresh-install** snippets to `linux.pkg.redpanda.com`. They don't cover the customer **migration** described in TSB-2026-38 — re-pointing an *existing* apt/yum/zypper install, the new GPG signing key, the manual (non-piped) configuration path, the deprecation timeline, or SUSE/zypper. This PR adds that durable content.

### What's added

- **New page** `migrate:linux-package-repository.adoc` — _Migrate the Linux Package Repository_ (in the **Migrate** section, alongside Console v3 / Helm-to-Operator / Strimzi):
  - Check whether you're affected (`grep dl.redpanda.com`)
  - Re-point via the setup script **or** manual configuration (GPG key import + explicit repo definition) for Debian/Ubuntu and Fedora/RHEL/**SUSE**
  - Disable the legacy repo, verify (`apt-cache policy` / `yum info`), validate-before-cutover, and config-management guidance
- **New partial** `linux/repo-migration-note.adoc` — a single deprecation/deadline `IMPORTANT` note, included from the Linux install, FIPS install, Console (Linux) install, and rolling-upgrade pages so the callout isn't duplicated.
- **nav** — new page under the Migrate section.

### Verification

- All four new URLs return HTTP 200 and serve valid setup scripts.
- Manual repo definitions, key URLs, repo names, and the SUSE path were taken from the **live** `setup-redpanda.{deb,rpm}.sh` scripts (not just the TSB text), so they match what the scripts actually write.

### Notes for reviewers

- **Overlap with #1709:** that PR changes the setup-script URL line in `install-redpanda.adoc`; this PR only *adds* a note + include to that file (different lines), so the conflict should be trivial. Suggest landing #1709 first.
- **Dates** (dual-publish June 1 → legacy shutdown Sept 28, 2026) come from TSB-2026-38 — please confirm before publishing.
- **SUSE/zypper** is documented because the `setup-redpanda.rpm.sh` script supports it; confirm it's officially supported.

## Checks

- [ ] New feature
- [x] Content gap
- [ ] Support Follow-up


modules/migrate/pages/linux-package-repository.adoc
```diff
@@ -0,0 +1,169 @@
...
+----
+--
+Fedora/RHEL/SUSE::
++
+--
+Look for a Redpanda entry in `/etc/yum.repos.d/` (or `/etc/zypp/repos.d/` on SUSE):
+
+[,bash]
+----
+grep -r dl.redpanda.com /etc/yum.repos.d/
```
kbatuigas: From pr-review: 

The SUSE grep command only checks /etc/yum.repos.d/, but SUSE uses /etc/zypp/repos.d/. A SUSE user running only this command would see no match even if affected.
- Current: grep -r dl.redpanda.com /etc/yum.repos.d/
- Suggested: Add a second line or a note, e.g.: grep -r dl.redpanda.com /etc/yum.repos.d/ /etc/zypp/repos.d/ 2>/dev/null

kbatuigas (approved):
Looks good. Just one thing to flag from pr-review but not a blocker.

PR #1758, opened 2026-06-18
upgrade: document broker restarts during operator upgrades

david-yu:
## What

Adds a **Broker restarts during operator upgrades** section to [Upgrade the Redpanda Operator](https://docs.redpanda.com/streaming/current/upgrade/k-upgrade-operator/), plus a `NOTE` in the existing *Operator-only upgrades* section.

## Why

Operator-only upgrades restart every broker Pod even when the Redpanda version doesn't change — the operator re-renders the broker StatefulSet, and the configurator init + sidecar image references it injects into the broker Pod template track the operator version. Users reasonably expect an "operator-only" upgrade to be control-plane-only, so this catches them off guard during prod planning.

This came out of a support thread (USE2 prod, operator bump to v25.3.7 rolled every broker with no Redpanda version change) and was confirmed against the operator source + an end-to-end validation on Kubernetes 1.36 (upgrade from v26.1.4, both v1 `Cluster` and v2 `Redpanda` operators).

## What the section covers

- **Expectation:** plan for a rolling restart of every broker on each operator version bump.
- **Why:** the configurator/sidecar image tag tracks the operator version (`--configurator-tag` defaults to the operator version), so an upgrade changes the broker Pod template → rolling restart. Also notes restarts happen when a release changes other parts of the Pod template (args/containers/mounts).
- **It's safe:** maintenance mode + wait-for-healthy between brokers.
- **Planning:** one cluster at a time, window sized by broker count, test in non-prod first.
- **Advanced staging option:** pin the configurator/sidecar image via the chart's `additionalCmdFlags` so the control-plane upgrade lands *without* rolling brokers, then bump the pinned image in a separate window.

## For reviewers — please sanity-check the advanced section

The "Stage the broker restart separately" subsection documents pinning `--configurator-base-image`/`--configurator-tag`. Two things to confirm before merge:

1. **Support stance.** In the originating thread, the answer to "is there a way to upgrade without rolling brokers" was *"not currently … no supported mechanism."* The pin is real and works (validated), but if the operator team considers it unsupported, we may want to soften/remove the advanced subsection or label it explicitly as unsupported.
2. **Accuracy of caveats.** I've scoped it carefully: the pin only avoids the roll when a release's *sole* broker-Pod-template change is the image tag (true for many patch releases, including the v25.3.7 case), and it temporarily runs a control-plane/sidecar version mismatch (warned against running long-term). Please confirm these match the operator team's guidance on control-plane↔sidecar version skew.

Happy to adjust framing (e.g., keep only the "expected + plan" guidance and drop the staging technique) based on the operator team's call.

## Preview pages

- [Upgrade the Redpanda Operator](https://deploy-preview-1758--redpanda-docs-preview.netlify.app/streaming/current/upgrade/k-upgrade-operator/) (updated)

### DIFF
--- a/modules/upgrade/pages/k-upgrade-operator.adoc
+++ b/modules/upgrade/pages/k-upgrade-operator.adoc
@@ -9,6 +9,8 @@ For complete upgrade instructions, including how to upgrade both the Redpanda Op
 
 In some cases, you might need to upgrade only the Redpanda Operator without upgrading your Redpanda cluster. For example, you might need to apply a patch release that fixes an operator bug.
 
+NOTE: An operator-only upgrade still restarts your broker Pods, even though the Redpanda version does not change. The Redpanda Operator injects its own container images (the configurator init container and the sidecar) into each broker Pod, and these image references track the operator version. When you upgrade the operator, the broker Pod template changes and the brokers roll one at a time. For why this happens and how to plan or stage it, see <<broker-restarts,Broker restarts during operator upgrades>>.
+
 To upgrade only the operator:
 
 . https://github.com/redpanda-data/redpanda-operator/releases[Review the Redpanda Operator release notes^] and xref:upgrade:k-compatibility.adoc[the Kubernetes compatibility matrix].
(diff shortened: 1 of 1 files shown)


micheleRP (approved):
thank you @david-yu!

PR #1775, opened 2026-07-02
fix(properties): correct topic property name in kafka_batch_max_bytes reference

Feediver1:
## Why

The `kafka_batch_max_bytes` cluster property referenced the topic-level message-size property by the **wrong name** — `message.max.bytes` (the Kafka *broker*-config name) — in two places: its description and its **Related topics** link. The correct topic property is **`max.message.bytes`** (defined in `topic-properties.adoc`).

This is the self-managed counterpart to the same error fixed in cloud-docs prose (redpanda-data/cloud-docs#627).

## What

- **`docs-data/property-overrides.json`** — corrected both `message.max.bytes` → `max.message.bytes` in the `kafka_batch_max_bytes` override (`description` + `related_topics`). This is the source of truth.
- **`modules/reference/partials/properties/cluster-properties.adoc`** and **`modules/reference/attachments/redpanda-properties-v26.1.10.json`** — regenerated via `doc-tools generate property-docs --tag v26.1.10 --generate-partials` (cloud-support tags intact), so the generated artifacts reflect the override.

Diff is limited to this one property: 3 files, 6 lines, all the `message.max.bytes` → `max.message.bytes` correction.

## Note for reviewers

Regenerating at v26.1.10 surfaced one **unrelated** drift I deliberately excluded from this PR: `topic_memory_per_partition`'s default rendered as the raw constant `DEFAULT_TOPIC_MEMORY_PER_PARTITION` instead of the resolved `200 KiB (204800)` currently on `main`. I reverted that line to match `main` so this PR stays scoped. Worth a separate look — a future auto-docs regen may reintroduce it (looks like a default-resolution regression in the property extractor).


micheleRP (approved):
lgtm

PR #1491, opened 2025-11-25
DOC-1667 Document Shadow Link in Cloud

micheleRP:
## Description
This pull request single sources and conditionalizes Shadowing content for Cloud. 

It also restructures the configuration page, moving conceptual content into the overview & moving **Create shadow link** section toward the top of the config page.

(Single sourcing for this in `cloud-docs` repo is done in https://github.com/redpanda-data/cloud-docs/pull/462. That one will merge Dec 12, to make Shadowing visible in Cloud docs.)

* Added `// tag::single-source[]` and `// end::single-source[]` markers to support reuse and conditional rendering across cloud and self-hosted documentation. [[1]](diffhunk://#diff-4b8d4bcdaa909a433d1efbc5cf3ea6d613666bc3172d11863fa32e9908677d76R5-L12) [[2]](diffhunk://#diff-2caaee7a7fdd71e6bc315efcdee281499136240f8c84211e6ee8bc58042bf23cR6-L13) [[3]](diffhunk://#diff-088a4519e87e79d8409ccc68d47c11f470716be741575467fdd0ee22c0dac706R5-L12) [[4]](diffhunk://#diff-794c9a00985adf6bf3d0d06a0e6b716aadc1f3fbe4f6f2629474ac86682909d0R6-R13) [[5]](diffhunk://#diff-11f9a8ee158683c0fea693d9138ed1e0f4891e9e9c165538a081786b1f1c28d7R6-R32)
* Introduced `ifdef::env-cloud[]` and `ifndef::env-cloud[]` blocks for environment-specific notes, requirements, and tips—such as licensing, authentication, and cluster property configuration. [[1]](diffhunk://#diff-794c9a00985adf6bf3d0d06a0e6b716aadc1f3fbe4f6f2629474ac86682909d0R35-R41) [[2]](diffhunk://#diff-11f9a8ee158683c0fea693d9138ed1e0f4891e9e9c165538a081786b1f1c28d7R54-R64) [[3]](diffhunk://#diff-11f9a8ee158683c0fea693d9138ed1e0f4891e9e9c165538a081786b1f1c28d7R82-R88)
[...]

PR #1776, opened 2026-07-02
DOC-2254: Add Confluent Schema Registry schema migration via Shadowing

Feediver1:
Documents the new v26.2 Schema Registry API-mode replication (`shadow_schema_registry_api`) for shadow links:

- New page: Migrate Schemas from Confluent Schema Registry (intro, use cases, prerequisites, limitations, config steps, filtering, context mapping, validation policy, monitoring, failover)
- setup.adoc: present topic mode vs API mode in the Schema Registry synchronization section
- overview.adoc: add Schema Registry Sync task tab and update the replication lists
- monitor.adoc: add Schema Registry sync status bullet
- nav.adoc: add nav entry under Shadowing

## Description

Resolves https://redpandadata.atlassian.net/browse/DOC-2254

## Page previews

- https://deploy-preview-1776--redpanda-docs-preview.netlify.app/streaming/26.2/manage/disaster-recovery/shadowing/migrate-schemas-confluent/
- https://deploy-preview-1776--redpanda-docs-preview.netlify.app/streaming/26.2/manage/disaster-recovery/shadowing/setup/
- https://deploy-preview-1776--redpanda-docs-preview.netlify.app/streaming/26.2/manage/disaster-recovery/shadowing/overview/
- https://deploy-preview-1776--redpanda-docs-preview.netlify.app/streaming/26.2/manage/disaster-recovery/shadowing/monitor/

## Sources of truth

All YAML field names verified against `src/go/rpk/pkg/cli/shadow/types.go` (dev) and the `types_test.go` example config. Defaults (10s tail / 5m full sync / 30 rps, `FAIL` policy) from `src/v/cluster_link/model/types.h`. Merged engineering PRs: redpanda#30695, #30723, #30764, #30835, #30860, #30905, #30933.

## ⚠️ Verify before merge (unreleased feature, ENG-1171)

Two engineering PRs were still open at draft time. `// DRAFT-VERIFY (DOC-2254)` comments in the new page mark each item:

- [ ] HTTP source reader + destination bootstrap (redpanda#30999): confirm whether the destination Schema Registry must be empty
- [ ] Failover mechanism via `paused` (redpanda#30984)
- [ ] Exact list of unsupported Confluent schema features (SME: @pgellert)
- [ ] Wildcard support in `source_filter` (epic mentions wildcards; rpk source shows plain string lists)
- [ ] `rpk shadow create` argument order against a 26.2 build
- [ ] Re-check defaults at code-complete

## Related

- Cloud twin: DOC-2307 (deferred; new content is wrapped in `single-source` tags with `env-cloud` conditionals, ready for the cloud-docs stub)
  - Note for the DOC-2307 stub: the `:learning-objective-N:` attributes are defined outside the `single-source` tag, but the `{learning-objective-N}` references are inside it. The cloud-docs stub header must define the same three attributes or the cloud build will render the references literally.
- Companion glossary PR (shared branch): #1777 (merge before or with this PR). Overlaps DOC-1660 (in progress).

## Checks

- [x] New feature
- [ ] Content gap
- [ ] Support Follow-up

## Preview pages

- [Migrate Schemas from Confluent Schema Registry](https://deploy-preview-1776--redpanda-docs-preview.netlify.app/streaming/26.2/manage/disaster-recovery/shadowing/migrate-schemas-confluent/) (new)
- [Configure Shadowing](https://deploy-preview-1776--redpanda-docs-preview.netlify.app/streaming/26.2/manage/disaster-recovery/shadowing/setup/) (updated)
- [Shadowing Overview](https://deploy-preview-1776--redpanda-docs-preview.netlify.app/streaming/26.2/manage/disaster-recovery/shadowing/overview/) (updated)
- [Monitor Shadowing](https://deploy-preview-1776--redpanda-docs-preview.netlify.app/streaming/26.2/manage/disaster-recovery/shadowing/monitor/) (updated)

### DIFF
--- a/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
+++ b/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
@@ -20,7 +20,7 @@ include::shared:partial$enterprise-license.adoc[]
 ====
 endif::[]
 
-Replicate glossterm:schema[,schemas] from a Confluent Schema Registry into a Redpanda glossterm:shadow cluster[] automatically and continuously. Schema replication through a glossterm:shadow link[] keeps subjects, versions, and compatibility settings synchronized with the source registry, so your applications keep working after a cutover without a separate schema migration step.
+Redpanda allows you to replicate glossterm:schema[,schemas] from a Confluent Schema Registry into a Redpanda glossterm:shadow cluster[] automatically and continuously. Schema replication through a glossterm:shadow link[] keeps subjects, versions, and compatibility settings synchronized with the source registry, so your applications keep working after a cutover without a separate schema migration step.
 
 After reading this page, you will be able to:
 
--- a/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
+++ b/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
@@ -1,0 +1,272 @@
+= Migrate Schemas from Confluent Schema Registry
+:description: Replicate subjects, versions, and compatibility settings from a Confluent Schema Registry into a Redpanda shadow cluster.
+:env-linux: true
+:page-topic-type: how-to
+:page-categories: Management, High Availability, Disaster Recovery
+:learning-objective-1: Configure a shadow link that continuously replicates schemas from a Confluent Schema Registry
+:learning-objective-2: Filter replication by context or subject and map source contexts to destination contexts
+:learning-objective-3: Monitor schema replication status and resolve validation errors
+// tag::single-source[]
+
+// DRAFT-VERIFY (DOC-2254): This page documents an unreleased v26.2 feature (ENG-1171).
+// Engineering PRs redpanda#30999 (HTTP source reader) and redpanda#30984 (failover) were
+// still open when this draft was written (2026-07-02). Re-verify every DRAFT-VERIFY item
+// against merged code and a running 26.2 build before publishing. SME: @pgellert.
+
+ifndef::env-cloud[]
+[NOTE]
+====
+include::shared:partial$enterprise-license.adoc[]
+====
+endif::[]
+
+Replicate glossterm:schema[,schemas] from a Confluent Schema Registry into a Redpanda glossterm:shadow cluster[] automatically and continuously. Schema replication through a glossterm:shadow link[] keeps subjects, versions, and compatibility settings synchronized with the source registry, so your applications keep working after a cutover without a separate schema migration step.
+
+After reading this page, you will be able to:
+
+* [ ] {learning-objective-1}
+* [ ] {learning-objective-2}
+* [ ] {learning-objective-3}
+
+== How schema replication works
+
+When you configure a shadow link with the `shadow_schema_registry_api` option, the shadow cluster polls the source Schema Registry over HTTP and imports changes into its own glossterm:Schema Registry[]. Two sync cycles keep the registries in step:
+
+* *Tail syncs* run frequently (default: every 10 seconds) to pick up incremental changes.
+* *Full syncs* scan all selected subjects (default: every 5 minutes) to catch anything a tail sync missed.
+
+Replicated schemas keep their original glossterm:subject[] names and version IDs, so producers and consumers that reference schemas by ID continue to work after failover. Schemas that reference other schemas import in dependency order.
+
+Before importing a schema, Redpanda validates it against the Redpanda Schema Registry implementation. If a schema uses features that Redpanda does not support, the sync either reports an error and skips the schema, or removes the unsupported fields and imports the rest, depending on the <<choose-a-validation-policy,validation policy>> you choose.
+
+While the link is active, the destination contexts that the link replicates into are read-only: the shadow cluster rejects client writes to those contexts so that replicated schemas remain identical to the source. Contexts outside the link's filter remain writable.
+
+[NOTE]
+====
+This API-based mode is an alternative to the byte-for-byte `_schemas` topic replication described in xref:manage:disaster-recovery/shadowing/setup.adoc#schema-registry-synchronization[Configure Shadowing]. A shadow link uses one Schema Registry sync mode or the other, not both. Use the API-based mode when the source is a Confluent Schema Registry, or when you need to replicate only selected contexts or subjects.
+
+Schema replication does not add any cluster configuration properties. All settings live in the shadow link configuration.
+====
+
+== Use cases
+
+* *Migrate from Confluent to Redpanda*: Replicate schemas continuously while xref:manage:disaster-recovery/shadowing/setup.adoc[Shadowing] replicates your topic data, then cut applications over to Redpanda once both are in sync. No separate schema migration tooling is required.
+* *Phased migration*: Use context and subject filters to migrate one team, application, or environment at a time.
+* *Registry reorganization*: Map source contexts to different destination contexts to restructure your Schema Registry as part of the migration.
+
+== Prerequisites
+
+* A shadow cluster running Redpanda version 26.2 or later. The schema replication feature activates after all brokers complete the upgrade.
+* Network connectivity from the shadow cluster to the source Schema Registry HTTP endpoint.
+* Credentials for the source registry with permission to read subjects, versions, and configuration. For Confluent Cloud, use a Schema Registry API key and secret.
+* A shadow link to the source cluster, or a plan to create one. See xref:manage:disaster-recovery/shadowing/setup.adoc[Configure Shadowing].
+
+// DRAFT-VERIFY (DOC-2254): Confirm with @pgellert whether the destination Schema Registry
+// (_schemas topic) must be empty before enabling API-mode sync, or whether the destination
+// bootstrap in redpanda#30999 relaxes this. Topic mode requires it; API mode is unconfirmed.
+
+== Limitations
+
+* HTTP basic authentication is the only supported authentication method for the source registry.
+* A shadow link replicates Schema Registry data in one mode only: either `shadow_schema_registry_topic` or `shadow_schema_registry_api`.
+* Replication is one way, from the source registry to the shadow cluster. Destination contexts owned by the link are read-only until failover.
+* Schemas that use Confluent features not supported by the Redpanda Schema Registry are not replicated as-is. Choose a <<choose-a-validation-policy,validation policy>> to control whether these schemas are skipped or imported without the unsupported fields.
+
+// DRAFT-VERIFY (DOC-2254): Get the exact list of unsupported Confluent schema features
+// (rule sets? metadata tags?) from @pgellert and document them here.
+// DRAFT-VERIFY (DOC-2254): ENG-1171 mentions wildcard support in filters, but the rpk
+// source shows plain string lists for contexts/subjects. Confirm whether wildcards or
+// patterns are accepted before documenting either behavior.
+
+== Configure schema replication
+
+Add the `shadow_schema_registry_api` option to the `schema_registry_sync_options` section of your shadow link configuration file.
+
+=== Generate a configuration template
+
+Generate a configuration file template that includes all available fields with comments:
+
+[,bash]
+----
+rpk shadow config generate --print-template -o shadow-config-template.yaml
+----
+
+For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-config-generate.adoc[`rpk shadow config generate`].
+
+=== Connect to the source registry
+
+Configure the connection to the source Schema Registry:
+
+[,yaml]
+----
+schema_registry_sync_options:
+  shadow_schema_registry_api:
+    source_url: https://psrc-xxxxx.us-east-1.aws.confluent.cloud  # Source Schema Registry endpoint
+    auth_options:
+      basic:
+        username: <sr-api-key>                # Confluent Schema Registry API key
+        password: <sr-api-secret>             # Confluent Schema Registry API secret
+    tls_settings:
+      enabled: true                           # Use TLS for the connection
+      tls_file_settings:
+        ca_path: /path/to/ca.crt              # Optional: CA certificate for custom trust
+    tail_interval: 10s                        # How often to poll for incremental changes
+    full_sync_interval: 5m                    # How often to run a full scan
+    max_source_requests_per_second: 30        # Rate limit for requests to the source registry
+----
+
+The intervals and rate limit are optional. If you omit them, Redpanda uses the defaults shown above.
+
+// DRAFT-VERIFY (DOC-2254): Defaults (10s / 5m / 30 rps) are from
+// src/v/cluster_link/model/types.h on dev. Re-check at code-complete.
+
+=== Select contexts and subjects
+
+By default, the link replicates the entire source registry. To replicate a subset, add a `source_filter` with the contexts or subjects to include:
+
+[,yaml]
+----
+schema_registry_sync_options:
+  shadow_schema_registry_api:
+    # ...connection settings...
+    source_filter:
+      contexts:
+      - "."                                   # The default context
+      - ".prod"                               # A named context
+      subjects:
+      - orders-value                          # A subject in the default context
+----
+
+Schema Registry contexts provide independent namespaces for subjects within one registry. The default context is named `.`. For more information, see xref:manage:schema-reg/schema-reg-contexts.adoc[Schema Registry Contexts].
+
+=== Map source contexts to destination contexts
+
+Choose how replicated contexts are named on the shadow cluster:
+
+* `identity`: Keep the source context names (default behavior for migrations).
+* `exact`: Map each source context to a different destination context.
+
+[,yaml]
+----
+schema_registry_sync_options:
+  shadow_schema_registry_api:
+    # ...connection settings and filters...
+    destination:
+      identity: {}                            # Keep source context names
+----
+
+To rename contexts during replication:
+
+[,yaml]
+----
+schema_registry_sync_options:
+  shadow_schema_registry_api:
+    # ...connection settings and filters...
+    destination:
+      exact:
+        mappings:
+        - source: "."                         # Source context
+          destination: ".shadow"              # Destination context on the shadow cluster
+----
+
+[[choose-a-validation-policy]]
+=== Choose a validation policy
+
+The `unsupported_schema_feature_policy` setting controls what happens when a source schema uses features that the Redpanda Schema Registry does not support:
+
+[cols="1,3"]
+|===
+| Policy | Behavior
+
+| `FAIL` (default)
+| The schema is not replicated. The sync records an error, reports it in the link status, and continues with the remaining schemas.
+
+| `REMOVE`
+| The unsupported fields are removed and the rest of the schema is imported. The sync counts each modification in the link status.
+|===
+
+[,yaml]
+----
+schema_registry_sync_options:
+  shadow_schema_registry_api:
+    # ...connection settings, filters, and destination...
+    unsupported_schema_feature_policy: FAIL
+----
+
+=== Create the shadow link
+
+Create the shadow link with your completed configuration file:
+
+[,bash]
+----
+rpk shadow create <link-name> --config-file shadow-config.yaml
+----
+
+For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-create.adoc[`rpk shadow create`].
+
+To change the schema replication settings on an existing link, see xref:reference:rpk/rpk-shadow/rpk-shadow-update.adoc[`rpk shadow update`].
+
+// DRAFT-VERIFY (DOC-2254): Confirm the exact `rpk shadow create` argument order
+// (link name vs --config-file only) against the 26.2 rpk build.
+
+== Verify the configuration
+
+Confirm that the link is configured for API-based schema replication:
+
+[,bash]
+----
+rpk shadow describe <link-name>
+----
+
+The output includes the shadowing mode, source URL, sync intervals, validation policy, and your context and subject filters. For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-describe.adoc[`rpk shadow describe`].
+
+== Monitor replication status
+
+Check schema replication progress and errors for a link:
+
+[,bash]
+----
+rpk shadow status <link-name>
+----
+
+The Schema Registry section of the output reports:
+
+[cols="1,3"]
+|===
+| Field | Description
+
+| Inventory
+| The number of selected subjects and subject versions on the source, compared with the number of subjects and versions on the destination. The registries are synchronized when the destination counts match the selected source counts. A destination that is behind the source indicates that replication is still in progress.
+
+| Current sync
+| The type of the sync in progress (`FULL` or `TAIL`) and the number of subject versions, compatibility configurations, and modes it has changed, including how many unsupported features were removed and how many errors occurred.
+
+| Last full sync
+| Start time, finish time, and change counts for the most recent completed full sync.
+
+| Totals since task start
+| Cumulative change and error counts since the schema replication task started.
+
+| Last error
+| The most recent replication error. With the `FAIL` validation policy, schemas that fail validation appear here.
+|===
+
+For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-status.adoc[`rpk shadow status`]. For general link monitoring, see xref:manage:disaster-recovery/shadowing/monitor.adoc[Monitor Shadowing].
+
+== Fail over
+
+When you fail over a shadow link, schema replication from the source registry stops and the write-blocked destination contexts become writable, so your applications can register new schemas on the promoted cluster.
+
+// DRAFT-VERIFY (DOC-2254): Failover behavior (single `paused` field pauses the sync task
+// and lifts per-context write blocks) is from open PR redpanda#30984. Confirm the final
+// mechanism and whether any manual steps apply to schema replication during failover.
+
+For the complete failover procedure, see xref:manage:disaster-recovery/shadowing/failover.adoc[Failover].
+
+== Next steps
+
+* xref:manage:disaster-recovery/shadowing/setup.adoc[Configure Shadowing] to replicate topic data, consumer offsets, and ACLs alongside your schemas.
+* xref:manage:disaster-recovery/shadowing/monitor.adoc[Monitor Shadowing]
+* xref:manage:schema-reg/schema-reg-contexts.adoc[Schema Registry Contexts]
+
+// end::single-source[]
--- a/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
+++ b/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
@@ -43,7 +43,10 @@ While the link is active, the destination contexts that the link replicates into
 
 [NOTE]
 ====
-This API-based mode is an alternative to the byte-for-byte `_schemas` topic replication described in xref:manage:disaster-recovery/shadowing/setup.adoc#schema-registry-synchronization[Configure Shadowing]. A shadow link uses one Schema Registry sync mode or the other, not both. Use the API-based mode when the source is a Confluent Schema Registry, or when you need to replicate only selected contexts or subjects.
+This API-based mode is an alternative to the byte-for-byte `_schemas` topic replication described in xref:manage:disaster-recovery/shadowing/setup.adoc#schema-registry-synchronization[Configure Shadowing]. A shadow link uses one Schema Registry sync mode or the other, not both:
+
+* Use *topic mode* (`shadow_schema_registry_topic`) when the source is another Redpanda cluster and you want an exact, complete replica of its Schema Registry. Topic mode shadows the `_schemas` topic byte for byte, so it does not filter, remap, or validate schemas.
+* Use *API mode* (`shadow_schema_registry_api`) when the source is a Confluent Schema Registry, or when you need to replicate only selected contexts or subjects, map source contexts to different destination contexts, or control how schemas that use unsupported features are handled with a validation policy.
 
 Schema replication does not add any cluster configuration properties. All settings live in the shadow link configuration.
 ====
--- a/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
+++ b/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
@@ -82,9 +82,11 @@ Schema replication does not add any cluster configuration properties. All settin
 
 Add the `shadow_schema_registry_api` option to the `schema_registry_sync_options` section of your shadow link configuration file.
 
-The following sample configuration file creates a shadow link that shadows all source topics and replicates schemas from a Confluent Schema Registry. The sections that follow explain each schema replication setting.
+The following sample configuration file creates a shadow link that shadows all source topics and replicates schemas from a Confluent Schema Registry. The highlighted lines show the schema replication settings, which the sections that follow explain.
 
-[,yaml]
+// NOTE: keep the lines= range pointing at the schema_registry_sync_options block
+// if you edit this sample.
+[,yaml,lines=20-38]
 ----
 # Sample shadow link configuration with API-mode Schema Registry replication
 
--- a/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
+++ b/modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
@@ -20,7 +20,7 @@ include::shared:partial$enterprise-license.adoc[]
 ====
 endif::[]
 
-Redpanda allows you to replicate glossterm:schema[,schemas] from a Confluent Schema Registry into a Redpanda glossterm:shadow cluster[] automatically and continuously. Schema replication through a glossterm:shadow link[] keeps subjects, versions, and compatibility settings synchronized with the source registry, so your applications keep working after a cutover without a separate schema migration step.
+When you migrate to Redpanda from a deployment that uses a Confluent Schema Registry, your producers and consumers depend on the glossterm:schema[,schemas] stored in that registry. Shadowing removes this migration obstacle: a glossterm:shadow link[] continuously replicates schemas from the source Confluent Schema Registry into the glossterm:Schema Registry[] built into the Redpanda glossterm:shadow cluster[], preserving glossterm:subject[] names, versions, and compatibility settings. Because both registries stay synchronized until cutover, your applications keep working on Redpanda without a separate schema migration step. Use this approach when you migrate from Confluent to Redpanda, or when you maintain a Redpanda disaster recovery cluster for a system that keeps its schemas in a Confluent Schema Registry.
 
 After reading this page, you will be able to:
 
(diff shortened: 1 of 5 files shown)


modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,272 @@
...
+// against merged code and a running 26.2 build before publishing. SME: @pgellert.
+
+ifndef::env-cloud[]
+[NOTE]
+====
+include::shared:partial$enterprise-license.adoc[]
+====
+endif::[]
+
+Replicate glossterm:schema[,schemas] from a Confluent Schema Registry into a Redpanda glossterm:shadow cluster[] automatically and continuously. Schema replication through a glossterm:shadow link[] keeps subjects, versions, and compatibility settings synchronized with the source registry, so your applications keep working after a cutover without a separate schema migration step.
```
Feediver1:
```suggestion
Redpanda allows you to replicate glossterm:schema[,schemas] from a Confluent Schema Registry into a Redpanda glossterm:shadow cluster[] automatically and continuously. Schema replication through a glossterm:shadow link[] keeps subjects, versions, and compatibility settings synchronized with the source registry, so your applications keep working after a cutover without a separate schema migration step.
```

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,320 @@
...
+This API-based mode is an alternative to the byte-for-byte `_schemas` topic replication described in xref:manage:disaster-recovery/shadowing/setup.adoc#schema-registry-synchronization[Configure Shadowing]. A shadow link uses one Schema Registry sync mode or the other, not both. Use the API-based mode when the source is a Confluent Schema Registry, or when you need to replicate only selected contexts or subjects.
+
+Schema replication does not add any cluster configuration properties. All settings live in the shadow link configuration.
+====
+
+== Use cases
+
+* *Migrate from Confluent to Redpanda*: Replicate schemas continuously while xref:manage:disaster-recovery/shadowing/setup.adoc[Shadowing] replicates your topic data, then cut applications over to Redpanda once both are in sync. No separate schema migration tooling is required.
+* *Phased migration*: Use context and subject filters to migrate one team, application, or environment at a time.
+* *Registry reorganization*: Map source contexts to different destination contexts to restructure your Schema Registry as part of the migration.
```
bartoszpiekny-redpanda: @pgellert do we want to also mention here about disaster recovery? 🤔

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+Before importing a schema, Redpanda validates it against the Redpanda Schema Registry implementation. If a schema uses features that Redpanda does not support, the sync either reports an error and skips the schema, or removes the unsupported fields and imports the rest, depending on the <<choose-a-validation-policy,validation policy>> you choose.
+
+While the link is active, the destination contexts that the link replicates into are read-only: the shadow cluster rejects client writes to those contexts so that replicated schemas remain identical to the source. Contexts outside the link's filter remain writable.
+
+[NOTE]
+====
+This API-based mode is an alternative to the byte-for-byte `_schemas` topic replication described in xref:manage:disaster-recovery/shadowing/setup.adoc#schema-registry-synchronization[Configure Shadowing]. A shadow link uses one Schema Registry sync mode or the other, not both. Use the API-based mode when the source is a Confluent Schema Registry, or when you need to replicate only selected contexts or subjects.
```
bartoszpiekny-redpanda: I'm thinking if it would be worth to state when to use one instead of the other 🤔

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+client_options:
+  bootstrap_servers:                         # Source Kafka cluster brokers
+  - <source-broker-1>:<port>                 # Example: "pkc-xxxxx.us-east-1.aws.confluent.cloud:9092"
+  - <source-broker-2>:<port>
+  # For the TLS and authentication settings that the shadow cluster uses to
+  # connect to the source Kafka cluster, see the complete configuration file
+  # reference in Configure Shadowing.
+
+topic_metadata_sync_options:
```
bartoszpiekny-redpanda: I'm not sure we need to provide `topic_metadata_sync_options` in context of schema registry shadowing 🤔

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+
+[NOTE]
+====
+This API-based mode is an alternative to the byte-for-byte `_schemas` topic replication described in xref:manage:disaster-recovery/shadowing/setup.adoc#schema-registry-synchronization[Configure Shadowing]. A shadow link uses one Schema Registry sync mode or the other, not both:
+
+* Use *topic mode* (`shadow_schema_registry_topic`) when the source is another Redpanda cluster and you want an exact, complete replica of its Schema Registry. Topic mode shadows the `_schemas` topic byte for byte, so it does not filter, remap, or validate schemas.
+* Use *API mode* (`shadow_schema_registry_api`) when the source is a Confluent Schema Registry, or when you need to replicate only selected contexts or subjects, map source contexts to different destination contexts, or control how schemas that use unsupported features are handled with a validation policy.
+
+Schema replication does not add any cluster configuration properties. All settings live in the shadow link configuration.
```
pgellert: This is false as stated, because we do add 2 new cluster configurations: https://github.com/redpanda-data/redpanda/blob/c153614a6972b0ee9e109ae93dc9d0a9369a1d37/src/v/config/configuration.cc#L3981-L3996

But they are escape hatches / tunables that won't commonly need to be adjusted.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+* Credentials for the source registry with permission to read subjects, versions, and configuration. For Confluent Cloud, use a Schema Registry API key and secret.
+* A shadow link to the source cluster, or a plan to create one. See xref:manage:disaster-recovery/shadowing/setup.adoc[Configure Shadowing].
+
+// DRAFT-VERIFY (DOC-2254): Confirm with @pgellert whether the destination Schema Registry
+// (_schemas topic) must be empty before enabling API-mode sync, or whether the destination
+// bootstrap in redpanda#30999 relaxes this. Topic mode requires it; API mode is unconfirmed.
+
+== Limitations
+
+* HTTP basic authentication is the only supported authentication method for the source registry.
```
pgellert: mTLS is also supported as an authentication mechanism, in addition to HTTP Basic

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+* A shadow link replicates Schema Registry data in one mode only: either `shadow_schema_registry_topic` or `shadow_schema_registry_api`.
+* Replication is one way, from the source registry to the shadow cluster. Destination contexts owned by the link are read-only until failover.
+* Schemas that use Confluent features not supported by the Redpanda Schema Registry are not replicated as-is. Choose a <<choose-a-validation-policy,validation policy>> to control whether these schemas are skipped or imported without the unsupported fields.
+
+// DRAFT-VERIFY (DOC-2254): Get the exact list of unsupported Confluent schema features
+// (rule sets? metadata tags?) from @pgellert and document them here.
```
pgellert: In schema definitions:
- rule sets
- metadata tags
In configs:
- override metadata
- override rule set
- default metadata
- default rule set
- compatibility group (note: not the same as compatibility level)

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+// DRAFT-VERIFY (DOC-2254): ENG-1171 mentions wildcard support in filters, but the rpk
+// source shows plain string lists for contexts/subjects. Confirm whether wildcards or
+// patterns are accepted before documenting either behavior.
```
pgellert: Only plain string lists for contexts and subjects is supported

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
pgellert: Yes, we can remove the `topic_metadata_sync_options` section here, it's not needed.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+
+| Current sync
+| The type of the sync in progress (`FULL` or `TAIL`) and the number of subject versions, compatibility configurations, and modes it has changed, including how many unsupported features were removed and how many errors occurred.
+
+| Last full sync
+| Start time, finish time, and change counts for the most recent completed full sync.
+
+| Totals since task start
+| Cumulative change and error counts since the schema replication task started.
+
```
pgellert: I am not sure where exactly to place this information on this page, but I think we should document somewhere that:
- These counters are ephemeral / not persisted, so it is expected that they reset to 0 on broker restarts or on partition movement of the `_schemas/0` topic partition.
- That the task is always running on the broker & shard that is the leader of the `_schemas` topic's partition.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+== Fail over
+
+When you fail over a shadow link, schema replication from the source registry stops and the write-blocked destination contexts become writable, so your applications can register new schemas on the promoted cluster.
+
+// DRAFT-VERIFY (DOC-2254): Failover behavior (single `paused` field pauses the sync task
+// and lifts per-context write blocks) is from open PR redpanda#30984. Confirm the final
+// mechanism and whether any manual steps apply to schema replication during failover.
+
+For the complete failover procedure, see xref:manage:disaster-recovery/shadowing/failover.adoc[Failover].
```
pgellert: On failover, it is necessary to also pause the schema registry syncing task (through setting the paused field in the config to true) to stop any further syncing and to unblock the schema registry for writes.

As the draft verify note says, the relevant implementation is in https://github.com/redpanda-data/redpanda/pull/30984.

modules/manage/pages/disaster-recovery/shadowing/overview.adoc
```diff
@@ -109,6 +109,20 @@ The task is controlled by the `security_sync_options` section in the configurati
...
+--
+[#schema-registry-sync-task]
+The **Schema Registry Sync task** replicates Schema Registry content so that applications that depend on schemas keep working after failover.
+
+The task is controlled by the `schema_registry_sync_options` section in the configuration file. It supports two modes:
+
+* **Topic mode** (`shadow_schema_registry_topic`): Shadows the `_schemas` system topic for byte-for-byte replication from another Redpanda cluster.
+* **API mode** (`shadow_schema_registry_api`): Polls the source Schema Registry over HTTP and imports selected contexts and subjects, with validation. Use this mode to replicate schemas from a Confluent Schema Registry. See xref:manage:disaster-recovery/shadowing/migrate-schemas-confluent.adoc[Migrate Schemas from Confluent Schema Registry].
+
+A shadow link uses one mode or the other, not both.
```
pgellert: One nuance here is that only API mode shadowing actually creates a explicit "Schema Registry Sync Task" that shows up in `rpk shadow status`. Topic mode replication just adds the `_schemas` topic to the list of topic partitions to shadow, it doesn't create a separate "Schema Registry Sync Task".

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+== Prerequisites
+
+* A shadow cluster running Redpanda version 26.2 or later. The schema replication feature activates after all brokers complete the upgrade.
+* Network connectivity from the shadow cluster to the source Schema Registry HTTP endpoint.
+* Credentials for the source registry with permission to read subjects, versions, and configuration. For Confluent Cloud, use a Schema Registry API key and secret.
+* A shadow link to the source cluster, or a plan to create one. See xref:manage:disaster-recovery/shadowing/setup.adoc[Configure Shadowing].
+
+// DRAFT-VERIFY (DOC-2254): Confirm with @pgellert whether the destination Schema Registry
+// (_schemas topic) must be empty before enabling API-mode sync, or whether the destination
+// bootstrap in redpanda#30999 relaxes this. Topic mode requires it; API mode is unconfirmed.
```
pgellert: In API mode, the schemas topic doesn't have to be empty, but the target contexts the link would import (scoped to the mapped contexts based on the `source_filter` and `destination` fields) need to be empty.

Ref: https://github.com/redpanda-data/redpanda/pull/31011

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+----
+rpk shadow create <link-name> --config-file shadow-config.yaml
+----
+
+For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-create.adoc[`rpk shadow create`].
+
+To change the schema replication settings on an existing link, see xref:reference:rpk/rpk-shadow/rpk-shadow-update.adoc[`rpk shadow update`].
+
+// DRAFT-VERIFY (DOC-2254): Confirm the exact `rpk shadow create` argument order
+// (link name vs --config-file only) against the 26.2 rpk build.
```
pgellert: Create doesn't take a link-name parameter:
```suggestion
[,bash]
----
rpk shadow create --config-file shadow-config.yaml
----

For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-create.adoc[`rpk shadow create`].

To change the schema replication settings on an existing link, see xref:reference:rpk/rpk-shadow/rpk-shadow-update.adoc[`rpk shadow update`].
```

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+----
+
+Schema Registry contexts provide independent namespaces for subjects within one registry. The default context is named `.`. For more information, see xref:manage:schema-reg/schema-reg-contexts.adoc[Schema Registry Contexts].
+
+=== Map source contexts to destination contexts
+
+Choose how replicated contexts are named on the shadow cluster:
+
+* `identity`: Keep the source context names (default behavior for migrations).
+* `exact`: Map each source context to a different destination context.
```
pgellert: It's important to call out here that `exact` mapping needs to be exhaustive, and if the source schema registry has any contexts that are not mapped, that will lead to a failure of the schema syncing task.
We should recommend that people consider setting up a `source_scope` filter for the mapped source contexts if they expect that other contexts may later be created on the source schema registry that would fail the sync at a future inconvenient time.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+* A shadow cluster running Redpanda version 26.2 or later. The schema replication feature activates after all brokers complete the upgrade.
+* Network connectivity from the shadow cluster to the source Schema Registry HTTP endpoint.
+* Credentials for the source registry with permission to read subjects, versions, and configuration. For Confluent Cloud, use a Schema Registry API key and secret.
+* A shadow link to the source cluster, or a plan to create one. See xref:manage:disaster-recovery/shadowing/setup.adoc[Configure Shadowing].
+
+// DRAFT-VERIFY (DOC-2254): Confirm with @pgellert whether the destination Schema Registry
+// (_schemas topic) must be empty before enabling API-mode sync, or whether the destination
+// bootstrap in redpanda#30999 relaxes this. Topic mode requires it; API mode is unconfirmed.
+
+== Limitations
```
pgellert: Another limitation to add is that there is currently no coordination between topic data replication and the schema registry syncing task, so topic data serialized with Confluent's SerDes format may contain schema IDs on the shadow cluster earlier than those schemas get replicated over.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,323 @@
...
+Schema replication does not add any cluster configuration properties. All settings live in the shadow link configuration.
+====
+
+== Use cases
+
+* *Migrate from Confluent to Redpanda*: Replicate schemas continuously while xref:manage:disaster-recovery/shadowing/setup.adoc[Shadowing] replicates your topic data, then cut applications over to Redpanda once both are in sync. No separate schema migration tooling is required.
+* *Phased migration*: Use context and subject filters to migrate one team, application, or environment at a time.
+* *Registry reorganization*: Map source contexts to different destination contexts to restructure your Schema Registry as part of the migration.
+
+== Prerequisites
```
pgellert: One more to add here: if schema registry contexts are being used, customers must have the [schema_registry_enable_qualified_subjects](https://docs.redpanda.com/streaming/current/reference/properties/cluster-properties/#schema_registry_enable_qualified_subjects) config enabled -- https://docs.redpanda.com/streaming/current/manage/schema-reg/schema-reg-contexts/#prerequisites

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Fixed in be454986 — replaced the claim with the two tunables (`schema_registry_sync_memory_bytes`, `schema_registry_sync_parallelism`) and noted the defaults suit most deployments.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Fixed in be454986 — the limitation now lists HTTP basic and mTLS, and the connection section explains mTLS means a client certificate and key in `tls_settings` instead of `auth_options`.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Documented in be454986 — the validation policy section now lists these (schema definitions: rule sets, metadata tags; configs: override/default metadata and rule sets, compatibility groups), with your note that compatibility groups are not compatibility levels.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Documented in be454986 — 'The `contexts` and `subjects` lists accept literal names only. Wildcard and prefix patterns are not supported.' DRAFT-VERIFY removed.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Removed in be454986 — the sample now goes straight from `client_options` to `schema_registry_sync_options`, and the highlight range is updated to match.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Added in be454986, at the end of the Monitor replication status section: the task runs on the broker/shard leading the `_schemas` partition, and the counters are not persisted — expect resets on broker restart or partition leadership movement.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Rewritten in be454986 — the Fail over section now states that failover alone does not stop schema replication, and that setting `paused: true` in `schema_registry_sync_options` is the step that stops syncing and unblocks the destination contexts for writes. DRAFT-VERIFY removed.

modules/manage/pages/disaster-recovery/shadowing/overview.adoc
Feediver1: Added in be454986 — the tab now notes that only API mode runs as a separate task in the shadow link status, and topic mode is monitored like any other shadow topic.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Fixed in be454986 — the DRAFT-VERIFY is replaced with a prerequisite: the destination contexts the link replicates into (per `source_filter` and `destination` mapping) must be empty, and the rest of the shadow cluster's registry is unaffected.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Added in be454986 as an IMPORTANT under the `exact` mapping example: mappings must cover every replicated context, an unmapped context fails the task, and we recommend scoping the `source_filter` contexts list to the mapped contexts to guard against contexts created later.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Added in be454986 as a limitation: topic data replication and schema replication are not coordinated, so records in the Confluent SerDes wire format can arrive before the schema IDs they contain have replicated.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Added in be454986 as a prerequisite, with xrefs to the `schema_registry_enable_qualified_subjects` property reference and the Schema Registry Contexts prerequisites. Noted it's enabled by default.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,310 @@
...
+
+When you migrate to Redpanda from a deployment that uses a Confluent Schema Registry, your producers and consumers depend on the glossterm:schema[,schemas] stored in that registry. Shadowing removes this migration obstacle: a glossterm:shadow link[] continuously replicates schemas from the source Confluent Schema Registry into the glossterm:Schema Registry[] built into the Redpanda glossterm:shadow cluster[], preserving glossterm:subject[] names, versions, and compatibility settings. Because both registries stay synchronized until cutover, your applications keep working on Redpanda without a separate schema migration step. Use this approach when you migrate from Confluent to Redpanda, or when you maintain a Redpanda disaster recovery cluster for a system that keeps its schemas in a Confluent Schema Registry.
+
+After reading this page, you will be able to:
+
+* [ ] {learning-objective-1}
+* [ ] {learning-objective-2}
+* [ ] {learning-objective-3}
+
+== How schema replication works
```
trevpanda:
```suggestion
== How HTTP API schema replication works
```
Need to distinguish between Confluent compatible SR replication and Redpanda

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,310 @@
...
+
+== Use cases
+
+* *Migrate from Confluent to Redpanda*: Replicate schemas continuously while xref:manage:disaster-recovery/shadowing/setup.adoc[Shadowing] replicates your topic data, then cut applications over to Redpanda once both are in sync. No separate schema migration tooling is required.
+* *Phased migration*: Use context and subject filters to migrate one team, application, or environment at a time.
+* *Registry reorganization*: Map source contexts to different destination contexts to restructure your Schema Registry as part of the migration.
+
+== Prerequisites
+
+* A shadow cluster running Redpanda version 26.2 or later. The schema replication feature activates after all brokers complete the upgrade.
```
trevpanda:
```suggestion
* A cluster running Redpanda version 26.2 or later. The schema replication feature activates after all brokers complete the upgrade.
```
It's just a cluster, doesn't need to be specified as a shadow cluster

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
trevpanda: @pgellert what's the impact of this? Users should not attempt to read from topics that rely on SR until SR synced? Or will this cause write or replication errors?

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,310 @@
...
+    # ...connection settings...
+    source_filter:
+      contexts:
+      - "."                                   # The default context
+      - ".prod"                               # A named context
+      subjects:
+      - orders-value                          # A subject in the default context
+----
+
+The `contexts` and `subjects` lists accept literal names only. Wildcard and prefix patterns are not supported.
```
trevpanda: @pgellert how do the source filters on subjects work in conjunction with context filters? Does the example above really only apply "orders-value" subject in the default context? What about the .prod context?

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,310 @@
...
+
+[cols="1,3"]
+|===
+| Policy | Behavior
+
+| `FAIL` (default)
+| The schema is not replicated. The sync records an error, reports it in the link status, and continues with the remaining schemas.
+
+| `REMOVE`
+| The unsupported fields are removed and the rest of the schema is imported. The sync counts each modification in the link status.
```
trevpanda:
```suggestion
| The unsupported fields are removed and the rest of the schema is imported. The sync records each modification in the link status.
```

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,310 @@
...
+* *Migrate from Confluent to Redpanda*: Replicate schemas continuously while xref:manage:disaster-recovery/shadowing/setup.adoc[Shadowing] replicates your topic data, then cut applications over to Redpanda once both are in sync. No separate schema migration tooling is required.
+* *Phased migration*: Use context and subject filters to migrate one team, application, or environment at a time.
+* *Registry reorganization*: Map source contexts to different destination contexts to restructure your Schema Registry as part of the migration.
+
+== Prerequisites
+
+* A shadow cluster running Redpanda version 26.2 or later. The schema replication feature activates after all brokers complete the upgrade.
+* Network connectivity from the shadow cluster to the source Schema Registry HTTP endpoint.
+* Credentials for the source registry with permission to read subjects, versions, and configuration. For Confluent Cloud, use a Schema Registry API key and secret.
+* A shadow link to the source cluster, or a plan to create one. See xref:manage:disaster-recovery/shadowing/setup.adoc[Configure Shadowing].
```
trevpanda:
```suggestion
* Basic Shadow link settings. See xref:manage:disaster-recovery/shadowing/setup.adoc[Configure Shadowing].
```

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,310 @@
...
+| The most recent replication error. With the `FAIL` validation policy, schemas that fail validation appear here.
+|===
+
+The schema replication task runs on the broker and shard that hosts the leader of the `_schemas` topic's partition. The counters in the status output are not persisted: expect them to reset to zero when that broker restarts or when leadership of the `_schemas` partition moves to another broker.
+
+For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-status.adoc[`rpk shadow status`]. For general link monitoring, see xref:manage:disaster-recovery/shadowing/monitor.adoc[Monitor Shadowing].
+
+== Fail over
+
+Failing over shadow topics does not stop schema replication. As part of failover, pause the schema replication task by setting `paused: true` in the `schema_registry_sync_options` section of the link configuration. Pausing the task stops further syncing from the source registry and makes the write-blocked destination contexts writable, so your applications can register new schemas on the promoted cluster.
```
trevpanda:
```suggestion
For best results, shadow links should be failed over as a single unit, which prevents partial failover scenarios and unexpected results. Selective failover, such as topics only, does not stop schema replication. As part of failover, pause the schema replication task by setting `paused: true` in the `schema_registry_sync_options` section of the link configuration. Pausing the task stops further syncing from the source registry and makes the write-blocked destination contexts writable, so your applications can register new schemas on the promoted cluster.
```

trevpanda (changes requested):
This looks great, I left a few suggestions.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
pgellert: It's explained here: https://github.com/redpanda-data/redpanda/blob/4ae658a0dea73bb117f458238f2e16c2c515b6fd/proto/redpanda/core/admin/v2/shadow_link.proto#L501-L517

The example above effectively turns into a filter for the default + prod contexts, as the subject filter part (orders-value) is a subset of the default context. So perhaps a better example would be removing the `.` context from this, and it's probably worth a paragraph explaining the behaviour here.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
pgellert: It's the former. We don't validate schema ids during topic partition data replication at the moment, so it wouldn't cause replication errors there.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -0,0 +1,310 @@
...
+
+== Limitations
+
+* HTTP basic authentication and mTLS are the supported authentication methods for the source registry.
+* A shadow link replicates Schema Registry data in one mode only: either `shadow_schema_registry_topic` or `shadow_schema_registry_api`.
+* Replication is one way, from the source registry to the shadow cluster. Destination contexts owned by the link are read-only until failover.
+* Schemas that use Confluent features not supported by the Redpanda Schema Registry are not replicated as-is. Choose a <<choose-a-validation-policy,validation policy>> to control whether these schemas are skipped or imported without the unsupported fields.
+* Topic data replication and schema replication are not coordinated with each other. Records serialized in the Confluent SerDes wire format can arrive on the shadow cluster before the schema IDs they contain have been replicated.
+
```
pgellert: Sorry I missed this from the first round, but let's please also add this limitation too:

```suggestion
* Deleting and recreating a subject: A hard delete on the source replicates on the next sync, not instantly. Wait until the subject is removed from the shadow cluster before registering a new schema under the same name, as recreating it too soon prevents the sync job from detecting the subject/subject-version change.

```

pgellert (commented):
lgtm otherwise

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Applied in 84f515f1 per Gellert's explanation: the example no longer overlaps (contexts: `.prod` in full, plus the single `orders-value` subject from the default context), and a new paragraph documents the union semantics from shadow_link.proto — contexts replicate entire contexts, subjects use qualified syntax (`:.staging:orders-value`), both lists union, and overlaps replicate once.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: Extended in 84f515f1 with the impact per Gellert's answer: no replication errors (schema IDs aren't validated during topic data replication), but consumers that look up unreplicated schema IDs on the shadow cluster fail until the schemas arrive — so wait for schema replication to catch up before consuming schema-dependent shadow topics, with a pointer to Monitor replication status.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
Feediver1: @bartoszpiekny-redpanda closing the loop on this one: the page's Use cases section covers the disaster-recovery scenario explicitly ("when you maintain a Redpanda disaster recovery cluster for a system that keeps its schemas in a Confluent Schema Registry"), and the intro's final sentence names both migration and DR as the two use cases. If you'd like DR called out more prominently than that, happy to adjust — otherwise I think this is covered.

Feediver1 (commented):
## Final-pass review: DOC-2254 Confluent Schema Registry migration (#1776)

**Files reviewed:** 5 (new migration guide, setup, overview, monitor, nav)
**Overall assessment:** Content-wise the most thoroughly SME-vetted page in the shadowing set — 23/23 review threads addressed across three rounds with commit-cited replies. The final-pass single-sourcing audit found one pre-merge blocker, **now fixed in 6cedb4a5**.

### Critical issue — found and fixed (6cedb4a5)

**Four ungated xrefs to `migrate-schemas-confluent.adoc`, which has no cloud-docs consuming stub** (the cloud twin is deferred to DOC-2307). All four sat in shared regions of the single-sourced pages — verified by conditional-depth analysis — so every cloud build would have failed with `target of xref not found` once this content flowed downstream. Fixed:

- `monitor.adoc`: the "Schema Registry sync status" bullet is now `ifndef::env-cloud[]`-gated in full (it describes API-mode-only output; consistent with the adjacent role-sync bullet's treatment).
- `overview.adoc` (task tab + replication paragraph): cross-link sentences gated; the replication paragraph gets a cloud variant that mentions topic-mode replication only.
- `setup.adoc` (API-mode bullet): cross-link sentence gated.

All four sites re-audited as gated after the fix.

### Review-state summary

- All 23 human threads content-addressed; the last two (filter-union semantics per shadow_link.proto, SerDes consumer impact) landed in 84f515f1. Six threads await only resolve-clicks.
- Trevor's re-review is requested; his CHANGES_REQUESTED decision is the remaining gate.
- Mechanical lint of all changed files: clean.
- Glossterm dependencies: `shadow link`/`shadow cluster` resolve against #1777 — **merge order stands: #1777 first**, then this PR to `beta` with branch-delete so #1780 auto-retargets.

### Suggestion (non-blocking)

`overview.adoc` lines ~197–220 carry pre-existing relative xrefs (`xref:./setup.adoc[…]`, from PR #1491 — not introduced here). They happen to resolve in both builds because all four targets have cloud stubs, but they violate the full-resource-ID convention. Cheap cleanup while this file is open, or a follow-up.

### Impact notes

- cloud-docs: no action needed now that the gating is in; DOC-2307 owns the eventual cloud stub and un-gating.
- Release notes: confirm the 26.2 release-notes entry for Confluent SR replication is covered by the GA regen rather than assumed.

### Jira

DOC-2254 (ENG-1171) fully addressed for the Self-Managed scope, with the Cloud deferral now leak-free.

trevpanda (approved):
Looks great! Thank you!

micheleRP (approved):
lgtm

PR #1777, opened 2026-07-02
DOC-2254: Add Shadowing and Schema Registry context glossary terms

Feediver1:
## Description

Adds three glossary terms used by the new Confluent Schema Registry migration page in #1776:

- `shadow link` (Redpanda features)
- `shadow cluster` (Redpanda features)
- `Schema Registry context` (Redpanda features)

Related to https://redpandadata.atlassian.net/browse/DOC-2254

> **Note**: DOC-1660 (Add Shadow Linking and DR terms to Glossary) is In Progress and may cover the two shadow terms. If that work already has definitions drafted, prefer those and drop the duplicates here.

**Merge order**: This PR should merge before or together with #1776, which uses `glossterm:shadow link[]` and `glossterm:shadow cluster[]`.

## Preview pages

This PR changes term partials only, so the rendered surface is the glossary page:

- [Glossary](https://deploy-preview-1777--redpanda-docs-preview.netlify.app/streaming/25.3/reference/glossary) (updated — all three new terms verified rendering in this preview: `shadow link`, `shadow cluster`, `Schema Registry context`)

Note: this shared-branch preview assembles the 25.3 build; the `shared` branch feeds every version, so the terms will appear in all versions' glossaries once merged.

### DIFF
--- a/modules/terms/partials/schema-registry-context.adoc
+++ b/modules/terms/partials/schema-registry-context.adoc
@@ -2,5 +2,3 @@
 :term-name: Schema Registry context
 :hover-text: An independent namespace for subjects within a Schema Registry. Contexts let multiple groups of schemas coexist in one registry without naming conflicts. The default context is named ".".
 :category: Redpanda features
-
-For more information, see xref:manage:schema-reg/schema-reg-contexts.adoc[Schema Registry Contexts].
--- a/modules/terms/partials/schema-registry-context.adoc
+++ b/modules/terms/partials/schema-registry-context.adoc
@@ -1,0 +1,6 @@
+=== Schema Registry context
+:term-name: Schema Registry context
+:hover-text: An independent namespace for subjects within a Schema Registry. Contexts let multiple groups of schemas coexist in one registry without naming conflicts. The default context is named ".".
+:category: Redpanda features
+
+For more information, see xref:manage:schema-reg/schema-reg-contexts.adoc[Schema Registry Contexts].
--- a/modules/terms/partials/shadow-cluster.adoc
+++ b/modules/terms/partials/shadow-cluster.adoc
@@ -2,5 +2,3 @@
 :term-name: shadow cluster
 :hover-text: A read-only Redpanda cluster that continuously receives replicated data from a source cluster through a shadow link. During a disaster, you can fail over to the shadow cluster so it handles production traffic.
 :category: Redpanda features
-
-For more information, see xref:manage:disaster-recovery/shadowing/overview.adoc[Shadowing Overview].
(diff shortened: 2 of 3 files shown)


modules/terms/partials/schema-registry-context.adoc
```diff
@@ -0,0 +1,6 @@
+=== Schema Registry context
```
pgellert, line 1: FYI, there are a few schema registry context-related definitions here: https://docs.redpanda.com/streaming/current/manage/schema-reg/schema-reg-contexts/#key-concepts

I am not sure what is worth documenting as glossary terms vs as concepts defined on the docs page, but I just wanted to highlight those in case you think they should be defined as glossary terms as well.

pgellert (approved):

Feediver1 (commented):
## PR Review: Add Shadowing and Schema Registry context glossary terms (#1777)

**Files reviewed:** 3 glossary term partials (base: `shared`)
**Overall assessment:** Clean, accurate definitions that follow glossary conventions — including the no-links rule, now that Michele's feedback is applied in d9f9124a. One useful discovery: the overlapping ticket DOC-1660 is effectively orphaned, which resolves the PR body's open question.

### Michele's feedback — applied (d9f9124a)

Removed the "For more information" xref line from all three partials. The rationale verified against the corpus: **none of the other 133 glossary entries carries an `xref:` link** — the single remaining link (`admin-api.adoc`) is a version-independent `link:/api/doc/admin[]` URL, which is exactly the versioned-xref breakage that motivated the rule.

### Verification performed

- No duplicate term files (`shadow*` / `*context*` sweep across all 134 partials — only these three plus the unrelated `context-window.adoc`).
- Consumer usage matches exactly: #1776/#1780 use `glossterm:shadow link[]` and `glossterm:shadow cluster[]` with case-sensitive term-name matches.
- The shadow-link definition's replication list matches the Shadowing overview, correctly omitting 26.2-only role sync since the `shared` branch serves all versions.
- Category and casing conventions match existing entries.

### DOC-1660 overlap — resolved, needs Jira triage

The PR body defers to DOC-1660 (Shadow Linking glossary terms) if its definitions are further along. They aren't going anywhere: DOC-1660 is In Review but assigned to an **inactive account** (Angela Simms has left). Its definitions are engineering-internal (ShadowPanda, mirror partitions) and consistent with this PR's user-facing ones — no conflict. This PR should proceed; DOC-1660 needs reassignment or closing with these two terms counted as partial delivery.

### Suggestion

No consumer currently uses `glossterm:Schema Registry context[]` (#1776 uses plain prose in the contexts section). The term is still worth having, but consider wiring #1776's first mention to it — that was the point of adding it.

### Merge order

As the PR body states: before or with #1776, whose `glossterm:` macros resolve against these terms.

micheleRP (approved):
lgtm

PR #1796, opened 2026-07-14
chore: upgrade Antora to 3.1.15

JakeSCahill:
Bumps `@antora/cli` and `@antora/site-generator` from 3.1.2 to 3.1.15 (patch series).

Fixes the local build crash (`RangeError: Maximum call stack size exceeded` in `glob-stream/node_modules/glob`) that made `npx antora local-antora-playbook.yml` unusable. Verified: full site build now completes with exit 0; remaining log output is the usual version-mapping notices.

`main` (and other active branches) likely want the same bump — happy to open a matching PR if wanted.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,8 @@
       "version": "1.0.0",
       "license": "ISC",
       "dependencies": {
-        "@antora/cli": "3.1.2",
-        "@antora/site-generator": "3.1.2",
+        "@antora/cli": "3.1.15",
+        "@antora/site-generator": "3.1.15",
         "@asciidoctor/tabs": "^1.0.0-beta.5",
         "@redpanda-data/docs-extensions-and-macros": "^5.0.0",
         "@sntke/antora-mermaid-extension": "^0.0.6"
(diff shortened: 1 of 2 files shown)


Feediver1 (commented):
## Dependency review

**Files reviewed:** 2 (`package.json` + `package-lock.json`, 1,291-line lockfile diff)
**Overall assessment:** Clean, verified, correctly scoped dependency bump. No issues — one coordination recommendation.

### Verified

- **`package.json` contains exactly the two declared bumps** (`@antora/cli`, `@antora/site-generator` 3.1.2 → 3.1.15) — nothing else.
- **3.1.15 is the genuine upstream release and the current npm `latest`** — the head of the stable patch line, not an arbitrary version (3.2.0 exists only as an RC under the `testing` tag; staying on 3.1.x is the right conservative call).
- **All 37 changed lockfile packages sit inside the Antora dependency tree** — Antora subpackage vendored deps (`vinyl`, `replace-ext`, `picomatch`, `commander`, `yauzl`) plus the `fast-glob`/`@nodelib` family. No unrelated or newly introduced top-level packages riding along.
- **The dependency delta matches the claimed fix mechanism**: the arriving `fast-glob`/`picomatch` stack and the departing nested legacy `glob` copies are precisely upstream Antora's glob replacement in the 3.1.x series — which is why the `RangeError` crash goes away. The lockfile tells the same story as the PR body.
- **Stronger-than-claimed build verification**: beyond the local exit-0, this PR's own Netlify deploy-preview passed — Antora 3.1.15 built the entire `beta` site in CI, and `run-tests` is green.

### Recommendation

**Yes to the offered `main` twin — ideally before today's beta→main promotion.** Two reasons: `main`'s local builds presumably crash the same way today, and promoting `beta` (with the bump) over a `main` without it makes the promotion diff noisier than necessary. Same two-file change, same verification.

### What works well

A dependency PR whose body states the symptom, the root cause, the verification, and the follow-up offer — nothing to reverse-engineer.

Feediver1 (approved):
Claude made one recommendation. Your call.

PR #1802, opened 2026-07-16
docs: fix misleading shadow linking troubleshooting check (DOC-2231)

JakeSCahill:
The troubleshooting section on k-shadow-linking told users to verify `enable_shadow_linking` on **both** clusters, but only the shadow (target) cluster requires it — the prerequisites partial on the same page already says so. The mismatch misleads readers and LLM assistants using the Docs MCP treat it as a debugging gotcha even when irrelevant (DOC-2231).

Now the check targets the shadow cluster only, with a sentence explaining that the source cluster needs the property only when it also acts as the shadow cluster of another link (bi-directional topologies).

Beta counterpart PR follows. Fixes DOC-2231.


Feediver1 (approved):
one minor suggestion/nit

PR #1816, opened 2026-07-21
docs: deferred upgrade finalization on Kubernetes (DOC-2273)

JakeSCahill:
Documents deferred ("unfinalized") upgrades for Kubernetes deployments (DOC-2273, eng epic ENG-1067; core feature per ENG-1029 and the [TOI writeup](https://github.com/redpanda-data/cupboard/blob/main/engineering/core/toi/v26-2/2026-07-17-unfinalized-upgrades.md)).

## Changes

- **New `Defer upgrade finalization` section** in `upgrade:k-rolling-upgrade.adoc`: set `features_auto_finalization: false` before the upgrade (26.1.9+/25.3.15+, Enterprise), roll as usual (no operator/chart changes needed), verify with `rpk cluster upgrade status` (real observed output), soak, then `rpk cluster upgrade finalize` (irreversible) — plus the three sharp caveats (config flip does not finalize; finalize requests are not persisted across controller leadership changes; one hop at a time).
- **Roll back section** extended: with deferred finalization you can roll back after all brokers are upgraded, until you finalize.
- **Shared limitations partial** (`rolling-upgrades/upgrade-limitations.adoc`): Rollbacks and Downgrades bullets gain the deferred-finalization exception, phrased generically because the partial also serves the Linux page (DOC-2013, Joyce).
- **Troubleshooting**: the `Incompatible downgrade detected` guard after finalization, why the roll halts at the first refused broker, and the validated recovery (set the image forward; deleting the crash-looping Pod does not help because the StatefulSet recreates it at the old revision).

## Validation (kind v1.34.0, operator 26.2.1-beta.3, 3 brokers)

Full loop verified live: 26.1.13 with the flag set → roll to 26.2.1-rc4 → `ready to finalize` → image-tag rollback to 26.1.13 (cluster healthy on old binary) → re-upgrade → `finalize` → `finalized` on first poll → downgrade attempt refused with the guard signature, roll halted at broker 0 while the cluster stayed available → recovery by setting the image forward.

## Coordination

- DOC-2013 (Joyce) owns the concept page and Linux procedure; the shared partial change here is written to serve both. The `features_auto_finalization` property reference already exists; the `rpk cluster upgrade` reference pages come with the 26.2 rpk regen.
- GA framing per release decision: no beta/RC mentions.

### DIFF
--- a/modules/upgrade/pages/k-rolling-upgrade.adoc
+++ b/modules/upgrade/pages/k-rolling-upgrade.adoc
@@ -295,9 +295,84 @@ kubectl get console <console-name> --namespace <namespace> -o jsonpath='{.spec.w
 +
 If warnings are present, review them and manually update your Console configuration as needed. For details about migrating to Console v3, see xref:migrate:console-v3.adoc[].
 
+== Defer upgrade finalization
+
+By default, Redpanda finalizes an upgrade automatically as soon as every broker runs the new version: version-gated features activate and downgrade is no longer possible. Deferred finalization keeps the cluster in an unfinalized state after the roll completes, so you can soak test the new version and still <<Roll back,roll back>>, even after all brokers are upgraded. This feature requires an xref:get-started:licensing/overview.adoc[Enterprise Edition license].
+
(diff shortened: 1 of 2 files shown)


Feediver1 (approved):
Check out the suggestion

PR #1821, opened 2026-07-22
docs: fetch read coalescing (DOC-2386)

Feediver1:
Resolves https://redpandadata.atlassian.net/browse/DOC-2386

Adds documentation for fetch read coalescing, the opt-in per-shard fetch read coalescer new in 26.2 that removes redundant read CPU and fetch-response memory under high consumer fan-out.

## Changes

- **New page** `modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc`: contextualized intro (the ~N×→~1× redundancy story and an explicit use-when), how it works (coalescing key, the ready-hit/in-flight-hit/miss outcomes, weak retention, per-shard scope), enable procedure (`kafka_fetch_read_coalescing_enabled`, live toggle, ~0.5 MB/shard cache), an **Evaluate coalescing effectiveness** section with the four `vectorized_kafka_fetch_read_coalescer_*` counters and the hit-ratio formula for judging workload fit, and limitations.
- **Nav**: entry after Follower Fetching under Consume Data.
- **`follower-fetching.adoc`**: an `ifndef::env-cloud[]`-gated TIP cross-linking the new page (that page is single-sourced to cloud-docs; the new page is docs-only, so the cloud build never parses the xref).

All content is sourced from the engineering TOI: [engineering/core/toi/v26.2/2026-07-21-fetch-read-coalescing.md](https://github.com/redpanda-data/cupboard/blob/main/engineering/core/toi/v26.2/2026-07-21-fetch-read-coalescing.md) (Brandon Allard, SME), and verified against the implementation PR [redpanda#31010](https://github.com/redpanda-data/redpanda/pull/31010) (merged to dev 2026-07-13, 26.2-only, no backports): property description, coalescing-key definition, obligatory-vs-strict duplicate-read caveat, and weak-retention behavior all match the source.

## Single-sourcing for Redpanda Cloud (added after SME input)

Per Brandon Allard: fetch read coalescing **can be enabled on BYOC and Dedicated clusters**, so the page is now single-sourced:

- `// tag::single-source[]` wraps the page; companion stub PR: [cloud-docs#645](https://github.com/redpanda-data/cloud-docs/pull/645) (draft — must merge only after this content reaches docs `main` at 26.2 promotion, since cloud-docs consumes `streaming` from `main`).
- Cloud build gets: an availability line (BYOC/Dedicated), the enable procedure plus a gated pointer to cloud's cluster-configuration page, and the limitations.
- The **Evaluate coalescing effectiveness** section is `ifndef::env-cloud[]`-gated: Cloud exposes only `public_metrics` to customers, and the coalescer counters are internal `vectorized_*` metrics. (Worth a Brandon confirmation that there's no Cloud-side visibility planned.)
- The cross-link TIP on `follower-fetching.adoc` stays `ifndef::env-cloud[]`-gated for now; un-gating it is a follow-up after the cloud stub merges (otherwise the cloud build gets an xref to a page that doesn't exist there yet).

## Review notes

- `kafka_fetch_read_coalescing_enabled` is not yet in the generated cluster-properties partial (arrives with the 26.2 property regen), so the page mentions it in code font without an xref; the link can be added after regen.
- The four coalescer metrics arrive via the 26.2 internal-metrics regen.
- Deliberately no version callout in prose (the page ships in the 26.2 docs version).
- Per the ticket chain: findings from this write-up inform the final disposition of the Cloud Topics tuning-material ticket (DOC‑2129, written with a non-linking hyphen on purpose — this PR implements DOC-2386 only).

## Preview pages

- [Fetch Read Coalescing](https://deploy-preview-1821--redpanda-docs-preview.netlify.app/streaming/26.2/manage/cluster-maintenance/fetch-read-coalescing/) (new)
- [Follower Fetching](https://deploy-preview-1821--redpanda-docs-preview.netlify.app/streaming/26.2/develop/consume-data/follower-fetching/) (updated)
- [Licensing overview](https://deploy-preview-1821--redpanda-docs-preview.netlify.app/streaming/26.2/get-started/licensing/overview/) (updated: Enterprise features table)

## Updates during review

- **Enterprise feature** (leadership decision 2026-07-23): enterprise-license NOTE + license prerequisite on the page (Self-Managed only) and a new row in the Enterprise features table in `get-started:licensing/overview.adoc`.
- **Moved to Manage > Cluster Maintenance** (PM feedback): page now lives at `modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc`, next to Manage Throughput; nav, TIP xref, and licensing-table xref updated; cloud-docs#645 stub mirrored.


modules/develop/pages/consume-data/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,84 @@
...
+:page-categories: Clients, Development
+
+When many consumers fetch the same partition at the same offset concurrently (high read fan-out), the broker does redundant work: each fetch performs its own log read, its own serialization, and allocates its own copy of the response bytes. With a fan-out of N consumers, that is roughly N times the read CPU and N times the fetch-response memory for byte-identical output.
+
+Fetch read coalescing removes that redundancy. The broker reads and serializes each unique read once, and shares the single result with every concurrent (and eligible back-to-back) consumer of the same data: one read, one serialization, one buffer, fanned out to all requesters. Under aligned fan-out, read CPU and fetch-response memory drop from roughly N times to one.
+
+Fetch read coalescing is off by default. Enable it when many consumers tail the same partitions with the same fetch settings, for example fan-out delivery of one stream to many downstream applications. Workloads with misaligned or low fan-out see little to no benefit and should leave it disabled. See <<Evaluate coalescing effectiveness>> for how to measure whether it helps your workload.
```
Feediver1:
```suggestion
Fetch read coalescing is disabled by default. Enable it when many consumers tail the same partitions with the same fetch settings, for example fan-out delivery of one stream to many downstream applications. Workloads with misaligned or low fan-out see little-to-no benefit and should leave it disabled. See <<Evaluate coalescing effectiveness>> for how to measure whether it helps your workload.
```

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,84 @@
...
+----
+coalescing hit ratio =
+  (ready_hits_total + inflight_hits_total)
+  / (insertions_total + reinsertions_total + ready_hits_total + inflight_hits_total)
+----
+
+A ratio near 0 means little fan-out is being collapsed: the reads aren't aligning on the same partition, offset, and fetch settings, and the workload is unlikely to benefit from coalescing. A high ratio means the coalescer is absorbing most of the fan-out.
+
+== Limitations
+
```
Feediver1: Before enabling fetch read coalescing, familiarize yourself with the following limitations:

modules/develop/pages/consume-data/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,107 @@
+= Fetch Read Coalescing
+:description: Reduce redundant read CPU and fetch-response memory under high consumer fan-out by sharing one read result across concurrent fetches of the same data.
+:page-categories: Clients, Development
+// tag::single-source[]
+
+When many consumers fetch the same partition at the same offset concurrently (high read fan-out), the broker does redundant work: each fetch performs its own log read, its own serialization, and allocates its own copy of the response bytes. With a fan-out of N consumers, that is roughly N times the read CPU and N times the fetch-response memory for byte-identical output.
+
+Fetch read coalescing removes that redundancy. The broker reads and serializes each unique read once, and shares the single result with every concurrent (and eligible back-to-back) consumer of the same data: one read, one serialization, one buffer, fanned out to all requesters. Under aligned fan-out, read CPU and fetch-response memory drop from roughly N times to one.
```
ballard26: Not sure what `Under aligned fan-out` means in the last sentence here; there is no reference/mention to what it means for fanout to be aligned prior to this.

modules/develop/pages/consume-data/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,107 @@
...
+
+ifdef::env-cloud[]
+Fetch read coalescing is available on BYOC and Dedicated clusters.
+
+Fetch read coalescing is disabled by default. Enable it when many consumers tail the same partitions with the same fetch settings, for example fan-out delivery of one stream to many downstream applications. Workloads with misaligned or low fan-out see little-to-no benefit and should leave it disabled.
```
ballard26: Note that cloud users can't enable this themselves AFAIK. Maybe it is something they should be able to do though. Let me look into whitelisting the config option for user control.

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,107 @@
...
+ifdef::env-cloud[]
+Fetch read coalescing is available on BYOC and Dedicated clusters.
+
+Fetch read coalescing is disabled by default. Enable it when many consumers tail the same partitions with the same fetch settings, for example fan-out delivery of one stream to many downstream applications. Workloads with misaligned or low fan-out see little-to-no benefit and should leave it disabled.
+endif::[]
+ifndef::env-cloud[]
+Fetch read coalescing is disabled by default. Enable it when many consumers tail the same partitions with the same fetch settings, for example fan-out delivery of one stream to many downstream applications. Workloads with misaligned or low fan-out see little-to-no benefit and should leave it disabled. See <<Evaluate coalescing effectiveness>> for how to measure whether it helps your workload.
+endif::[]
+
+== How fetch read coalescing works
```
ballard26, line 32: Do we normally give in-depth details on how a particular feature is implemented in our docs?

modules/develop/pages/consume-data/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,107 @@
...
+* *Obligatory and strict reads are grouped apart.* A single partition and offset fetched both ways at the same time incurs one duplicate read by design.
+* *Retention is best-effort.* Completed results are only reusable by later readers while a fetch still references them; the coalescer never keeps a result alive on its own. The reliable savings are on genuinely concurrent readers of the same data, and back-to-back reuse is opportunistic.
+* *Per-shard scope.* Coalescing collapses only the fan-out that lands on the same shard. Where consumers are spread across replicas and shards, for example with follower fetching, each shard coalesces the fan-out local to it.
+* *It is a de-duplication mechanism, not a throughput mechanism.* Coalescing reduces read CPU and fetch-response memory; it does not change where reads land or reduce the total volume of unique reads.
```
ballard26: Not sure if this last point is entirely accurate or needed.

modules/develop/pages/consume-data/fetch-read-coalescing.adoc
Feediver1: Fair catch — "aligned fan-out" was shorthand this page invented and never defined. It was meant to describe the ideal case where the concurrent fetches are duplicates of each other: same partition, same offset, same fetch settings, so every fetch can share one read. Fixed in 8324dec by spelling that out instead of naming it: the sentence now reads "When all N consumers fetch the same partition at the same offset with the same fetch settings, read CPU and fetch-response memory drop from roughly N times to one." Also replaced the matching "misaligned or low fan-out" phrasing later on the page with plain language (consumers at different offsets, or few consumers per partition).

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
Feediver1: We do, or at least that is the goal. This is now a standard part of the doc template we feed to Claude, and it creates a consistent experience for users regardless of what feature they are reading about. We found that users wanted/needed more context about features--what is this? What does it do? How do I use it? When should I use it? And how does it work? Of course, we try to keep it fairly high-level. 

Are you thinking that this material should not be in the doc?

modules/develop/pages/consume-data/fetch-read-coalescing.adoc
Feediver1: K--I'll remove it.

modules/develop/pages/consume-data/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,107 @@
...
```
Feediver1:
```suggestion
```

modules/develop/pages/consume-data/fetch-read-coalescing.adoc
Feediver1: I will update to make this clear. If you end up whitelisting this option for user control before we release, please let me know asap. thx.

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
ballard26: It's fine as-is, my only thought is that the way the feature is implement is likely to change more frequently than the intention/goal of the feature. So its something that could become stale in future updates and you folks aren't pinged for any changes outside of `src/v/config` AFAIK.

modules/develop/pages/consume-data/fetch-read-coalescing.adoc
ballard26: Will do

ballard26 (approved):
LGTM

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,128 @@
...
+* `max_bytes` budget
+* Whether the read is obligatory (guarantees at least one batch and ignores the strict byte limit) or strict (honors `max_bytes`). These are grouped apart, so neither serves the other.
+
+For each fetch read, one of three things happens:
+
+* *Ready hit*: A previously completed read is still retained and covers what this fetch needs, so its result is reused with no new read.
+* *In-flight hit*: A read for the same data is already running, so this fetch awaits that read instead of starting its own.
+* *Miss*: The read runs exactly once, and its result is shared with any waiting fetches and retained for later readers. If the read fails, the error is propagated to all waiters.
+
+The shared result is reference-counted. The coalescer keeps only a weak handle to it, so it never pins memory; the requesting fetches hold the only strong references until their responses are sent. This is where the memory savings come from: coalesced consumers share one buffer instead of each holding its own copy.
```
JakeSCahill: Following on from Brandon's comment: I'd keep this section (the what/why/when/how framing is deliberate and consistent across our feature pages), but trim it to behavior instead of implementation so it doesn't quietly rot when the internals change. We only get pinged on `src/v/config` changes, so the ref-counting/weak-handle detail and the named hit/miss taxonomy are the parts most likely to drift. The metrics table under Evaluate already defines each counter, so nothing downstream depends on the taxonomy. Cloud readers also currently get the taxonomy for metrics they can't see, since the Evaluate section is `ifndef::env-cloud`.

```suggestion
Redpanda coalesces fetch reads only when they request the same data in the same way: the same partition and offset, with the same isolation level and fetch size (`max_bytes`). When matching reads arrive concurrently, or back-to-back while a previous result is still in use, the broker performs the read once and shares the single result with every requester. Reads that differ in any of these properties run independently. If a shared read fails, all coalesced fetches receive the error.

Coalesced consumers share one response buffer instead of each holding its own copy, which is where the memory savings come from. A completed result is retained only while at least one fetch still references it, so the coalescer never pins memory on its own.
```

Non-blocking, and happy for this to land as a follow-up if you'd rather not touch it now.

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+ifndef::env-cloud[]
+Before enabling fetch read coalescing, familiarize yourself with the following limitations:
+endif::[]
+ifdef::env-cloud[]
+Before requesting fetch read coalescing, familiarize yourself with the following limitations:
+endif::[]
+
+* *Only identical reads coalesce.* Consumers reading the same data but with a different `max_bytes`, a different offset, or a different isolation level do not share a read. The benefit scales with how closely the concurrent fetches match: the ideal case is many consumers tailing the same partition at the same offset with the same fetch sizing.
+* *Obligatory and strict reads are grouped apart.* A single partition and offset fetched both ways at the same time incurs one duplicate read by design.
```
JakeSCahill: If you take the How it works trim above, the obligatory/strict definition disappears from that section, so this bullet needs to carry it:

```suggestion
* *Obligatory and strict reads are grouped apart.* Redpanda distinguishes reads that must return at least one batch regardless of the byte limit (obligatory) from reads that strictly honor `max_bytes` (strict), and neither serves the other. A single partition and offset fetched both ways at the same time incurs one duplicate read by design.
```

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+Fetch read coalescing removes that redundancy. The broker reads and serializes each unique read once, and shares the single result with every concurrent (and eligible back-to-back) consumer of the same data: one read, one serialization, one buffer, fanned out to all requesters. When all N consumers fetch the same partition at the same offset with the same fetch settings, read CPU and fetch-response memory drop from roughly N times to one.
+
+ifdef::env-cloud[]
+Fetch read coalescing is available on BYOC and Dedicated clusters.
+
+Fetch read coalescing is disabled by default, and the cluster property that controls it is managed by Redpanda; see <<Enable fetch read coalescing>>. It benefits workloads where many consumers tail the same partitions with the same fetch settings, for example fan-out delivery of one stream to many downstream applications. Request it only for high fan-out workloads: at low fan-out (for example, one or two consumers per partition), the de-duplication logic saves little to nothing and its overhead can increase reactor utilization.
```
JakeSCahill: Semicolon sweep (style guide says split into separate sentences):

```suggestion
Fetch read coalescing is disabled by default, and the cluster property that controls it is managed by Redpanda. See <<Enable fetch read coalescing>>. It benefits workloads where many consumers tail the same partitions with the same fetch settings, for example fan-out delivery of one stream to many downstream applications. Request it only for high fan-out workloads: at low fan-out (for example, one or two consumers per partition), the de-duplication logic saves little to nothing and its overhead can increase reactor utilization.
```

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,128 @@
...
+ifndef::env-cloud[]
+Before enabling fetch read coalescing, familiarize yourself with the following limitations:
+endif::[]
+ifdef::env-cloud[]
+Before requesting fetch read coalescing, familiarize yourself with the following limitations:
+endif::[]
+
+* *Only identical reads coalesce.* Consumers reading the same data but with a different `max_bytes`, a different offset, or a different isolation level do not share a read. The benefit scales with how closely the concurrent fetches match: the ideal case is many consumers tailing the same partition at the same offset with the same fetch sizing.
+* *Obligatory and strict reads are grouped apart.* A single partition and offset fetched both ways at the same time incurs one duplicate read by design.
+* *Retention is best-effort.* Completed results are only reusable by later readers while a fetch still references them; the coalescer never keeps a result alive on its own. The reliable savings are on genuinely concurrent readers of the same data, and back-to-back reuse is opportunistic.
```
JakeSCahill: Same semicolon rule:

```suggestion
* *Retention is best-effort.* Completed results are only reusable by later readers while a fetch still references them. The coalescer never keeps a result alive on its own. The reliable savings are on genuinely concurrent readers of the same data, and back-to-back reuse is opportunistic.
```

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+ifndef::env-cloud[]
+Enable coalescing with the `kafka_fetch_read_coalescing_enabled` cluster property:
+
+[,bash]
+----
+rpk cluster config set kafka_fetch_read_coalescing_enabled true
+----
+
+This is a runtime change; no restart is required. The coalescing cache is fixed-size (about 0.5 MB per shard) and is allocated only while the feature is enabled; there is no separate sizing property. Setting the property back to `false` disables coalescing and clears the per-shard cache immediately.
```
JakeSCahill: Two more here:

```suggestion
This is a runtime change, and no restart is required. The coalescing cache is fixed-size (about 0.5 MB per shard) and is allocated only while the feature is enabled. There is no separate sizing property. Setting the property back to `false` disables coalescing and clears the per-shard cache immediately.
```

modules/manage/pages/cluster-maintenance/fetch-read-coalescing.adoc
```diff
@@ -0,0 +1,128 @@
...
+ifdef::env-cloud[]
+The `kafka_fetch_read_coalescing_enabled` cluster property that controls coalescing is not user-configurable in Redpanda Cloud. To enable coalescing on a BYOC or Dedicated cluster, contact https://support.redpanda.com/hc/en-us/requests/new[Redpanda Support^].
+
+Enabling or disabling coalescing is a runtime change: no cluster restart is required, and the coalescing cache (about 0.5 MB per shard) is allocated only while the feature is enabled.
+endif::[]
+
+ifndef::env-cloud[]
+== Evaluate coalescing effectiveness
+
+The coalescer exports four counters on the internal `/metrics` endpoint, one series per shard. The metrics are registered only when internal metrics are enabled (that is, when `disable_metrics` is not set).
```
JakeSCahill: "not set" reads a little ambiguous since the property always has a value:

```suggestion
The coalescer exports four counters on the internal `/metrics` endpoint, one series per shard. The metrics are registered only when internal metrics are enabled (that is, when `disable_metrics` is set to `false`, the default).
```

modules/get-started/pages/licensing/overview.adoc
```diff
@@ -128,6 +128,10 @@ Continuous Intra-Broker Partition Balancing is enabled by default for all new cl
 
 | Continuous Intra-Broker Partition Balancing is disabled.
 
+| xref:manage:cluster-maintenance/fetch-read-coalescing.adoc[Fetch Read Coalescing]
+| Shares one read result across concurrent fetches of the same data, reducing read CPU and fetch-response memory under high consumer fan-out.
+| You can no longer enable `kafka_fetch_read_coalescing_enabled`.
```
JakeSCahill: @ballard26 quick one while you're here: on license expiry, does coalescing keep running if it was already enabled (which is what this wording implies), or does it get switched off like the balancing features above? Want to make sure the row matches actual behavior. Not blocking merge on this, we can patch the row if the answer differs.

JakeSCahill (approved):
Approving, this is in good shape. The single-sourcing wiring all checks out: I verified `develop:consume-data/follower-fetching.adoc` and `manage:monitor-cloud.adoc` both exist in cloud-docs, and the cloud-docs#645 stub mirrors the final page path.

I've left some optional inline suggestions. The main one picks up @ballard26's point about implementation detail in the How it works section: keep the section, trim it to behavior. None of these need to block merge, take or leave them.

Two things I couldn't leave inline because the files aren't in this diff:

1. **Release notes**: `get-started/release-notes/redpanda.adoc` on `beta` covers the other 26.2 features but has no entry for this one yet. Something like this, maybe after the User-based throughput quotas section:

   ```asciidoc
   == Fetch read coalescing

   Redpanda can now coalesce concurrent fetch reads that request the same data with the same fetch settings, sharing one read result across all requesting consumers. For workloads where many consumers tail the same partitions, this reduces read CPU and fetch-response memory from roughly N times to one. Fetch read coalescing is an Enterprise feature and is disabled by default. See xref:manage:cluster-maintenance/fetch-read-coalescing.adoc[] to get started.
   ```

2. **Manage Throughput**: it's the natural tuning-story neighbour, so a `See also` line on `manage-throughput.adoc` pointing here would help discovery. Note that page is single-sourced to cloud-docs, so gate the link with `ifndef::env-cloud[]` until cloud-docs#645 merges (same reason as your follower-fetching TIP).

Follow-ups for after the 26.2 regens land: link `kafka_fetch_read_coalescing_enabled` back to this page via property-overrides.json and swap the code-font property mention here for an xref, and same idea for the four coalescer metrics once the internal metrics regen lands.

modules/get-started/pages/licensing/overview.adoc
ballard26: It will switch off automatically; same as the balancing features.

Feediver1:
```suggestion
| It will switch off automatically, similar to the other balancing features.
```

PR #225, opened 2024-01-05
Fix incorrect network id used in vpc-peering-aws.adoc

sbocinec:
Redpanda VPC uses Redpanda network ID in the Name tag, not the Redpanda cluster ID. This addresses the issue when BYOC customers were not able to create correct routes from Redpanda network to their peered VPC.

PR #221, opened 2024-01-04
Add connector sizing

tomasz-sadura:

PR #161, opened 2023-12-06
Deprecate delete_retention_ms and replace with log_retention_ms

Feediver1:
View the new entry under Deprecated features: https://deploy-preview-161--redpanda-docs-preview.netlify.app/23.3/upgrade/deprecated/

View the deprecation notice in the context of `delete_retention_ms`: https://deploy-preview-161--redpanda-docs-preview.netlify.app/23.3/reference/cluster-properties/#delete_retention_ms

View the new entry for `log_retention_ms`: https://deploy-preview-161--redpanda-docs-preview.netlify.app/23.3/reference/cluster-properties/#log_retention_ms

Updates in `retention.ms`: https://deploy-preview-161--redpanda-docs-preview.netlify.app/23.3/reference/topic-properties/#retentionms

Update in topic<->cluster corresponding properties table: https://deploy-preview-161--redpanda-docs-preview.netlify.app/23.3/reference/topic-properties/

PR #1834, opened 2026-07-28
ci: regenerate a single rpk plugin's docs when the plugin releases

JakeSCahill:
## Summary

rpk plugins (connect, ai, k8s, check) release independently of Redpanda: connect ~weekly, rpk ai several times a week (cloudv2 `adp/v*` tags), rpk k8s per operator release, rpk check occasionally. Their docs currently refresh only when a Redpanda release triggers full rpk regeneration.

This adds `update-rpk-plugin-docs.yml`: each plugin's release workflow sends a `repository_dispatch` with `{plugin, version}`, and this workflow refreshes **only that plugin's subtree** in the committed rpk snapshot (doc-tools `--plugin` mode, docs-extensions-and-macros#225) and opens a PR against `main`.

## Routing: main only

Plugin versions are independent of Redpanda's major.minor docs versioning, so plugin doc changes always land on `main`. `beta` receives them through the regular main→beta syncs and the RC-triggered full regenerations. A plugin whose command is missing from main's newest rpk snapshot (k8s and check, until 26.2 GA rolls into main) skips green with a notice instead of failing.

New commands found during a refresh are stamped with the **plugin's own version** in `rpk-overrides.json`, so pages render "This command was introduced in `<plugin>` version X."

PR branches are version-less (`rpk-plugin-docs/<plugin>-main`) so rapid successive releases update the same open PR instead of stacking, and a per-plugin concurrency group serializes runs.

## update-extensions.yml

Adds `beta` to the branch matrix. Without it, beta never receives doc-tools version bumps — the RC-triggered `update-rpk-docs` runs check out beta and `npm ci` from beta's lockfile, so beta must track doc-tools releases too.

## End-to-end test (already run)

Exercised in CI from this branch (temporary push trigger + doc-tools installed from the feature branch, both since removed): [run 30354665055](https://github.com/redpanda-data/docs/actions/runs/30354665055), plugin=connect, version=4.102.0. The main leg produced #1836 — exactly the expected surgical diff (snapshot `plugin_versions` + one nav line). A beta leg also ran in that test and confirmed why beta is excluded here: while beta's pages are ahead of its committed snapshot (until #1831 lands the rc2 snapshot), any direct regeneration on beta deletes 26.2 content. The test also surfaced that the `rpk cluster health` Prometheus-metrics note exists only in beta's page, not in `rpk-overrides.json` — needs restoring separately.

## Merge order

1. docs-extensions-and-macros#225 (publishes doc-tools 5.3.0; #221 recommended alongside).
2. This PR, then let the auto dependency-bump PRs land on main and beta.
3. Sender PRs (independent): redpanda-data/connect#4636, redpanda-operator#1703, redpanda-check#10, redpanda-data/cloudv2#28552 (rpk ai).

#1831 no longer blocks this PR (beta is out of the direct-generation path), but merging it remains important for beta's own consistency.

| PR | Role | Depends on |
|---|---|---|
| redpanda-data/docs-extensions-and-macros#225 | Generator: `--plugin` mode, change detection, deprecations, What's-new merge, flag extraction, stub reconciler (publishes doc-tools 5.3.0) | — |
| redpanda-data/docs#1834 | Receiver workflow for plugin-release dispatches | #225 |
| redpanda-data/docs#1844 | Release workflow: auto What's-new diff base + pre-GA plugin pins | #225 |
| redpanda-data/docs#1849 | Removes two stale rpk ai partials | merge before #1852/adp-docs#165 activate |
| redpanda-data/docs#1852 | Sender: docs main → adp-docs stub sync dispatch | #225, adp-docs#165 |
| redpanda-data/adp-docs#165 | Receiver: reconciles rpk ai stubs and nav | #225, docs#1852 |
| Senders (merged/open) | redpanda-data/connect#4636, redpanda-operator#1703 (merged), redpanda-check#10, cloudv2#28552 | docs#1834 |

Merge order: #225 → dependency bumps on docs main and beta → #1834 + #1844 (+ #1849) → #1852 + adp-docs#165 → remaining senders. Jira: DOC-2355, DOC-1090.


micheleRP (commented):
Reviewed alongside docs-extensions-and-macros#225 and #1844; tested the generator side locally end-to-end. The receiver design is right (main-only routing, version-less PR branches, per-plugin concurrency, the `rpai` slug guard), and all four senders match the contract. One critical fix needed, one behavior change to request for `ai`, and a few smaller items:

**1. (Critical) `sort -V` picks the rc snapshot over the GA snapshot.** The eligibility step resolves the newest snapshot with `ls docs-data/rpk-v*.json | sort -V | tail -1`, but GNU sort ranks `rpk-v26.2.1-rc2.json` *after* `rpk-v26.2.1.json` (verified on ubuntu:24.04), and nothing deletes superseded rc snapshots. Today main has 26.1.12 + 26.2.1-rc2; the moment the GA regen commits `rpk-v26.2.1.json`, every plugin refresh here will pass `--from-json` the **rc2 snapshot** and re-render the full tree from it — reverting GA page content to rc2 state (the same failure class the beta leg of your e2e test demonstrated). Verified fix: normalize the prerelease suffix before sorting, e.g. `sed 's/-rc/~rc/' | sort -V | sed 's/~rc/-rc/'` (`~` sorts lowest in GNU version sort) — or delete superseded `-rcN` snapshots when the GA snapshot lands. Same bug in #1844's diff-base resolution (CodeRabbit caught it there; this workflow has the identical pattern).

**2. (Critical) The backport guard is inverted for prerelease→GA.** `sort -V` also ranks `26.3.1` *before* `26.3.1-beta.1` (verified GNU + BSD). Concrete scenario: RC regens pin `k8s=26.3.1-beta.1` (what #1844 enables), beta merges to main with that recorded in `plugin_versions`, then the operator GA release dispatches `k8s 26.3.1` — the guard calls GA "older than the documented version" and skips it as a backport. Same tilde normalization on both sides of the comparison fixes it.

**3. Skip `--update-whats-new` for `plugin == ai`.** rpk ai's docs home is adp-docs (redpanda-data/adp-docs#161, now merged: this repo generates the `rpk-ai` partials, adp-docs publishes them via single-source stubs — the `rpk cloud`/cloud-docs pattern). The ADP release notes are already generated from cloudv2's `adp/RELEASE_NOTES.md` per release and cover rpk ai CLI changes with curated entries, so an `=== ai plugin <version>` block in the Self-Managed What's-new would (a) promote a non-Self-Managed surface and (b) accumulate fast at ai's release cadence (several/week, one permanent block per version). The `ai` dispatch itself should absolutely stay — it keeps the shared partials fresh. Suggest a per-plugin opt-in map for the What's-new step (connect/k8s/check keep it).

**4. Notify adp-docs when an ai refresh adds or removes commands.** With #161's static stubs, a refresh that ADDS an ai command creates a partial with no adp-docs stub/nav entry (command invisible on the ADP site), and one that REMOVES a command deletes the partial and leaves a stub with an unresolved include (broken ADP page). The run already computes exactly this (`newCommands`/`removedCommands` in the plugin diff). Ask: when `plugin == ai` and the diff has structural changes, raise a signal — open an issue in adp-docs, or add a distinctive label + reviewer on the docs PR — so the stub/nav follow-up isn't discovered via a broken ADP page. Longer term an adp-docs stub/nav regenerator off the same dispatch would close the loop.

**5. First automated PR won't be surgical.** The 5.3.0 `command.hbs` `:description:` change touches every rpk page on first regeneration (~370 files in my local runs against clean main). If a plugin release fires before the GA regen lands that churn, the first plugin PR here carries it and the auto-description's "All other pages … should show no changes" will be wrong. Consider a one-time re-render right after the dependency bump, or a note in the PR body template.

**6. (From CodeRabbit — worth taking) Validate the dispatched `version` string** before writing it to `$GITHUB_OUTPUT` (newline injection can override the whitelisted `plugin` output; it also flows into the CLI and PR title). Low real-world risk since dispatch requires write access, but cheap hardening that also catches malformed sender payloads. Note its proposed patch has a shell syntax error (`! "$VERSION" =~` outside `[[ ]]`) — adapt rather than paste.

micheleRP (approved):
Re-tested after your updates. Both items are addressed. Approving.

The generated PR now uses the actions bot token, so these PRs will pick up `setup` and `run-tests` like any human PR, which was the gap I raised. And the adp-docs follow-up comment now describes the automatic sync and tells the reviewer to merge the follow-on PR, rather than asking them to hand-edit stubs, which is what would have been wrong once #1852 and redpanda-data/adp-docs#165 land.

I re-ran the resolve-parameters guards as extracted shell, no token in scope: `banana` and `rpai` still blocked with their specific errors, `v`-prefix stripped, empty version accepted as latest, and both a newline-bearing version and shell metacharacters rejected. CI green.

Tiny nit, ignore if you like: the `rpk-ai-structural-change` label description still reads "adp-docs stubs and nav need a follow-up", which is the pre-automation framing. The comment body is the part people read, and that is now correct.

PR #239, opened 2024-01-10
#2180 update what's new

micheleRP:
fixes https://github.com/redpanda-data/documentation-private/issues/2180

This adds the fast commission and decommission feature and the glossary doc enhancement.

Preview: https://deploy-preview-239--redpanda-docs-preview.netlify.app/current/get-started/whats-new/

PR #1838, opened 2026-07-28
DOC-2407: add selfHostedOnly overrides for cloud-unavailable rpk commands

JakeSCahill:
Follow-up to #1837, which fixed 25 broken-xref build errors by hand-editing autogenerated rpk pages. Those edits would be reverted by the next rpk regeneration because the generator had no way to express them.

This PR encodes the intent in `rpk-overrides.json` so regeneration reproduces the fixes:

- Adds `selfHostedOnly: true` to the 22 commands whose reference pages exist only in the self-managed docs (cluster brokers/health/license/loggers/maintenance/partitions/self-test/upgrade, cluster config edit/export/force-reset/import/lint, cluster storage restore, transform pause/resume, topic analyze/describe-storage, shadow config, profile validate, generate license/prometheus-config).
- Documents the command-level `cloudOnly`/`selfHostedOnly` keys in a new "Cloud availability" section of the overrides guide, and fixes a module-prefix lint warning in an existing guide example.

Requires the generator support in redpanda-data/docs-extensions-and-macros PR "feat(rpk-docs): gate subcommand table rows by cloud availability". Verified by running that generator build against `rpk-v26.2.1-rc2.json` with these overrides: the regenerated pages are byte-identical to the pages fixed in #1837. The overrides are inert until a regeneration runs with the new generator, so merge order does not matter.

## Jira

Fixes [DOC-2407](https://redpandadata.atlassian.net/browse/DOC-2407): the 22 `selfHostedOnly` entries here match the ticket's 22 broken cloud-docs xref targets exactly (verified 2026-08-02). Once this merges and the rpk pages regenerate, the parent pages' rows for these commands render inside `ifndef::env-cloud[]` and the cloud-docs builds stop emitting the broken links.

[DOC-2407]: https://redpandadata.atlassian.net/browse/DOC-2407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs-data/RPK_OVERRIDES_GUIDE.adoc
+++ b/docs-data/RPK_OVERRIDES_GUIDE.adoc
@@ -331,6 +331,26 @@ Mark commands as platform-specific:
 
 Valid platforms: `linux`, `darwin`, `windows`
 
+== Cloud availability
+
+Mark a whole command as available in only one environment:
+
+[,json]
+----
+"rpk cluster health": {
+  "selfHostedOnly": true
+}
+----
+
+Use `selfHostedOnly` for commands that do not apply to Redpanda Cloud clusters, and `cloudOnly` for commands that apply only to them.
+
+Setting either key does two things:
+
+* Wraps the command's own page in an `ifndef::env-cloud[]` (or `ifdef::env-cloud[]`) conditional.
+* Wraps the command's row in the parent command's Subcommands table in the same conditional, so cloud-docs pages that single-source the table do not link to pages that only exist in the self-managed docs.
+
+NOTE: The `rpk cloud` and `rpk security secret` command families need no override. Their pages are generated into the cloud partials directory and published by cloud-docs, and their rows in regular pages automatically link across to the `cloud-data-platform` component.
+
 == Deprecation
 
 Mark commands as deprecated:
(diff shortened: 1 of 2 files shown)


micheleRP (approved):
Approving. This is the split-down version I asked for on the earlier revision, and the data does what it claims.

Merged onto current main and regenerated with the published 5.3.1: the 22 `selfHostedOnly` entries produce **69 gated rows across 42 pages**, wrapped as `ifndef::env-cloud[]` with the directives on their own lines. Spot-checked `rpk-cluster.adoc`, where brokers, health, license, loggers, and maintenance are all correctly gated.

Override validation reports zero errors with this merged in, so none of the 22 entries collide with an existing `cloudOnly` (the mutual-exclusion check from #239 would have caught that).

Going from +42390/-2486 to +74/-15 makes this genuinely reviewable, which matters for a data change that reshapes 42 published pages.

One coordination note: three other open PRs also edit `rpk-overrides.json`. I test-merged all four pairwise. This one is clean against #1860 and #1865 in either order; only #1862 conflicts. Details on #1863 where I wrote up the full ordering.

PR #1843, opened 2026-07-28
DOC-2335: Document Console UI for API-mode Schema Registry replication

micheleRP:
Documents the Redpanda Console UI for API-mode Schema Registry replication on shadow links, shipped in Console 3.9.0 (2026-07-23) via console#2557, console#2564, and console#2569.

DOC-2254 documented the whole configuration surface, but only through `rpk` and the configuration file. None of the Console UI was covered, and there is no `Redpanda Console` path anywhere in the shadowing docs today.

Jira: https://redpandadata.atlassian.net/browse/DOC-2335

## What this adds to `migrate-schemas-confluent.adoc`

- **Prerequisites**: Redpanda Console v3.9.0 or later, plus the cluster version gate. The Schema Registry fields render only when the shadow cluster reports Redpanda 26.2.0 or later; on older clusters Console falls back to the legacy `_schemas` toggle and API mode is unavailable. This is the likely "why can't I see this field?" support question.
- **`Configure schema replication in Redpanda Console`**: the create-wizard walkthrough in UI order, cross-referencing the existing configuration sections for the scope, context-mapping, and validation-policy semantics rather than restating them.
- **`Console fields and configuration keys`**: a mapping table, because the UI labels do not match the configuration keys (for example, the mode tabs are *Redpanda* / *Other* / *None*, not `shadow_schema_registry_topic` / `shadow_schema_registry_api`).
- **`Edit an existing shadow link`**: constraints that exist only in the UI and are not inferable from the configuration reference. No switching between *Redpanda* and *Other* on an existing link (Redpanda Console locks the unavailable tab; create a new link instead); switching away from *Other* discards stored credentials, scope, and sync behavior; turning off *Redpanda* mode leaves an already-added `_schemas` shadow topic in place; editing an *HTTP Basic* link requires re-entering the password.
- **`Verify the configuration`**: split into `Use rpk` and `Use Redpanda Console`, documenting the read-only Schema Registry detail card.
- **Screenshot**: the *Shadow Schema Registry* card in *Other* mode, captured from Console v3.9.0 against a Redpanda v26.2.1 broker so the version-gated fields actually render. Trimmed the field mapping table to the fields whose Console label does not name the configuration key it sets, and corrected four things the live UI contradicted: the navigation item (*Shadow Links*), the step name (*Configuration*, not *Schema Registry*), the final button (*Create shadow link*), and that *Enable TLS* is on by default.

## Placement rationale

Extended the existing page rather than adding a new one. The reader's job is unchanged (replicate schemas from a Confluent registry); Console vs `rpk` is a change of interface. A separate Console page would duplicate all four semantic blocks the page already owns, which is where drift starts.

Used a dedicated section rather than `rpk::` / `Redpanda Console::` tabs. The existing tab blocks in these pages are all inside `ifdef::env-cloud[]`, so there is no Self-Managed tab precedent to match, and tabs would restate the semantics four times. Happy to convert if reviewers prefer tabs for consistency with `setup.adoc`.

## Single-sourcing

All Console content is wrapped in `ifndef::env-cloud[]`. Cloud UI ships the same code but feature-flagged off pending Cloud API support, so this must not surface in cloud-docs yet. Tracked separately in DOC-2336. cloud-docs `main` has no copy of this page today, so there is nothing to verify on the Cloud side.

## Review outcomes

All three questions raised for review are answered. The resulting changes are in 2b6a2587.

### Backend contract - @pgellert

Fully unauthenticated access to the source Schema Registry **is** supported. The page now documents it in four places: *None* is a first-class authentication choice in the Console procedure (not just the route to mTLS), the configuration-file section covers the case (omit `auth_options`, no client certificate), the field mapping table maps *None* to an omitted `auth_options`, and the Limitations bullet is widened accordingly.

### Console UI - @r-vasquez

1. v3.9.0 is the correct documented minimum Console version for these forms.
2. The wizard step that validates the source connection and previews matching subjects and versions will ship in a later release. The page's statement that Console does not test the connection or list matching subjects before saving is accurate for v3.9.0, so it stays as written; the step gets documented when it ships.
3. Switching Schema Registry replication mode on an existing link is not supported, so the *None* round-trip is no longer documented. The page now states that the unavailable tab is locked and directs readers to create a new shadow link instead. Awaiting confirmation on that thread that this is the right guidance.

Two further corrections came out of verifying the review feedback:

- Console requires the *Password* to be re-entered when you edit a link that uses *HTTP Basic*, and it does not accept an empty field. The previous wording described `rpk shadow update` behavior, which Console does not share.
- The read-only detail card has a third password state (the date the password last changed) and a "no authentication configured" case, both now documented.

## Verification

- `npm run build` is clean: no warnings or errors for this page.
- Checked the rendered output locally. All new headings appear in the correct order, the mapping table renders as a table, and the prerequisites bullet renders inside the existing list.
- Reviewed against the docs-team-standards style guide and terminology resources: sentence-case headings, active voice, serial commas, no em dashes, full-module xrefs, imperative step openings, and minimalist UI phrasing.

## Preview pages

- [Migrate Schemas from Confluent Schema Registry](https://deploy-preview-1843--redpanda-docs-preview.netlify.app/streaming/current/manage/disaster-recovery/shadowing/migrate-schemas-confluent/) (updated)


modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -257,9 +260,92 @@ For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-creat
...
+| `unsupported_schema_feature_policy: FAIL` or `REMOVE`
+|===
+
+=== Edit an existing shadow link
+
+You can change the schema replication settings of an existing link in Redpanda Console, with the following constraints:
+
+* You cannot switch directly between *Redpanda* and *Other*. To change between these modes, select *None*, save the link, then select the mode you want. Redpanda Console locks the unavailable tab.
+* Switching away from *Other* discards the stored Schema Registry connection settings, including credentials, scope, and sync behavior. Redpanda Console warns you before you save.
+* Turning off *Redpanda* mode does not remove a `_schemas` shadow topic that the link already added. To stop shadowing that topic, fail over or delete the shadow topic after you save.
```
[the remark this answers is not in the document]
micheleRP: Addressed in e2606629.

You were right that failover and topic deletion were presented as equivalent when they are not. The original wording mirrored Console's own warning text, which is why it read that way, but Console's mode description states that topic mode "replaces the shadow cluster's Schema Registry" - so in topic mode the shadow `_schemas` topic backs that registry, and deleting it discards the replicated schemas while failing over does not.

The bullet now names the consequence and points readers at failover when they need to keep the schemas:

> Turning off *Redpanda* mode does not remove a `_schemas` shadow topic that the link already added. To stop shadowing that topic, fail over the link, or delete the shadow topic after you save. Deleting the topic also discards the schemas that the link replicated into it, so fail over instead if you need to keep them.

For the record on the other two findings in this review, both assessed and declined:

**`Use rpk` conditional (lines 346-355)** - the asymmetry is intentional. Only the heading is conditional, because only the heading is made necessary by the sibling `Use Redpanda Console` subsection that Cloud builds exclude. In Cloud the rpk content sits directly under *Verify the configuration*, which is the page's original structure. Guarding through line 355 as suggested would remove `rpk shadow describe` from Cloud output entirely, and that is a documented Cloud command (`modules/reference/pages/rpk/rpk-shadow/rpk-shadow-describe.adoc` ships in cloud-docs), so it would be a regression rather than a cleanup.

**Title-resolving xref (`xref:console:index.adoc[]`)** - that page's title is "Introduction to Redpanda Console", so empty brackets would render the sentence as "Introduction to Redpanda Console writes the same schema replication settings as a configuration file." Explicit link text is also the house convention here: `xref:console:index.adoc[Redpanda Console]` appears 4 times in this repo versus 1 empty-bracket use, including in `setup.adoc`, the sibling page in this same family.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -257,9 +260,100 @@ For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-creat
...
+
+| *Unsupported schema features*: *Fail the sync* or *Remove unsupported features*
+| `unsupported_schema_feature_policy: FAIL` or `REMOVE`
+|===
+
+=== Edit an existing shadow link
+
+You can change the schema replication settings of an existing link in Redpanda Console, with the following constraints:
+
+* You cannot switch directly between *Redpanda* and *Other*. To change between these modes, select *None*, save the link, then select the mode you want. Redpanda Console locks the unavailable tab.
```
r-vasquez: Is better to not document this. Switching between modes is not allowed by Redpanda, and the API will fail if you do this.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -257,9 +260,100 @@ For detailed command options, see xref:reference:rpk/rpk-shadow/rpk-shadow-creat
...
+* Switching away from *Other* discards the stored Schema Registry connection settings, including credentials, scope, and sync behavior. Redpanda Console warns you before you save.
+* Turning off *Redpanda* mode does not remove a `_schemas` shadow topic that the link already added. To stop shadowing that topic, fail over the link, or delete the shadow topic after you save. Deleting the topic also discards the schemas that the link replicated into it, so fail over instead if you need to keep them.
+* Redpanda Console never populates a password field with the stored value. Leave the field empty to keep the existing credential, or enter a new value to replace it.
```
r-vasquez: Right now, Console always asks for the Password on updates. So leaving the field empty will also ask for the credentials; I think we should remove this.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
micheleRP: Removed the *None* round-trip in 2b6a2587. The bullet now reads:

> You cannot switch an existing link directly between *Redpanda* and *Other*. Redpanda Console locks the unavailable tab. To use a different Schema Registry replication mode, create a new shadow link.

I kept a one-line statement of the constraint because the locked tab is visible in the UI and readers will ask why it is locked. One thing to confirm: is "create a new shadow link" the right guidance to give here?

A small heads-up while you're in this code: the edit section in v3.9.0 still describes the round-trip as a supported route ("the link must go through None (and a save) first. From None both modes are reachable."), and the locked-tab hint reads "Switching between Redpanda and Other isn't supported." Might be worth aligning that comment and hint text with the actual behavior so the next reader of the file doesn't reach the same conclusion I did.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
micheleRP: Yes, replaced in 2b6a2587. Verified: `addSrBasicAuthIssues` in `create/model.ts` at v3.9.0 raises "Password is required when HTTP Basic is enabled" for any api-mode form with *HTTP Basic* selected, and the edit page uses the same `FormSchema` (the password-exempt `FormSchemaWithoutSchemaRegistryRules` applies only while the feature gate is closed). The old bullet described `rpk shadow update`, which does accept an empty password to keep the stored one; Console does not share that behavior.

New wording:

> When you edit a link that uses *HTTP Basic*, Redpanda Console requires you to re-enter the *Password*. It never populates the field with the stored value, and it does not accept an empty field.

While checking this I also corrected the read-only detail card in the verification section, which is a different code path: `formatPasswordStatus` has a third state, so *Password* is now documented as *Set*, *Not set*, or the date it last changed, and the card's "No authentication configured." case is covered now that no-auth is documented.

Feediver1 (commented):
## Docs team standards review (automated final pass)

**Files reviewed:** 1 .adoc file (+117 lines net, 6 commits)
**Overall assessment:** Excellent shape — the PR has already absorbed three reviewers' feedback with source-cited fixes, and independent verification passes clean. Nothing critical.

### Independent verification

- **Both build variants render structurally clean.** Rendered the full page with asciidoctor in self-managed and cloud modes: the cloud variant collapses exactly to the original page shape — no orphaned headings, no empty Verify section, and the commit-2 colon fix does its job. This gating quality matters because DOC-2336's future stub will consume this page.
- **Console 3.9.0 claim verified** — console#2557 ("UX: New Schema Registry Sync options in Shadow Link (Create Only)") is merged.
- **Cloud xref targets pre-verified for DOC-2336**: cloud-docs has `rpk-shadow` reference pages, so the cloud build's `rpk shadow describe` xrefs will resolve once a stub exists.

### Suggestion (non-blocking)

- **[PR body, "What this adds"]** The summary bullets still describe pre-review content: "No direct Redpanda <-> Other switch (route through *None* and save)" and "password fields stay empty to preserve the stored secret" — both reversed by 2b6a2587. The "Review outcomes" section is correct, so the body currently contradicts itself; worth a quick edit so the squash commit doesn't preserve the stale claims.

### Before merging

- Rogger's confirmation is still pending on the open question: is "create a new shadow link" the right guidance for changing replication mode?

### Follow-ups

- **Console repo**: the v3.9.0 edit-page text still describes the unsupported *None* round-trip as a valid path (as noted on the thread) — worth confirming a console issue exists for that UI bug.
- **DOC-2336**: since Cloud UI is feature-flagged off pending Cloud API support, the Cloud slice waits on the flag — though wiring the cloud-docs stub for this page (whose rpk/config content is cloud-valid, with the Console sections self-gating) could land earlier as scoped work on that ticket.

### What works well

- The review loop here is a model: every correction cites Console source at the shipped tag (`schema-registry-request.ts`, `addSrBasicAuthIssues` in `create/model.ts`), and Console's own misleading warning text was caught rather than mirrored.
- The cloud gating anticipates DOC-2336 precisely — commit 2 fixed a cloud-rendering artifact on a page no cloud build even consumes yet.
- The field-mapping table (UI labels → config keys) is exactly the artifact readers need to move between interfaces.

trevpanda (approved):
Looks good.

modules/manage/pages/disaster-recovery/shadowing/migrate-schemas-confluent.adoc
```diff
@@ -55,6 +55,9 @@ Schema replication settings live in the shadow link configuration. Two cluster p
 == Prerequisites
 
 * A cluster running Redpanda version 26.2 or later. The schema replication feature activates after all brokers complete the upgrade.
+ifndef::env-cloud[]
+* To configure schema replication in xref:console:index.adoc[Redpanda Console] rather than with `rpk`, Redpanda Console v3.9.0 or later. The Schema Registry fields appear only when the shadow cluster reports Redpanda 26.2.0 or later. On an earlier cluster, Redpanda Console falls back to a single toggle that enables `_schemas` topic replication, and API mode is unavailable.
```
kbatuigas:
```suggestion
* To configure schema replication in xref:console:index.adoc[Redpanda Console] rather than with `rpk`, Redpanda Console v3.9.0 or later. The Schema Registry fields appear only when the shadow cluster reports Redpanda 26.2 or later. On an earlier cluster, Redpanda Console falls back to a single toggle that enables `_schemas` topic replication, and API mode is unavailable.
```
nit: not sure if we ever do `.0` releases for Core? I think we start from `.1` afaict

kbatuigas (approved):
Pointing out just one nit, but lgtm

PR #1844, opened 2026-07-29
ci: fix What's-new diff base and accept pre-GA plugin pins

JakeSCahill:
## Summary

Two related fixes to the RC/stable rpk docs workflow.

### 1. Diff against the newest committed snapshot

The What's-new diff never fired automatically: the diff base was resolved from antora.yml's `full-version`, then looked up as `docs-data/rpk-v<full-version>.json` — a file that doesn't exist whenever full-version has advanced past the newest snapshot. That was every RC run on beta (snapshots carry `-rcN` suffixes), and now main too (post-GA full-version is `26.2.1`, newest snapshot is `rpk-v26.2.1-rc2.json`, so the GA run would diff against itself and skip).

Now the diff base is the newest committed `rpk-v*.json` on the target branch — the version the docs currently document, the correct What's-new baseline in every case. Explicit `diff_version` still wins; the self-diff guard still covers re-dispatches.

| Dispatch | Old behavior | New behavior |
|---|---|---|
| `v26.2.1` (GA) | diff vs itself → skipped | diff vs `v26.2.1-rc2` |
| `v26.2.1-rc2` (re-run) | skipped | skipped (self-diff guard) |
| `v26.2.2` (future) | snapshot missing → skipped | diff vs newest documented |

### 2. Per-plugin version pins for pre-GA plugins

Full regenerations install the latest published version of every plugin, so a beta-only plugin with no version promoted to latest installs nothing and its commands drop out of the docs — how the `rpk k8s` pages went missing during the 26.2 beta window (#1831). A new `plugin_pins` input/payload field (comma-separated `name=version`, e.g. `k8s=26.3.1-beta.1`) forwards `--plugin-pin` flags to the generator so RC regens can install pre-GA plugin versions. Malformed pins fail loudly. Requires doc-tools ≥ 5.3.0 (docs-extensions-and-macros#225), which the same release train delivers.

## Verification

- Diff-base shell logic tested against main's real snapshot state (table above).
- End-to-end: production run [30433604980](https://github.com/redpanda-data/docs/actions/runs/30433604980) exercises the identical downstream path (diff → What's-new → PR) via explicit `diff_version=v26.1.12` for the v26.2.1 GA regen.
- `--plugin-pin` validated in doc-tools (#225) with retry-without-pin fallback and `plugin_versions` recording.

| PR | Role | Depends on |
|---|---|---|
| redpanda-data/docs-extensions-and-macros#225 | Generator: `--plugin` mode, change detection, deprecations, What's-new merge, flag extraction, stub reconciler (publishes doc-tools 5.3.0) | — |
| redpanda-data/docs#1834 | Receiver workflow for plugin-release dispatches | #225 |
| redpanda-data/docs#1844 | Release workflow: auto What's-new diff base + pre-GA plugin pins | #225 |
| redpanda-data/docs#1849 | Removes two stale rpk ai partials | merge before #1852/adp-docs#165 activate |
| redpanda-data/docs#1852 | Sender: docs main → adp-docs stub sync dispatch | #225, adp-docs#165 |
| redpanda-data/adp-docs#165 | Receiver: reconciles rpk ai stubs and nav | #225, docs#1852 |
| Senders (merged/open) | redpanda-data/connect#4636, redpanda-operator#1703 (merged), redpanda-check#10, cloudv2#28552 | docs#1834 |

Merge order: #225 → dependency bumps on docs main and beta → #1834 + #1844 (+ #1849) → #1852 + adp-docs#165 → remaining senders. Jira: DOC-2355, DOC-1090.

### DIFF
--- a/.github/workflows/update-rpk-docs.yml
+++ b/.github/workflows/update-rpk-docs.yml
@@ -15,6 +15,9 @@ on:
         required: false
         default: 'true'
         type: boolean
+      plugin_pins:
+        description: 'Comma-separated plugin version pins for pre-GA plugins (e.g. k8s=26.3.1-beta.1). Without a pin, a plugin with no promoted latest version installs nothing and its commands are absent from the docs.'
+        required: false
   repository_dispatch:
     types: [update-rpk-docs]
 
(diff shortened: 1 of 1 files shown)


micheleRP (commented):
Reviewed alongside docs-extensions-and-macros#225 and #1834. The diff-base fix is correct for the case it targets — I verified against main's real snapshot state that a v26.2.1 GA dispatch now diffs against `v26.2.1-rc2` (previously self-diff → skip), and the `plugin_pins` validation fails loudly on malformed input as intended. One issue to fix before merge:

**CodeRabbit's `sort -V` finding is real — confirmed empirically on ubuntu:24.04 (what the runners use).** GNU sort ranks `rpk-v26.2.1-rc2.json` *after* `rpk-v26.2.1.json`, and nothing deletes superseded rc snapshots. So the very first GA run this PR fixes also creates the failure state: once `rpk-v26.2.1.json` is committed next to the lingering rc2 file, every subsequent auto-resolved diff base (v26.2.2 and on) is `v26.2.1-rc2` — silently re-reporting rc2→GA changes in every What's-new, and the self-diff guard misses GA re-dispatches.

Verified fix: normalize prerelease suffixes before sorting —

```bash
NEWEST_SNAPSHOT=$(ls docs-data/rpk-v*.json 2>/dev/null | grep -v 'rpk-diff' | sed 's/-rc/~rc/' | sort -V | sed 's/~rc/-rc/' | tail -1)
```

(`~` sorts lowest in GNU version sort, so `-rc2` correctly ranks below GA; rc-vs-rc ordering like rc2 < rc10 is preserved.) Alternatively, delete superseded `rpk-vX.Y.Z-rc*.json` files when the GA snapshot is committed — that also keeps `docs-data/` tidy.

Note the same `sort -V` pattern (newest-snapshot selection *and* the plugin version backport guard) exists in #1834's `update-rpk-plugin-docs.yml` — details on that PR.

micheleRP (approved):
Approving. Verified both the diff-base fallback and the pre-GA plugin pin against real runs, not by reading. Detail in my comment above.

PR #1846, opened 2026-07-29
Fix all open issues: Kind guide template, stale Helm chart links, property name typo, Console consumer visibility

JakeSCahill:
## Summary

This PR resolves the open issues in this repo, one commit per issue. All fixes were reproduced before fixing and verified in a local Antora build after fixing.

Fixes #1754 ([DOC-2353](https://redpandadata.atlassian.net/browse/DOC-2353))
Fixes #1757
Fixes #1657
Fixes #1341

### #1754 — Missing `{{ end }}` closers in the local Kind guide

The `/etc/hosts` mapping command in `local-guide.adoc` had a Go template with 7 opening constructs but only 5 `{{ end }}` closers (truncated during the #1304 restructure), so it failed with `unexpected EOF`. Added the two missing closers. Verified the broken template fails and the fixed template parses cleanly with Go `text/template`, matching the balanced template in the shared `external-access-steps.adoc` partial.

### #1757 — Stale `helm-charts` repo links

The charts moved to `redpanda-data/redpanda-operator`, leaving 404 links. Fixed the reported `values.yaml` link on the Redpanda Helm spec page plus four more stale links found in a sweep (console, connectors, and operator Helm spec pages, and the `create_tls.sh` TLS example in `k-secrets.adoc`, now pinned to a permalink in the legacy repo since the script was never migrated). Every replacement URL was verified to return 200. Also added a sed post-processing step to `.github/workflows/generate-crd.yml` so the auto-generated Helm spec pages keep the corrected URLs on the next regeneration.

### #1657 — `default_topic_replication` missing the trailing "s"

The real property name is `default_topic_replications`. The wrong name is produced by the doc-tools property extractor, which harvests the C++ member identifier instead of the property name string literal in `configuration.cc` (regression first visible in the 25.3 docs). The same bug also affects `minimum_topic_replications`. This PR hand-corrects both names across the generated property partials, `docs-data/property-overrides.json`, the hand-written topic properties page, and the downloadable property JSON attachments. This also repairs the previously broken `#default_topic_replications` xref in `k-manage-topics.adoc`.

⚠️ Note: the edits to `modules/reference/partials/properties/` and the attachments are an interim fix. Regenerating property docs will revert them until the extractor fix lands in `docs-extensions-and-macros` (upstream fix: redpanda-data/docs-extensions-and-macros#233).

### #1341 — Spark consumers not visible in Console

[...]

PR #1848, opened 2026-07-29
ci: run doc tests on version bumps and support version overrides

JakeSCahill:
## Summary

Closes the trigger gap that let the Console v3.9 UI rework break the quickstart doc test silently: the version-bump PR never ran the test (antora.yml is not in the `quickstart` paths filter (since #1846 it is in `kindguide`, so the kind-guide test already runs on version bumps — this PR extends that to the quickstart test)), so the breakage only surfaced in nightly runs, which then failed for ~92 consecutive nights until #1846 fixed the selectors.

## Changes

1. **Run the quickstart test on version-bump PRs.** `antora.yml` pins the Redpanda and Console versions the test runs against, so a new `versions` paths filter now triggers the PR test step when it changes. UI or image breakage from a version bump is caught before merge instead of by the nightly.
2. **Support version overrides for dispatch and manual runs.** The `workflow_dispatch` trigger gains optional inputs (`redpanda_version`, `redpanda_docker_repo`, `console_version`, `console_docker_repo`), and the dispatch/schedule test step maps them (or the same keys from a `trigger-tests` `client_payload`) to the env vars the test spec already resolves before falling back to antora.yml. This makes the previously unused `trigger-tests` dispatch hook actually useful: upstream repos can now ask the docs repo to test a specific version combination, and RC combos are testable manually today (for example `redpanda_docker_repo: redpanda-unstable`).

No behavior change for scheduled runs or ordinary PRs: with no overrides set, the env vars are empty and the test resolves versions from antora.yml exactly as before.

## Upstream proposals enabled by this PR (not included)

- **Console GA**: the console repo's existing `repository-dispatch.yml` can add a `trigger-tests` dispatch to this repo on `v*` tags with `console_version` in the payload. Caveat for the Console team: at tag-push time the release image may not be published yet, so the dispatch may belong at the end of the release pipeline instead.
- **Redpanda RC**: Redpanda images build in Buildkite, so the dispatch belongs at the end of the RC build pipeline (after images publish to `redpanda-unstable`) with `redpanda_version` + `redpanda_docker_repo: redpanda-unstable` in the payload. A GitHub tag-push trigger would race image publication.

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -5,6 +5,23 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
   workflow_dispatch:
+    inputs:
+      redpanda_version:
+        description: 'Redpanda version to test, such as v26.2.1. Defaults to the version pinned in antora.yml.'
+        required: false
+        type: string
+      redpanda_docker_repo:
+        description: 'Redpanda Docker repo, such as redpanda or redpanda-unstable (for RCs). Defaults to the repo pinned in antora.yml.'
+        required: false
+        type: string
+      console_version:
+        description: 'Redpanda Console version to test, such as v3.9.0. Defaults to the version pinned in antora.yml.'
+        required: false
+        type: string
+      console_docker_repo:
+        description: 'Console Docker repo, such as console. Defaults to the repo pinned in antora.yml.'
+        required: false
+        type: string
   repository_dispatch:
     types: [trigger-tests]
 
@@ -15,6 +32,7 @@ jobs:
       console: ${{ steps.filter.outputs.console }}
       quickstart: ${{ steps.filter.outputs.quickstart }}
       kindguide: ${{ steps.filter.outputs.kindguide }}
+      versions: ${{ steps.filter.outputs.versions }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -30,6 +48,10 @@ jobs:
             kindguide:
               - 'modules/deploy/pages/redpanda/kubernetes/local-guide.adoc'
               - 'modules/deploy/partials/kubernetes/**'
+            # antora.yml pins the Redpanda and Console versions the quickstart
+            # tests against. Version-bump PRs must run the test so UI or image
+            # changes are caught before merge, not by the nightly run.
+            versions:
               - 'antora.yml'
 
   run-tests:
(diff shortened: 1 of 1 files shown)


micheleRP (commented):
## Docs-team-standards review

**Files reviewed:** 1 (`.github/workflows/test-docs.yml`, +33/−1)
**Overall assessment:** The change is correct and the part most likely to be silently broken — the env var names — checks out exactly. But the PR is currently **CONFLICTING** and its stated premise has been overtaken by #1846, so it needs a rebase and a description touch-up before it can land.

### What this PR does

Adds a `versions` paths filter on `antora.yml` so version-bump PRs run the quickstart doc test, and adds four optional `workflow_dispatch` inputs (`redpanda_version`, `redpanda_docker_repo`, `console_version`, `console_docker_repo`) mapped — along with the same keys from a `trigger-tests` `client_payload` — onto the env vars the test spec already honours. No behavior change when the overrides are unset.

### Jira ticket alignment

No linked Jira ticket. The PR references #1846 as the fix for the underlying selector breakage and frames itself as closing the trigger gap that hid it.

### Critical issues

1. **The PR cannot merge as-is: `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`.** The branch has one commit from 2026-07-29 and hasn't been updated since. #1846 merged 2026-07-30 and rewrote the same file — it added the whole `test-kind-guide` job, added `issue_body` to the "Run all tests" step (which is why this diff's context lines don't match current main), and added `antora.yml` to the `kindguide` filter.
   - **Fix:** rebase on main and re-verify the hunks still apply where intended.

2. **The premise in the description is no longer accurate.** "antora.yml is not in the paths filter" was true when written, but as of #1846 `antora.yml` **is** in the filter — under `kindguide` (line 33 on main). The substance of the PR still holds, because it's not in the `quickstart` filter, so version bumps still don't run the quickstart test. But as written, a reviewer checking main will conclude the premise is wrong.
   - **Fix:** reword to "antora.yml is not in the `quickstart` filter" (and note the kind guide already covers it).

### Suggestions

3. **After the rebase, `antora.yml` will appear in two filters.** That's harmless — dorny/paths-filter evaluates filters independently — and the new `versions` filter carries a comment explaining its purpose, which is the right call. Just make it a deliberate outcome of the rebase rather than a surprise: keep the dedicated `versions` filter (it's what gates the quickstart step and its intent is clearer) and leave `kindguide` alone, since removing `antora.yml` there would change kind-guide behavior.

4. **The payload key names are now a cross-repo API with no durable home.** `trigger-tests` plus `redpanda_version` / `redpanda_docker_repo` / `console_version` / `console_docker_repo` will be referenced by the console repo and the Redpanda RC pipeline, but they're documented only in this workflow's `env:` block and this PR description. A short comment in the workflow (or a line in the testing docs) naming the contract would keep an upstream sender from guessing.

5. **The coupling to `tests/setup-tests/fetch-versions-and-rpk.json` is invisible from the workflow.** That spec is what makes these overrides work; if someone renames a variable there, the override silently stops applying and the test quietly falls back to antora.yml with no failure. A one-line comment pointing at the spec would make the dependency discoverable.

6. **Nit:** the overrides are attached only to the "Run all tests" step, not to "Test Redpanda Streaming quickstart". That's correct today (PR-triggered runs have no overrides to apply), but if anyone later wants a dispatch that exercises only the quickstart step, the `env:` block will need duplicating.

### Impact on other files

- **`redpanda-data/console`** — I checked `.github/workflows/repository-dispatch.yml`: it currently sends `event-type: release` and `event-type: push`, both to `redpanda-data/console-enterprise`. So the description's framing is accurate — a `trigger-tests` dispatch to docs would be new, and it needs a token with dispatch permission on this repo, which is usually the blocker for cross-repo dispatch. Worth confirming that before promising the Console team it's a small change.
- **Redpanda RC pipeline (Buildkite)** — out of this repo entirely; the caveat about dispatching only after images publish to `redpanda-unstable` is well-taken and worth keeping in the ticket that tracks it, not just this PR description.
- **No `.adoc` files touched**, so the style checklist doesn't apply and no page previews are needed.

### What works well

- **All four env var names are correct**, verified against `tests/setup-tests/fetch-versions-and-rpk.json`: `REDPANDA_VERSION`, `REDPANDA_DOCKER_REPO`, `REDPANDA_CONSOLE_VERSION`, `CONSOLE_DOCKER_REPO`. The spec's fallback is exactly as the description claims — `if [ -n "$X" ]; then echo $X; else npx doc-tools get-…-version --from-antora …` — so overrides take precedence and an unset value resolves from antora.yml. This is the claim most likely to be silently wrong, and it's right.
- The `inputs.* || client_payload.*` chain is correct for both trigger types: `inputs` is null on `repository_dispatch` and `client_payload` is null on `workflow_dispatch`, so each event picks up its own source.
- Input descriptions state both the format and the default source, and the `redpanda-unstable` example makes the RC use case immediately usable.
- The filter comment explains why `antora.yml` gates a test rather than restating what the YAML does.

micheleRP (approved):

PR #1861, opened 2026-07-31
Correct swapped %a and %% escape descriptions in topic format partial

JakeSCahill:
## What

Ground-truth review of the rpk v26.2.1 `--help` output against the published docs found two factual errors in `modules/reference/partials/topic-format.adoc`, which is included by both `rpk topic consume` and `rpk topic produce`:

- `%a` and `%%` had their descriptions swapped. The help defines `%a` as record attributes and `%%` as the percent sign. The published table says the opposite on both pages.
- The strftime wrap sample reads `%d{strftime=== [%F] ===}` while the prose below it still explains "the surrounding three # on each side". The help sample is `%d{strftime### [%F] ###}`. Restored the sample so it matches both the help and its own explanation.

## How found

Part of the end-to-end review of the rpk docs automation train (see redpanda-data/docs-extensions-and-macros#225): every generated page was compared against `--help` dumps from the same rpk binary. This partial is hand-maintained rather than generated, so it needs a manual fix.

Note: the same review found that this partial documents consume semantics that do not all apply to `rpk topic produce` (produce has different `ascii`/`bool` semantics, encoded rather than raw sizes, and a smaller escape set). That is a larger content split best handled separately.

See redpanda-data/docs-extensions-and-macros#225 for the train overview.

## Jira

Part of [DOC-2408](https://redpandadata.atlassian.net/browse/DOC-2408) (stale rpk overrides hide or contradict source help).

### DIFF
--- a/modules/reference/partials/topic-format.adoc
+++ b/modules/reference/partials/topic-format.adoc
@@ -37,8 +37,8 @@ The percent encodings are represented like this:
 |`%[` |Partition log start offset
 |`%\|` |Partition last stable offset
 |`%]` |Partition high watermark
-|`%%` |Record attributes (formatting described below)
-|`%a` |Percent sign
+|`%%` |Percent sign
+|`%a` |Record attributes (formatting described below)
 |`%{` |Left brace
 |`%}` |Right brace
 |`%i` |Number of records formatted
@@ -107,7 +107,7 @@ An arbitrary amount of brackets (or braces, or # symbols) can wrap your date
 formatting:
 
 ```go
-%d{strftime=== [%F] ===}
+%d{strftime### [%F] ###}
 ```
 
 This prints `[YYYY-MM-DD]`, while the surrounding three # on each


micheleRP (approved):
Approving. Two things verified beyond reading the diff:

**The swap matches the source.** The extracted help in the committed snapshot reads `%a    record attributes (formatting described below)` — so the partial had the two escapes exactly backwards, and this ordering matches the binary. The second hunk (`===` to `###` in the strftime example) also makes the example agree with its own following prose, which already said "the surrounding three # on each".

**The vehicle is right.** `topic-format.adoc` is hand-maintained, not generator-owned: it carries no autogenerated header, nothing in the rpk-docs generator references it, and its history is hand commits. So a direct edit sticks — this is not the edit-a-generated-page trap, and no regeneration will revert it.

Independent of the release sequencing on the other PRs; can merge anytime. It does not touch the overrides file, so it triggers no rerender.

PR #1862, opened 2026-07-31
Restore help content hidden by stale rpk description overrides

JakeSCahill:
## What

A ground-truth review of every generated rpk reference page against the v26.2.1 binary's `--help` output (part of the automation work in redpanda-data/docs-extensions-and-macros#225) found that `description` overrides in `docs-data/rpk-overrides.json` had replaced the source help on dozens of commands. Because overrides win on every regeneration, the automation would faithfully reintroduce the stale content forever. This PR fixes the overrides so regeneration produces accurate pages, and pairs with a new generator report that keeps future drift visible.

## Changes by class

**Curated text that contradicts the current binary, replaced by the help:**
- `rpk cluster config lint` said "identify" where the command *removes* deprecated content
- `rpk cluster logdirs describe` claimed record counts where the command reports data sizes
- `rpk generate app` listed five supported languages where only `go` exists
- `rpk cluster partitions balance` claimed automatic balancing (Enterprise-only) and lost the CE/EE distinction
- `rpk cluster self-test start` carried an outdated test catalog (missing the dsync and 16 KB tests)
- `rpk group seek` described old `--to` semantics

**One-line summaries dropped so the full help renders** (concepts, preconditions, file formats, per-field descriptions): rpk group family, cluster maintenance family, cluster health, cluster storage mount/unmount, cluster partitions list/move, cluster license set, generate license, generate grafana-dashboard, profile/profile set/profile prompt, shadow status, topic delete/list/trim-prefix/describe-storage, transform logs.

**Curated summaries kept, hidden help paragraphs appended:** rpk security acl (ACL components, multiplicative basis, safety guidance), ai auth login, ai oauth-client dcr (admission modes), ai oauth-client revoke-tokens, connect echo/install/lint/list/template/test (exit codes, examples, experimental warning), cluster quotas import (import schema), shadow describe, topic create, transform init, debug remote-bundle download, cloud mcp install.

**Produce/consume format split:** `rpk topic produce` included the shared consume-oriented `topic-format.adoc` partial, which documents consume semantics that are wrong for produce (raw vs encoded sizes, print vs parse modifiers, a different escape set). Added `topic-format-produce.adoc` written from the produce help and switched the produce page and override to it. Also restored the consume unpack legend and connection-behavior section, and fixed the consume header example to match its own prose.

**Curated details the help lacks are preserved**, not deleted: topic delete irreversibility caution, maintenance enable upgrade guidance, cluster config propagation guidance, Serverless and BYOC notes, all xref-enriched items.

## Why the automation won't reintroduce this

- Overrides are the durable store: these fixes apply to every future regeneration.
- doc-tools now prints a "Curated descriptions replacing source help" table in every regen PR summary (redpanda-data/docs-extensions-and-macros#225), so any override hiding substantial help text is visible to reviewers instead of silent. After this PR, that table lists only six intentionally curated commands.
- The generator defects that mangled restored content (code samples as prose, caption transforms corrupting example commands, double periods) are fixed with unit tests in #225.

## Validation

Regenerated the full rpk tree with the updated overrides and doc-tools 5.3.0, then re-ran the help-comparison sweep across all 344 command pages: zero critical findings, all 364 files parse clean under Asciidoctor, and every remaining description override is intentional curation. Rendered pages ship in the next regeneration (the planned v26.2.1 re-dispatch), except the produce include swap, which is applied directly so the wrong format docs stop rendering now.

See redpanda-data/docs-extensions-and-macros#225 for the train overview. Merge alongside #1860 and #1861 (non-overlapping edits to the same files).

## Jira

Part of [DOC-2408](https://redpandadata.atlassian.net/browse/DOC-2408) (stale rpk overrides hide or contradict source help).


micheleRP (approved):
Approving, with one rebase note. This got a full verification pass rather than a read: I diffed the overrides semantically against the merge base, spot-checked the factual claims against the extracted help in the committed snapshot, and regenerated with the branch's overrides on top of current main.

## Verified

- **True change set vs merge base: 6 entries removed, 44 edited, 0 added.** Worth stating explicitly because a naive comparison of this branch's file against today's `main` shows three phantom `exclude: true` additions for the `rpk ai connection` family — that is base drift (#1869 removed those excludes on main after this branch was cut), and this PR's diff never touches them. Merging keeps main's removal, so the connection pages are unaffected.
- **All five factual claims I spot-checked hold** against the source help in the snapshot: `rpk cluster config lint` *removes* deprecated content where the curated text said "identify"; `rpk cluster logdirs describe` reports data *sizes* where the curated text claimed record counts; `rpk cluster partitions balance` has the CE/EE distinction the curated text erased; `rpk generate app` and `rpk group seek` as described.
- **Regenerated with these overrides on current main: no new validation errors** from any of the ~50 touched entries. (The 4 stale-path errors and the duplicate `== Example` warning in the log are main's pre-existing issues, owned by #1865 and #1860.)
- **The two committed pages are byte-stable under regeneration** — they are exactly what the generator produces from the new overrides, so nothing here drifts at the next regen.
- **62 pages change when the tree regenerates** — committing only the 2 and letting the rest propagate through the re-render pipeline is consistent with #1863's design, and worth knowing when that PR lands so the 60-page auto-PR is not a surprise.

## Rebase note

This is the one PR in the overrides queue that conflicts: with #1838 and #1865, in either direction. I merged the other three first and attempted this one — three conflict regions, all mechanical:

1. A same-entry field union where #1838 adds `selfHostedOnly: true` and this PR edits the same object's `description`. Keep both.
2. `rpk cluster config lint` (deleted here) adjacent to `rpk cluster storage restore-start` (deleted in #1865). Both deletions stand.
3. `rpk connect studio sync-schema` (deleted in #1865) adjacent to `rpk transform logs` and `rpk generate license` (deleted here). All three deletions stand.

So: rebase after #1838/#1860/#1865 merge, resolve by union, nothing judgmental.

One caveat for the record: the extended `rpk ai auth login` and `rpk ai oauth-client dcr` descriptions add operational claims (credentials-file permissions, the public registration endpoint path) that I could not verify against the gateway source from here. They read consistent with the documented behavior; flagging only so the unverified bits are known. Those descriptions flow into adp-docs through the single-source stubs.

PR #1863, opened 2026-07-31
Re-render rpk docs when the overrides file changes

JakeSCahill:
## What

Closes an automation gap surfaced by #1862: `docs-data/rpk-overrides.json` is the curated-content store for the generated rpk reference pages, but no workflow watched it. An overrides edit merged and the rendered pages kept the old content until the next release-driven regeneration, so the repo carried overrides the published docs did not reflect.

New workflow: on any push to `main` or `beta` that touches the overrides file or its schema, re-render the full rpk tree from the newest committed snapshot and open a PR with the changed pages.

## Design

- **Pure re-render**: `doc-tools generate rpk-docs --from-json <newest snapshot>` with no diff and no What's new update. No rpk binary is needed because the snapshot already carries the tree and flags, so the run takes about two minutes.
- **Snapshot selection** uses the tilde-normalized `sort -V` (GA ranks above rc), same as the plugin receiver workflow in #1834.
- **Bot token** from Secrets Manager for PR creation so the generated PR actually runs checks (`secrets.GITHUB_TOKEN` PRs do not trigger Actions).
- **No loops**: the workflow only watches the overrides file and its schema, and a pure re-render writes only generated pages. Re-renders that change nothing exit green without opening a PR (generation is idempotent).
- **Superseded runs are cancelled** per branch rather than queued, since only the newest overrides state matters.
- `beta` picks the workflow up through the regular main to beta sync.

## Sequencing

Merge after redpanda-data/docs-extensions-and-macros#225 publishes doc-tools 5.3.0 (the workflow resolves `@^5.3.0` at runtime), the same constraint as #1834. Once this is in, merging #1860, #1861, and #1862 automatically produces the re-render PR that brings the pages in line with the fixed overrides.

## Validation

Workflow parses clean (act listing and YAML check). The generation command, snapshot resolution, idempotence, and no-changes behavior are the paths already exercised end to end in the #225 validation (18+ full regeneration rounds on a pristine clone, including repeated no-op re-renders).

See redpanda-data/docs-extensions-and-macros#225 for the train overview.

## Jira

Prevention layer for [DOC-2408](https://redpandadata.atlassian.net/browse/DOC-2408), and part of the fix chain for [DOC-2407](https://redpandadata.atlassian.net/browse/DOC-2407): once merged, overrides changes like #1838 regenerate the affected pages automatically.

### DIFF
--- a/.github/workflows/rerender-rpk-docs-on-overrides-change.yml
+++ b/.github/workflows/rerender-rpk-docs-on-overrides-change.yml
@@ -1,0 +1,123 @@
+# Re-renders the generated rpk reference pages whenever the overrides file
+# changes on a docs branch.
+#
+# The overrides file (docs-data/rpk-overrides.json) is the curated-content
+# store for generated rpk pages: descriptions, flag rewrites, added sections,
+# exclusions. Without this workflow, an overrides edit merges but the rendered
+# pages keep the old content until the next release-driven regeneration, so
+# the repo carries overrides that the published docs do not reflect.
+#
+# This is a pure re-render from the newest committed rpk snapshot: no rpk
+# binary, no diff, no What's new update. The run is idempotent, so a re-render
+# that changes nothing exits green without opening a PR. The PR only touches
+# generated pages, which cannot re-trigger this workflow (it only watches the
+# overrides file and its schema).
+name: Re-render rpk docs on overrides change
+
+on:
+  push:
+    branches: [main, beta]
+    paths:
+      - 'docs-data/rpk-overrides.json'
+      - 'docs-data/rpk-overrides.schema.json'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+  id-token: write
+
+# Only the newest overrides state matters: a superseded run for the same
+# branch is cancelled rather than queued.
+concurrency:
+  group: rpk-overrides-rerender-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  rerender:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout docs repo
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref_name }}
+          persist-credentials: false
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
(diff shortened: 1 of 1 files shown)


micheleRP (approved):
Approving the mechanism, with one ordering requirement that I would treat as a merge gate rather than a nicety.

**This workflow must merge after #1865, not before.** It runs `--from-json`, so it never installs a plugin and never extracts flags. Combined with #1865 that is a live regression on the published `rpk connect run` page.

The chain, all verified with the published 5.3.1 against main plus the three merged overrides PRs:

- The committed snapshot has `rpk connect run` with **0 flags**. Connect flag data only exists via live extraction.
- Until now that was masked: the override carried a hand-written 7-row Flags table that rendered unconditionally. #1865 removes it, correctly, because #225 made the extracted table authoritative.
- So a `--from-json` regeneration today renders that page with **no Flags section at all**: Usage straight to Global flags. Not a stale table, no table.

Which means: merge #1865 while this is live, the overrides change triggers a re-render, and the page silently loses its flag documentation.

The mitigation works, and it is just sequencing:

```
--plugin connect --rpk-bin   ->  Extracted flags for 16 command(s)
                             ->  Saved versioned JSON to docs-data/rpk-v26.2.1-rc2.json
                             ->  snapshot connect run flags: 13
then --from-json             ->  Flags sections: 1, rows: 13
```

So once the re-dispatch has persisted extracted connect flags into the committed snapshot, from-json regeneration is safe and this workflow is fine. Concretely, before merging #1865 or this, confirm the committed `docs-data/rpk-v26.2.1-rc2.json` has 13 flags on `rpk connect run`.

**Suggested follow-up in the generator:** nothing catches this class today. A command that rendered flags previously and now renders none is almost always a data problem, not a real change. A warning on a nonzero-to-zero flag-count transition would have surfaced this before it reached a page, and it generalizes to any plugin family whose flags come only from extraction.

## Ordering, tested

Four open PRs edit `rpk-overrides.json`. I test-merged all pairs: #1838, #1860, and #1865 are mutually clean in any order, and **only #1862 conflicts**, with both #1838 and #1865, in either direction. Running #1838 then #1860 then #1865 as a real sequence is clean, leaves valid JSON, and produces all four intended effects at once.

So: re-dispatch, confirm the snapshot has connect flags, then #1860 and #1865, then this, then rebase #1862.

PR #1871, opened 2026-08-02
Detect page-set changes when flagging adp-docs follow-ups

JakeSCahill:
Fixes the first of @micheleRP's follow-ups from the first production rpk ai run: the "Flag adp-docs follow-up" step skipped even though three connection pages were new.

**Root cause is a detection-model gap, not a broken grep**: the step keys on the diff summary's New/Removed/Deprecated command rows, which reflect the *command tree*. The connection pages appeared with **zero tree changes** — `rpk ai connection` existed in the 0.2.31 tree already (as coming-soon stubs), and what created the pages was #1869 removing their overrides excludes. Tree diff said nothing changed, so `structural=false` was computed correctly and was still the wrong answer for adp-docs, whose stubs track the **page set**.

**Fix**: the step now treats any added, deleted, or renamed page/partial under the plugin's own directories (`partials/rpk-<plugin>`, `pages/rpk/rpk-<plugin>`) as structural — via `git diff --staged --name-status` — while keeping the summary grep as the second signal for tree-level changes. Pure content modifications remain non-structural, so routine description refreshes don't spam the label.

Verified against the actual production case: the 03:57 run's changeset (three added `rpk-ai-connection*` partials) matches `^A` and would have set `structural=true`. The merge-time notify hook covered this instance, so this is belt-and-braces for visibility at PR time.

Follow-up from the #1868/#1869 production run.

### DIFF
--- a/.github/workflows/update-rpk-plugin-docs.yml
+++ b/.github/workflows/update-rpk-plugin-docs.yml
@@ -209,10 +209,23 @@ jobs:
             git diff --staged --name-only | head -20
           fi
 
-          # Structural changes (commands added/removed/deprecated) matter for
-          # downstream consumers with static stubs (adp-docs for rpk ai)
-          if grep -qE '\| (New|Removed|Deprecated) commands +\|' "${{ runner.temp }}/pr-summary.md" 2>/dev/null; then
+          # Structural changes matter for downstream consumers with static
+          # stubs (adp-docs for rpk ai). Two independent signals, because the
+          # command tree and the page set can change independently: the rpk ai
+          # connection pages appeared with ZERO tree changes (the commands
+          # existed as stubs in 0.2.31; removing their overrides excludes is
+          # what created the pages), so the summary-table grep alone missed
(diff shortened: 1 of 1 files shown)


micheleRP (approved):
Approving. This closes a gap I would not have predicted from reading the original condition, and the comment explaining why is the valuable part of the change.

The summary-table grep only fires when the command tree changes, and the rpk ai connection pages appeared with the tree unchanged: the commands already existed as stubs, and removing their `exclude` overrides is what created the pages. So the one signal that mattered was invisible to the one signal being checked. Adding page-set add/delete/rename as an independent trigger is the right fix, and restricting it to `^[ADR]` so pure content edits do not flag is the right scope.

Two small things I checked rather than assumed: the pathspec includes `modules/reference/pages/rpk/rpk-${PLUGIN}`, which does not exist for `ai` since it renders as partials, and a `git diff` path filter that matches nothing is fine rather than an error. And `--name-status` has rename detection on by default, so the `R` case is reachable.

PR #1881, opened 2026-08-04
Fix link to relocated contribution/authoring guide (DOC-2303)

JakeSCahill:
## Description

Related to https://redpandadata.atlassian.net/browse/DOC-2303.

`docs-site/meta-docs/CONTRIBUTING.adoc` was split and renamed to `AUTHORING.adoc` (redpanda-data/docs-site#201), then that file was deleted entirely and ported to internal-only `redpanda-data/cupboard` (redpanda-data/cupboard#677) — every one of these repos' `local-antora-playbook.yml` cross-fetches the private `docs`/`cloud-docs`/`rp-connect-docs`/`adp-docs` repos to build locally at all, so the audience for that guide was already internal in practice.

This repoints both the "clone locally" contribution link and the "Set up GitHub authentication" pointer at the ported cupboard guide.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -53,7 +53,7 @@ If you are a Redpanda employee, submit doc issues in `redpanda-data/documentatio
 You have two options to contribute to the documentation:
 
 . Directly edit a page on GitHub by selecting **Make a contribution** > **Edit on GitHub** located at the bottom of a documentation page.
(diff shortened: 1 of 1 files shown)


micheleRP (approved):
Approving — links verified and the guidance is correct. Non-blocking nit: the blurb lists adp-docs among the repos this build fetches, but this repo's local-antora-playbook.yml doesn't fetch adp-docs. Since you'll rebase after #1879 merges anyway, consider dropping the repo names entirely ("fetches content from private Redpanda repositories — see the authoring guide") so the sentence stays accurate everywhere.

